PR #4, opened 2022-09-23
[WIP] data flow diagrams

gorkem:


amfred:
For the Personal Data, here's what we are processing and/or storing:

- We have user account information (username, user ID) that comes from the SSO system, gets passed through HAC, and then on to KCP to select the right workspace on each request.  KCP will also end up storing the username in the home workspace.
- Once we have the user registration service, it will also take the current user’s account information and probably their phone number as input, but we don’t store the phone number - only a hash of it with the userid.  Plus, users can enter invitations (name, email address) for other people, and the user registration controller will store that invitation data in CRs.
- The user registration service will also send phone numbers through Amazon SNS for messaging, and user account data + invitation data through Eloqua for email confirmation and account updates.
- For SPI, we have the users’ credentials for 3rd party services coming in.  Will all of those end up in Vault now?
- We’ll send audit logs from the workload clusters to Amazon CloudWatch, Amazon S3 (for long term storage), and Snyk.  Those contain usernames and IP addresses.
- We’ll also need to design a way for users to enter environment variables that need to be encrypted - such as authentication tokens and other secrets.  I suppose those would get stored in Vault as well?
- The GitOps artifacts include usernames in their URLs, and those also get stored in Application and Component CRs.

gorkem:
@amfred Could KCP team be handling some of these? User registration and SSO are under their purview. Let's check with KCP. I have just uploaded a new version with SPI diagram separated, it works better with GH UI.

amfred:
Can we get the data flow diagram done this week so we don't have to carry this over to M9?

amfred:
@gorkem  - now that we have #44 in we don't need what I requested in my earlier review to be in these diagrams.  I propose merging these in as well because they add to the overall picture.  What do you think?

amfred (approved):

amfred:
I think we can close this old PR.  Please re-open if you disagree.

PR #5, opened 2022-10-05
ADR-0005 -Service Account configuration for image builds in AppStudio

sbose78:


sbose78:
@adambkaplan @Michkov to review.

ADR/pipeline-service/adr-0001.md
```diff
@@ -0,0 +1,46 @@
...
+Use of any other service accounts will not lead to a successful assembling of the image.
+
+* The AppStudio Build controller would tap into this platform contract to create relevant PipelineRuns for building container images.
+
+## Consequences
+
+
+### Implementation - AppStudio Build Service Controller
+
+* The AppStudio Build controller creates an "image-builder" service account in the KCP workspace/namespace.
```
gabemontero, line 28: can jvm-build-service reuse the SA created by appstudio build controller, or does it have to separately do similar actions @sbose78 ?

@stuartwdouglas FYI

ADR/pipeline-service/adr-0001.md
```diff
@@ -0,0 +1,46 @@
...
+* The AppStudio Build controller links the image pull secret ( previously created by SPI ) with the "image-builder" `serviceAccount` in the KCP workspace/namespace. Upon doing so, the 'updated' "image-builder" `serviceAccount` would get synced into the workload cluster.
+* The AppStudio Build controller" creates the PipelineRun referencing the "image-builder" service account as the one that the PipelineRun/TaskRun `Pods` would run as. Upong doing so, the PipelineRun would get synced into the Pipeline Service Workload cluster.
```
Roming22, line 30: `controller"` -> `controller`
`Upong` -> `Upon`

ADR/pipeline-service/adr-0001.md
Roming22: If the service is always used in scoe of AppStudio or HACBS, I don't see a reason why you would need to create another service account.
If the service was to become standalone, then I understand why you would need to manage your own service account.

ADR/pipeline-service/adr-0001.md
```diff
@@ -0,0 +1,46 @@
...
+
+Date: Oct 5, 2022
+
+## Status
+
+Proposed
+
+## Context
+
+The `serviceAccount` used for running a `PipelineRun` workload in a workload cluster needs to have the right SCCs and linked credentials to be able to successfully execute a container image build. The linked credentials will mostly originate from the KCP workspace where the `PipelineRun` was synced from.
```
Roming22, line 11: Please add information about why linked credentials are required/why we cannot rely on SPI magic.

ADR/pipeline-service/adr-0001.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Consequences
+
+
+### Implementation - AppStudio Build Service Controller
+
+* The AppStudio Build controller creates an "image-builder" service account in the KCP workspace/namespace.
+* The AppStudio Build controller links the image pull secret ( previously created by SPI ) with the "image-builder" `serviceAccount` in the KCP workspace/namespace. Upon doing so, the 'updated' "image-builder" `serviceAccount` would get synced into the workload cluster.
+* The AppStudio Build controller" creates the PipelineRun referencing the "image-builder" service account as the one that the PipelineRun/TaskRun `Pods` would run as. Upong doing so, the PipelineRun would get synced into the Pipeline Service Workload cluster.
+* And that's it! Since the `[cluster]rolebinding` in the Workload cluster for all "image-builder" `serviceAccounts` would be pre-configured ( by the Pipeline Service ) with the SCC needed to build container images, the image should be successfully assembled.
```
Roming22, line 31: I'm fine with implementing this solution as long as we have an exit strategy.

We're starting to care about the workloads we are running, and I wish that was not the case. What happens when we need to build/push something else that also requires this type of credentials? How many *-builder SA will we have to maintain in the future. Do we have any guarantee that it won't be more than one?

ADR/pipeline-service/adr-0001.md
```diff
@@ -0,0 +1,46 @@
...
+* And that's it! Since the `[cluster]rolebinding` in the Workload cluster for all "image-builder" `serviceAccounts` would be pre-configured ( by the Pipeline Service ) with the SCC needed to build container images, the image should be successfully assembled. 
+* Additionally, since the service account had 
+the linked credentials, the container image should be successfully pushed as well.
+
+
+### Implementation - Pipeline Service 
+
+* The contract/API between users and the platform is the "image-builder" `serviceAccount` which consumers would be expected to use in the KCP workspace 
+for `PipelineRuns` if they wish to build container images.
+* The Pipeline Service platform would configure workload clusters with the `[cluster]rolebindings` necessary to have the "image-builder" `serviceAccount`
```
Roming22, line 40: Related to a previous comment of mine. I'm fine with deploying/maintaining the required SCC, as the requirement for a specific SCC is strictly needed to be able build images, and therefore should not apply or be required by any other workload ever.

ADR/pipeline-service/adr-0001.md
```diff
@@ -0,0 +1,46 @@
...
+
+* The contract/API between users and the platform is the "image-builder" `serviceAccount` which consumers would be expected to use in the KCP workspace 
+for `PipelineRuns` if they wish to build container images.
+* The Pipeline Service platform would configure workload clusters with the `[cluster]rolebindings` necessary to have the "image-builder" `serviceAccount` 
+configured with a tailor-made SCC such that they are sufficiently capable of building container images. 
+
+
+### Security 
+
+* Users should not be able to alter role bindings in their `workspaces`/`namespaces` in a way that they are synced into the workload clusters. This will help prevent SCC escalations by the users/services whose workloads would be scheduled on the PipelineService workload cluster.
```
Roming22, line 46: Ideally Pipeline Service would use a ClusterRole, and ClusterRoleBindings SHOULD NOT be synced back from a workspace consuming any APIExport, as it would mean that a user's configuration could impact a service cluster.

Ideally a user should be free to add any Role/RoleBinding, secrets they want/need in their workspace, including to the SA account in order to satisfy the requirements for their workload.

ADR/pipeline-service/adr-0001.md
gabemontero: yep my assumption/hope was that "share/reuse" would be the answer, but I did not want to assume / get explicit confirmation / consensus

ADR/pipeline-service/adr-0001.md
```diff
@@ -0,0 +1,46 @@
...
+
+* The AppStudio Build controller would tap into this platform contract to create relevant PipelineRuns for building container images.
+
+## Consequences
+
+
+### Implementation - AppStudio Build Service Controller
+
+* The AppStudio Build controller creates an "image-builder" service account in the KCP workspace/namespace.
+* The AppStudio Build controller links the image pull secret ( previously created by SPI ) with the "image-builder" `serviceAccount` in the KCP workspace/namespace. Upon doing so, the 'updated' "image-builder" `serviceAccount` would get synced into the workload cluster.
```
ramessesii2, line 29: Does this mean that only `AppStudio Build controller` can do a create/patch on the ServiceAccount? 
I ask this because if a user has this capability, they would be able to patch the SA and in that case, what would be the criteria for the controller to access if the SA is good to be synced(by the syncer) to the Workload cluster?

ADR/pipeline-service/adr-0001.md
Michkov: jvm-build-service or any other service except build-service should not edit/use linked secrets of the service account.

Michkov, line 28: By default `builder` SA is created on compute clusters, should not we use name connected with pipeline-service? eg. `pipeline-service-builder` ?

ADR/pipeline-service/adr-0001.md
Michkov, line 29: Since we are going to have only one service account there will be limitation at least for quay.io that all images in the build namespace must go into the same organization (cannot create robot account which would be able to push to multiple organizations).

sbose78: > there will be limitation at least for quay.io that all images in the build namespace must go into the same organization


Agreed, I will call that out. I think by "organization", you implied "quay.io organization".  I think that's a reasonable limitation.

ADR/pipeline-service/adr-0001.md
gabemontero: So we need a decision @sbose78 on what services outside of appstudio build service should be doing here

adambkaplan: I agree this should is something that should be tied to the Pipeline Service. Part of what the Pipeline Service should provide is a service account capable of executing builds with reasonably elevated pod security permissions.

Something to consider - would a future user or on-top service want this disabled? For example - a service wants to use Pipeline Service as a workflow orchestrator, and does not want the ability to build container images. We've seen this situation play out with OpenShift Builds, where customers want to disable things that could be a security risk.

I think this is feasible with an API that the pipeline service workspace controller reconciles.

ADR/pipeline-service/adr-0001.md
adambkaplan: @s-urbaniak can workspace controllers create any `ClusterRole`? Can a user create any `ClusterRole` in their workspace?

ADR/pipeline-service/adr-0001.md
sbose78: Let me ponder on this a little and respond.

ADR/pipeline-service/adr-0001.md
gabemontero: @adambkaplan 's second bullet here, at least as I like to read into it, has already come up with jvm-build-service on KCP progress I have been able to make today.  Though the added permissions I needed were not related to pod security, but additional API access.

Specifically, I have add to add RBAC permission on the workload cluster to allow the 
- "build SA" to be able to update taskruns/status as part of our `maven-run : analyse-dependencies` step (you get this step when you use the HACBS bundle to build java)
- to allow the maven proxy / artifact cache, which ideally we like to run under the "build SA" given the common permission requirements, to see reasonably simple things like get/watch/list configmaps to obtain needed metadata

It occurred to me that these short term hacks I had to perform conceivably all under the umbrella of this design and what it is trying to achieve.  Or at least close enough we could stretch this design ever so slightly.

WDTY @sbose78 @adambkaplan 

@stuartwdouglas FYI

adambkaplan: Conceptually the RBAC for SCC and the RBAC to update taskrun status is the same - it is an elevated permission that applies to the workload cluster, but not KCP.

gabemontero: Yep @adambkaplan 's  wording lines up with what I was trying to say previously.  And since it is conceptually the same, it would ideal if we can solve both concerns with this effort.

gorkem:
@sbose78 Can you put this directly under the `/ADR` folder?

gorkem:
How would we solve this if we were not syncing Tekton resources to worklaod clusters but only the low level APIs?

ADR/pipeline-service/adr-0001.md
sbose78: @Roming22  I've added the note on exit strategy in the "Security considerations" section. I'd say, the Pipeline Service team takes the call on who can be allowed to use elevated privileges. We'll need to build the machinery for that. Consumers may choose to opt-in but in the end, whoever owns the workload cluster is the ultimate approver. As of today, it's going to be us. However, when customers bring their own workload clusters, we should provide them with the machinery to manage such privileges.

sbose78: ^ We'd have to take a call on"when" we'd like to prioritize building the fancy machinery :)

sbose78: @gabemontero @adambkaplan Agreed.

I can almost see a related ADR cc @Roming22 :

"As a consumer Foo, I need special permissions on the service account that I'm going to use".

* What are the boundaries ? The service account permissions specific to the consumer cannot break out of the namespace in the workload cluster. In this case, the consumer wants to read/write TaskRuns in the same namespace which feels allow-able.
* How does the consumer define such a configuration ? Pipeline Service infra should provide a framework for specifying such custom ClusterRoles whose policies are local to a namespace.

Clearly, we shouldn't be proliferating our ClusterRole with unrelated rules, rather providing a framework to create new ones which would have to be "approved" into the system and then associated with new/existing service accounts.

ADR/pipeline-service/adr-0001.md
sbose78: > Something to consider - would a future user or on-top service want this disabled? For example - a service wants to use Pipeline Service as a workflow orchestrator, and does not want the ability to build container images. We've seen this situation play out with OpenShift Builds, where customers want to disable things that could be a security risk.

I'd say, whoever owns the workload cluster has a greater reason to worry and therefore is the one taking the call on who gets to use the feature.

ADR/pipeline-service/adr-0001.md
gabemontero: So @sbose78 @adambkaplan @Roming22 @jduimovich @stuartwdouglas I think the discussion in https://coreos.slack.com/archives/C032EJ007C0/p1666186257512109, which arose from our discussion in the cabal 2 days ago, where that cabal agenda item from @sbose78 cited this discussion thread, landed us in a tenable spot wrt jvm-build-service needs, though I feel we need to close the loop and reflect that outcome in your ADR here @sbose78 , and ultimately get the appropriate Jira's for @Roming22 's team (unless they already exist @Roming22 ?).

The high level outcome from https://coreos.slack.com/archives/C032EJ007C0/p1666186257512109:
- https://github.com/openshift-pipelines/pipeline-service-workspace-controller will sync an SA and linked secret containing the kubeconfig and token needed to communicate with the KCP api server
- jvm-build-service will use said SA/secret with the pipelineruns it generates for processing artifacts/dependencies
- those pipelineruns can then update jvmbuildservice defined custom resources with the deploy artifact information vs. the current updating of taskrun/status with results (where tekton then aggregates to pipelinerun/status, and the syncer currently sinks back to kcp)
- any RBAC for updating those customresources can now be defined on the kcp side vs. compute side, and hence make sense to be entirely included in the manifest list jvm-build-service would provide to  @rajivnathan 's https://issues.redhat.com/browse/ASC-181 as part of establishing a hacbs enabled workspace for a user.

And this alleviates the concern around jvm-build-service related manipulation of results off the pipelinerun which enterprise contract subsequently inspects, as that manipulation would cease.

My remaining questions
- any remaining concerns?
- is there just one SA with linked kcp token/kubeconfig secret and with workload cluster SCC related changes for image builds ?
- or do we want on SA for building images, and another SA for "kcp aware" pipelineruns like what jvm-build-service needs?

gabemontero: cross referencing with https://coreos.slack.com/archives/C02CTEB3MMF/p1666786789821739

learned from @scoheb today that other controllers under the HACBS umbrella besides jvm build service also have the requirement of a KCP aware SA+secret so that pipelineruns created by KCP aware controllers can manipulate KCP only CRs.

ADR/adr-0005-image-builder-sa.md
```diff
@@ -0,0 +1,52 @@
...
+
+
+
+### Implementation - Pipeline Service 
+
+* The contract/API between users and the platform is the "pipeline-service-image-builder" `serviceAccount` which consumers would be expected to use in the KCP workspace 
+for `PipelineRuns` if they wish to build container images. 
+* The Pipeline Service platform would configure workload clusters with the `[cluster]rolebindings` necessary to have the "pipeline-service-image-builder" `serviceAccount`. The infrastructure machinery driving the deployment of the `clusterrole`/`clusterrolebinding` should be capable of upgrading the privileges being assigned.
+configured with a tailor-made SCC such that they are sufficiently capable of building container images.
+* The service account does not necesarily need to be created in advance in the workload cluster by the Pipeline Service platform. Consumers of the Pipeline Service ( as specified in the next section ) would create this service account in the KCP workspace and have the same synced into the workload cluster's namespace. When the service account does get synced, the `[cluster]rolebindings` configured should kick-in to privide the appropriate elevation of privilege.
```
Roming22, line 35: `necesarily` -> `necessarily`
`privide` -> `provide`

ADR/adr-0005-image-builder-sa.md
```diff
@@ -0,0 +1,52 @@
...
+
+
+### Implementation - AppStudio Build Service Controller
+
+* The AppStudio Build controller creates an "pipeline-service-image-builder" service account in the KCP workspace/namespace. This service account needs to be created so that it's made available in the workspace for linking the secrets in the subsequent step.
```
Roming22, line 40: `creates an` -> `creates a`

Roming22, line 40: non-AppStudio/HACBS users will have to create this account manually. How about Pipeline Service creating the account and an empty secret, so that the user/controller only has to update the secret to have a working environment?

sbose78: Makes sense - would this be similar to your recommendation on setting up the service account on the kcp workspace for jvm-build-service users ?

Roming22: I think it is. We would provide 3 default users:
- `pipeline`: default account created by the tekton pipelines. Can be used for the most simple workloads.
- `pipeline-service-runner`: linked to KCP workspace. To be used when access to workspace CRs is required. Same permission levels as `pipeline`
- `pipeline-service-image-builder`: linked to KCP workspace, extra privileges on the workload cluster.

ADR/adr-0005-image-builder-sa.md
```diff
@@ -0,0 +1,52 @@
...
+* The AppStudio Build controller links the image pull secret ( previously created by SPI ) with the "pipeline-service-image-builder" `serviceAccount` in the KCP workspace/namespace. Upon doing so, the 'updated' "pipeline-service-image-builder" `serviceAccount` would get synced into the workload cluster.
+* The AppStudio Build controller creates the PipelineRun referencing the "pipeline-service-image-builder" service account as the one that the PipelineRun/TaskRun `Pods` would run as. Upon doing so, the PipelineRun would get synced into the Pipeline Service Workload cluster.
+* And that's it! Since the `[cluster]rolebinding` in the Workload cluster for all "pipeline-service-image-builder" `serviceAccounts` would be pre-configured ( by the Pipeline Service ) with the SCC needed to build container images, the image should be successfully assembled. 
+* Additionally, since the service account had 
+the linked credentials, the container image should be successfully pushed as well.
+
+
+### Security considerations
```
ralphbean, line 48: Suggestion on structure - move the `## Consequences` section below the `### Implementation ...` section.

The Implementation section is like an example with more information for the Decision section.

These security considerations are like additional consequences to be considered that become true as a result of the Decision.

ADR/adr-0005-image-builder-sa.md
```diff
@@ -0,0 +1,52 @@
...
+
+* Users should not be able to alter role bindings in their `workspaces`/`namespaces` in a way that they impact the execution environment in the workload clusters. This will help prevent SCC escalations by the users/services whose workloads would be scheduled on the PipelineService workload cluster.
```
ralphbean, line 50:
```suggestion
* Users must not be able to alter role bindings in their `workspaces`/`namespaces` in a way that they impact the execution environment in the workload clusters. This will help prevent SCC escalations by the users/services whose workloads would be scheduled on the PipelineService workload cluster.
```

ADR/adr-0005-image-builder-sa.md
```diff
@@ -0,0 +1,52 @@
...
+* Build Controller would have to be granted permissions to create/edit service accounts in the workspace to be to create/update the `pipeline-service-image-builder` service account for linking secrets/credentials.
```
ralphbean, line 51:
```suggestion
* Build Controller will have to be granted permissions to create/edit service accounts in the workspace to be able to create/update the `pipeline-service-image-builder` service account for linking secrets/credentials. 
```

ADR/adr-0005-image-builder-sa.md
```diff
@@ -0,0 +1,52 @@
...
+* And that's it! Since the `[cluster]rolebinding` in the Workload cluster for all "pipeline-service-image-builder" `serviceAccounts` would be pre-configured ( by the Pipeline Service ) with the SCC needed to build container images, the image should be successfully assembled. 
+* Additionally, since the service account had 
+the linked credentials, the container image should be successfully pushed as well.
+
+
+### Security considerations
+
+* Users should not be able to alter role bindings in their `workspaces`/`namespaces` in a way that they impact the execution environment in the workload clusters. This will help prevent SCC escalations by the users/services whose workloads would be scheduled on the PipelineService workload cluster.
+* Build Controller would have to be granted permissions to create/edit service accounts in the workspace to be to create/update the `pipeline-service-image-builder` service account for linking secrets/credentials. 
+* The Pipeline Service platform should be able to disallow usage of this elevated privilege by certain consumers of the service. The design for the same should be addressed in a future ADR.
```
ralphbean, line 52: Why disallow? Can you give an example - and, what are the problems that arise if we cannot disallow?

ADR/adr-0005-image-builder-sa.md
```diff
@@ -0,0 +1,52 @@
...
+Date: Oct 5, 2022
+
+## Status
+
+Accepted
+
+## Context
+
+The `serviceAccount` used for running a `PipelineRun` workload in a workload cluster needs to have the right SCCs and linked credentials to be able to successfully execute a container image build. The linked credentials will mostly originate from the KCP workspace where the `PipelineRun` was synced from.
+Additionally, the SCC association needed for elevating capabilities required for image builds isn't something we'd prefer to expose the user to.
```
ralphbean, line 12:
```suggestion
The `serviceAccount` used for running a `PipelineRun` workload in a workload cluster needs to have the right SCCs and linked credentials to be able to successfully execute a container image build.

* Users should be able to influence the linked credentials. These are typically things like push secrets (see [ADR 0004](./0004-component-image-location.md)) but also user-provided secrets for interacting with third-party services.
* Users should not be able to influence the SCC association needed for elevating capabilities generally, but we do need to give them a way to assume the SCC required for image builds.
```

ADR/adr-0005-image-builder-sa.md
```diff
@@ -0,0 +1,52 @@
...
+
+
+## Decision
+
+* As a Pipeline Service platform feature, any service/user intending to build a container image would need to create and use the "pipeline-service-image-builder" service account in the KCP workspace. 
+Use of any other service accounts will not lead to a successful assembling of the image.
+
+* The AppStudio Build controller would tap into this platform contract to create relevant PipelineRuns for building container images.
+
+* The Pipeline Service Platform would configure the "pipeline-service-image-builder" service account in the execution environment ( ie, the workload cluster ) out-of-the-box.
```
ralphbean, line 23:
```suggestion
* The Pipeline Service Platform would configure the "pipeline-service-image-builder" service account in the execution environment ( ie, the workload cluster ) out-of-the-box ([PLNSRVCE-809](https://issues.redhat.com/browse/PLNSRVCE-809)).
```

ADR/adr-0005-image-builder-sa.md
Roming22: My understandind is that we deal with that by syncing SA & secrets, but we won't ever be syncing roles and rolebindings.

PR #6, opened 2022-10-11
AppStudio. Workspace-Controller-Compute Layout diagram

alexeykazakov:


amfred:
We'll also have an App Studio Ops BYO cluster.  So far it just has Grafana on it; it will probably have Thanos later, and it will probably end up with other services like the QE dashboard as well.  At the moment these things are on the App Studio KCP cluster with everything else, and the ops tools can stay there indefinitely until there's a reason to move them. For Service Preview, I want to have a new App Studio Control Plane production cluster, for several reasons.

alexeykazakov:
@amfred this diagram is focusing on the Workspace - Controller - Compute relations. The ops tools is a bit different story. I think we will need a different diagram for that. I would like to avoid adding everything in a single diagram because it may become unreadable. But we can always add the Ops tools later to this diagram if we really want.

gorkem:
The `manages` labeled link from ArgoCD controller goes to only prod environment which creates the impression that only prod is managed by GitOps. Can we add links that extent to other environments as well?

alexeykazakov:
> Can we add links that extent to other environments as well?

Added.

amfred (approved):

alexeykazakov:
@amfred let's have a conversation about our Ops stack. I think it's clear what we want to do now but let's talk about what we want to do for SP. And we will either add it to this diagram (as a new PR) or create a new diagram for that.

PR #14, opened 2022-10-21
ADR-0003: Interacting with Internal Services

scoheb:


ADR/0003-interacting-with-internal-services.md
```diff
@@ -0,0 +1,81 @@
...
+
+---
+
+Use a "controller" running in a private cluster that can watch and reconcile **Request** custom resources in
+one or more workspaces. This will be referred to as the **Internal Service Controller**.
+
+**Request** is used here as a general type meaning that real use cases might involve custom resources
+such as **ServiceABCRequest**.
+
+This strategy will make use of [KCP]'s VirtualWorkspace model allowing a controller to watch a group of
```
ralphbean, line 34:
```suggestion
This strategy will make use of [KCP]'s VirtualWorkspace model allowing an internal service controller to watch a group of
```

ADR/0003-interacting-with-internal-services.md
```diff
@@ -0,0 +1,81 @@
...
+workspace via a single KUBECONFIG.
+
+This internal service controller is expected to trigger a specific job that encapsulates the Internal Service's unit of work
+that HACBS wants to initiate.
+
+It is expected that the controller should update the **status** of the **Request** CR to denote the progress of the
+triggered unit of work.
+
+The controller should also be able to update the **Request** CR to provide a **result** back to the process that
```
ralphbean, line 43:
```suggestion
The internal service controller should also be able to update the **Request** CR to provide a **result** back to the process that
```

ADR/0003-interacting-with-internal-services.md
```diff
@@ -0,0 +1,81 @@
...
+originally created the custom resource.
+
+**Example:**
+
+During the course of an attempt to release content, artifacts may need to be signed. The service that
+performs the signing process is an internal service within an organization with no publicly addressable API.
+The [release-service] may execute a release strategy that has a step that wants to access that signing service's
+API and obtain a result payload to be used in downstream steps in the release strategy.
```
ralphbean, line 51:
```suggestion
API and obtain a signature to be used in downstream steps in the release strategy.
```

ADR/0003-interacting-with-internal-services.md
```diff
@@ -0,0 +1,81 @@
...
+
```
ralphbean, line 52:
```suggestion

Using the pattern here, the user organization (called MyOrg) would create an signing controller hosted in a cluster inside their network, but which uses a KUBECONFIG pointed at the kcp VirtualWorkspace for a `MyOrgSigningRequest`. They would construct a release pipeline which creates those `MyOrgSigningRequest` CRs in their managed workspace, and which watches for `status` updates on those request CRs to determine when the internally hosted signing process has completed.

```

ADR/0003-interacting-with-internal-services.md
```diff
@@ -0,0 +1,81 @@
...
+New
+
+## Context
+
+---
+
+Many organizations, including Red Hat, possess numerous internal services that help productize their software.
+In many cases, these internal services will continue to play a role in the release workflows used in AppStudio/HACBS.
+
+It is extremely important to design the interaction with internal services in such as way as to avoid any potential
```
ralphbean, line 18:
```suggestion
We originally thought that we should expose access to an organization's internal services by encouraging the use of "[bastion](https://en.wikipedia.org/wiki/Bastion_host)" interfaces that are publicly addressable but which also had some degree of internal network access. On review, we see now that ingress like this opens up unwanted
```

scoheb:
@ralphbean changes applied. PTAL

ralphbean:
Newline breaks are a little weird now. Otherwise, looks good to me. @Allda can you review too?

Allda:
The ARD with architecture diagrams looks good to me. :+1:

ADR/0003-interacting-with-internal-services.md
```diff
@@ -0,0 +1,79 @@
...
+
+---
+
+* How can internal services controllers control who they accept requests from?
+
+## Consequences
+
+---
+
+* If the managed workspace for a customer is a cluster that they control, that teams will have to install their own
```
amfred, line 67: The wording is confusing here.  What CRDs will a customer have to install?  Also - what does "teams" refer to here?

scoheb: Updated wording. please review

amfred:
Have you considered using a reliable message queue instead of watching these resources?
https://aws.amazon.com/blogs/apn/using-amazon-sqs-in-a-multi-tenant-saas-solution/

ADR/0003-interacting-with-internal-services.md
amfred: The wording is better, thanks.

scoheb:
@ralphbean would you be able to give this another review and approval?

ralphbean (approved):
Yeah, this looks good. Sorry for letting the review hang. Let's merge it.

PR #15, opened 2022-10-29
ADR-0004 - Out-of-the-box image repository in Stonesoup

sbose78:
Longer discussion https://docs.google.com/document/d/1_1xB6aBB7XW6JyYZVdnS8mC5Ftg4oksSThiLt8fBUcs/edit#

**Problem**
AppStudio does not have a internal registry where images could be pushed to as an intermediate step before being deployed as a container.
As an application centric experience, AppStudio should not make it mandatory for its users to specify where the image ( built from source code ) needs
to be pushed to.

**Goals**: 
* Provide an out-of-the-box location for users’ images to be pushed after being built from source.

### DIFF
--- a/ADR/0004-component-image-location.md
+++ b/ADR/0004-component-image-location.md
@@ -91,7 +91,7 @@ We would use the user's API token for working with the relevant Quay.io organiza
   account with the user's Quay.io account ?
 - Considering the above is figured out, we would need to add existing users as owners to the relevant organizations. This is a backend job that 
   would need to be designed and executed at an appropriate time.
-
+- In the HACBS context, giving ownership of the image repo to users is not a problem for provenance, because integrity of images is based on attestations which are only treated as valid if signed by pipeline-service (by tekton chains). While users can in theory push images built on their laptop, the [release-service](../book/release-service.md) won't treat them as if they were built by the [pipeline-service](https://github.com/openshift-pipelines/pipeline-service). They will fail to pass the [enterprise contract](../book/enterprise-contract.md).
 
 
 ## Consequences
--- a/ADR/0004-component-image-location.md
+++ b/ADR/0004-component-image-location.md
@@ -175,7 +175,7 @@ spec:
 
 ## Consequences
 
-- We will be able to deprecate the use of https://quay.io/repository/redhat-StoneSoup/user-workload for storing users' container images.
+- We will be able to deprecate the use of https://quay.io/repository/redhat-appstudio/user-workload for storing users' container images.
 - Users will not be forced to put in an image repository location when they use StoneSoup to import and deploy source code.
 - The image repository used could conditionally be made available to users outside of StoneSoup.
 - Given that scoped Quay.io robot account tokens would be available in user's workspaces for pushing/pulling images, the principle of minimum privilege
@@ -184,4 +184,4 @@ spec:
 ## References
 
 * Originally drafted in an internal [document](https://docs.google.com/document/d/1KcXWZ8VGUg_iR0RjdGuDYedP8ZW63XCgF26KZUNgpeQ/edit)
-* Implementation: https://github.com/redhat-StoneSoup/image-controller
+* Implementation: https://github.com/redhat-appstudio/image-controller
--- a/ADR/0004-component-image-location.md
+++ b/ADR/0004-component-image-location.md
@@ -17,6 +17,8 @@ Accepted
 * Gorkem Ercan
 * Parag Dave
 * Ralph Bean
+* Andrew McNamara
+
 
 ## Context
 
@@ -69,7 +71,7 @@ to be pushed to.
 | ----------- | ----------- |
 | Quay.io organization      | quay.io/redhat-user-workloads       |
 | OAuth Application  name | Created, name redacted        |
-| Account used to generate token | `shbose` |
+| Account used to generate token | `shbose` , `mkovarik` |
 | Scope | Administer organizations, Adminster repositories, Create Repositories |
 
 <img width="1313" alt="image" src="https://user-images.githubusercontent.com/545280/212758440-5807cd4e-11b1-43bc-aa03-385b9284cb9e.png">
@@ -107,6 +109,7 @@ Until the capability to determine the associated user/tenant/Space a Component i
 
 * Upon removal of a user from Stonesoup,
     * The empty Quay.io organization associated with the user or the user's Space may not be deleted instantly, but would be scheduled for a delayed cleanup.
+* PR-based tags are to be deleted on a regular basis. Image tags associated with `main` may remain un-pruned for now.
     
     
 ### How - Implementation
@@ -170,7 +173,7 @@ spec:
   account with the user's Quay.io account ?
 - Considering the above is figured out, we would need to add existing users as members of the relevant organizations. This is a backend job that 
   would need to be designed and executed at an appropriate time.
-- In the HACBS context, giving ownership of the image repo to users is not a problem for provenance, because integrity of images is based on attestations which are only treated as valid if signed by pipeline-service (by tekton chains). While users can in theory push images built on their laptop, the [release-service](../book/release-service.md) won't treat them as if they were built by the [pipeline-service](https://github.com/openshift-pipelines/pipeline-service). They will fail to pass the [enterprise contract](../book/enterprise-contract.md).
+
 
 
 ## Consequences
--- a/ADR/0004-component-image-location.md
+++ b/ADR/0004-component-image-location.md
@@ -17,6 +17,7 @@ Accepted
 * Gorkem Ercan
 * Parag Dave
 * Ralph Bean
+* Ann Marie Fred
 
 ## Context
 
--- a/ADR/0004-component-image-location.md
+++ b/ADR/0004-component-image-location.md
@@ -1,4 +1,4 @@
-# 4. Out-of-the-box image repository for AppStudio users 
+# 4. Out-of-the-box image repository for StoneSoup users 
 
 Date: Oct 29, 2022 
 
@@ -17,13 +17,12 @@ Accepted
 * Gorkem Ercan
 * Parag Dave
 * Ralph Bean
-* Ann Marie Fred
 
 ## Context
 
 **Problem**
-AppStudio does not have a internal registry where images could be pushed to as an intermediate step before being deployed as a container.
-As an application centric experience, AppStudio should not make it mandatory for its users to specify where the image ( built from source code ) needs
+StoneSoup does not have a internal registry where images could be pushed to as an intermediate step before being deployed as a container.
+As an application centric experience, StoneSoup should not make it mandatory for its users to specify where the image ( built from source code ) needs
 to be pushed to.
 
 
@@ -38,7 +37,7 @@ to be pushed to.
 
 **Design Goals**
 * Use Quay.io as the out-of-the-box image registry in order to avoid getting into the business of maintaining an internal registry.
-* Align user permissions in an AppStudio workspace with those in Quay.io
+* Align user permissions in an StoneSoup workspace with those in Quay.io
 * Maintain the right levels of isolation between images being expected of a multi-tenant system.
 * Security:
   * Leverage short-lived Quay.io API tokens for repo/org management on Quay.io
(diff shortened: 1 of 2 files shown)


alexeykazakov (approved):

ADR/0004-component-image-location.md
```diff
@@ -0,0 +1,107 @@
...
+#### Short-term
+
+##### Setup
+* Configure the a pre-configured Quay.io API token for repo/org management, also known as the OAuth Access token. From a security perspective, this would be treated as a typical service account token
+  belonging to AppStudio. This token would be shortlived and rotated *as frequently as practically possible*  .
+
+##### When a user creates a Component
+1. Create a new Quay.io org owned by David using the pre-configured AppStudio-wide API token `quay.io/unique-org-david`
+2. Create a new image repo for Component “foo” using the the pre-configured David’s API token `quay.io/unique-org-david/unique-component-foo`
+3. Optionally, if needed, make the image repo public using the API token for easy image pulls.
```
gorkem, line 78: We can not make the repository public. The user owns the repo, it should be their decision to make it public. Otherwise, we will risk exposing users' private applications. Other than David nobody else should be able to access it.

ADR/0004-component-image-location.md
```diff
@@ -0,0 +1,107 @@
...
+4. Configure a robot account  `quay.io/unique-org-david` to be able to push to `quay.io/unique-org-david/unique-component-foo`
```
gorkem, line 79: Is the token only for push? Or push and pull?

ADR/0004-component-image-location.md
```diff
@@ -0,0 +1,107 @@
...
+* Security:
+  * Leverage short-lived Quay.io API tokens for repo/org management on Quay.io
+  * Leverage robot accounts for pushing images from CI/Build pipelines.
+
+
+
+## Decision
+
+### What
+
```
gorkem, line 51: Can we also discuss here what happens to the repo/images when the workspace or user is deleted?

ADR/0004-component-image-location.md
sbose78: Push and Pull.

ADR/0004-component-image-location.md
sbose78: sounds good 👍 "private" is anyway the 'default' in quay.io . For private repositories, we'd have to propagate the pull secret to the deployment environment.

ADR/0004-component-image-location.md
```diff
@@ -0,0 +1,107 @@
...
+* Per workspace, setup a new org “quay.io/unique-org-david/…”
```
ralphbean, line 52:
```suggestion
* Per workspace called "david", setup a new org “quay.io/unique-org-david/…”
```

ADR/0004-component-image-location.md
```diff
@@ -0,0 +1,107 @@
...
+* Create a new Quay.io org owned by David using David’s API token “quay.io/unique-org-david”
+* Create a new image repo for “foo” using David’s API token “quay.io/unique-org-david/appname-foo”
+* Optionally, if needed, make the image repo public using the David’s API token for easy image pulls.
+* Configure the robot account “appstudio” in “quay.io/unique-org-david” to be able to push to “quay.io/unique-org-david/appname-foo” Invalidate/Delete/Prune David’s API token. 
+
+
+#### Short-term
+
+##### Setup
+* Configure the a pre-configured Quay.io API token for repo/org management, also known as the OAuth Access token. From a security perspective, this would be treated as a typical service account token
```
ralphbean, line 72:
```suggestion
* Configure a pre-configured Quay.io API token for repo/org management, also known as the OAuth Access token. From a security perspective, this would be treated as a typical service account token
```

ADR/0004-component-image-location.md
```diff
@@ -0,0 +1,107 @@
...
+  belonging to AppStudio. This token would be shortlived and rotated *as frequently as practically possible*  .
```
ralphbean, line 73: Can we say which AppStudio service this belongs to? Does it belong specifically to `build-service`?

ADR/0004-component-image-location.md
```diff
@@ -0,0 +1,107 @@
...
+We would use the user's API token for working with the relevant Quay.io organizations and repositories.
+
+
+## Open Questions
+
+- Since the long-term goal is to have the Quay.io organizations owned by the user, how do we build a frictionless experience to map the user's 
+  account with the user's Quay.io account ?
+- Considering the above is figured out, we would need to add existing users as owners to the relevant organizations. This is a backend job that 
+  would need to be designed and executed at an appropriate time.
+
```
ralphbean, line 94:
```suggestion
- In the HACBS context, giving ownership of the image repo to users is not a problem for provenance, because integrity of images is based on attestations which are only treated as valid if signed by pipeline-service (by tekton chains). While users can in theory push images built on their laptop, the [release-service](../book/release-service.md) won't treat them as if they were built by the [pipeline-service](https://github.com/openshift-pipelines/pipeline-service).
```

ADR/0004-component-image-location.md
ralphbean: Do I understand it right?

ADR/0004-component-image-location.md
```diff
@@ -0,0 +1,107 @@
...
+* Alexey Kazakov
+* Gorkem Ercan
+
+## Reviewers
+
+* Alexey Kazakov
+* Gorkem Ercan
+* Parag Dave
+
```
ralphbean, line 19:
```suggestion
* Ralph Bean
```

ADR/0004-component-image-location.md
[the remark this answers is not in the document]
sbose78: 👍 I wanted to convey that a unique org name is generated, I don't quite have an opinion on the generation function.

sbose78: Correct.

ADR/0004-component-image-location.md
sbose78: I'm not really sure since build service is a mere consumer of this ?

ADR/0004-component-image-location.md
```diff
@@ -0,0 +1,107 @@
...
+#### Short-term
+
+##### Setup
+* Configure the a pre-configured Quay.io API token for repo/org management, also known as the OAuth Access token. From a security perspective, this would be treated as a typical service account token
+  belonging to AppStudio. This token would be shortlived and rotated *as frequently as practically possible*  .
+
+##### When a user creates a Component
+1. Create a new Quay.io org owned by David using the pre-configured AppStudio-wide API token `quay.io/unique-org-david`
+2. Create a new image repo for Component “foo” using the the pre-configured David’s API token `quay.io/unique-org-david/unique-component-foo`
```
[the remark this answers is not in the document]
sbose78: The UI should trigger the oauth flow with Quay.io using SPI to get us the API token.

ADR/0004-component-image-location.md
```diff
@@ -0,0 +1,105 @@
...
+Date: Oct 29, 2022 
+
+## Status
+
+Accepted
+
+## Approvers
+
+* Alexey Kazakov
+* Gorkem Ercan
```
amfred, line 12: I want to be an approver on this because it's a security requirement

amfred (changes requested):
I think this proposed design will meet our isolation requirements, but there are a few open issues to resolve in the comments.

ADR/0004-component-image-location.md
sbose78: Sure @amfred

amfred:
To me, the big open question is: which service is going to maintain the lifecycle of these Quay repos? We need to clean them up when we close accounts.

sbose78:
@amfred , I'll update the ADR addressing that. We'll most likely have a delayed approach to deleting the Quay.io repositories to let users move their images to a different location in case they were depending on them for deployments.

sbose78:
Update: In the initial phase, we'll have all repos under a single org. Once the concept of workspaces are better fleshed out and implemented, we'll move to one org per workspace. TODO: Confirm from the Quay.io team if programmatic creation of Quay.io orgs could be enabled.

sbose78:
> TODO: Confirm from the Quay.io team if programmatic creation of Quay.io orgs could be enabled.

Confirmed, this is implementable. Either way, I've defined what our iterative approach towards implementing this should be.

> It's definitely one of our data retention policies. I put a proposal in our record retention policy (which we may want to publish here). My proposal is that we would disable accounts for 30 days after the subscription expires 

+1, I added that to the ADR as a "delayed" deletion.


> I think we should delete the images from our internal image registry at the same time as deleting the other resources (deployments, application configuration, etc.). 

+1 , added that to the ADR

> Some service is going to have to manage the lifecycle of this.

Yeah, the initial goal of this ADR wasn't to get into the details of which service would implement it - rather what would the service need to do. Over time, I can confirm that I have some clarity that this needs to be an independent controller given that pull secret is of interest for both build and deploy scenarios.

arewm:
Asking about this after reading the comments but before reading the ADR itself.
>  We'll most likely have a delayed approach to deleting the Quay.io repositories to let users move their images to a different location in case they were depending on them for deployments.

Would any pull tokens generated (i.e. for private image repos) be deleted when the account/workspace is deleted? Would there be a way for a user to reactivate their account to reclaim the quay repos/organization if within the cleaning deadline?

ADR/0004-component-image-location.md
```diff
@@ -0,0 +1,187 @@
...
+```
+    
+    
+
+
+## Open Questions
+
+- What would be a progammatic way to regenerate the main Quay.io API token ?
+- Since the long-term goal is to have the Quay.io organizations owned by the user, how do we build a frictionless experience to map the user's 
+  account with the user's Quay.io account ?
```
arewm, line 170: Is there a way to transition between the types of quay ownership -- i.e. being able to hand over "keys" to an organization for an additional cost?

ADR/0004-component-image-location.md
```diff
@@ -0,0 +1,187 @@
...
+- Considering the above is figured out, we would need to add existing users as members of the relevant organizations. This is a backend job that 
+  would need to be designed and executed at an appropriate time.
+- In the HACBS context, giving ownership of the image repo to users is not a problem for provenance, because integrity of images is based on attestations which are only treated as valid if signed by pipeline-service (by tekton chains). While users can in theory push images built on their laptop, the [release-service](../book/release-service.md) won't treat them as if they were built by the [pipeline-service](https://github.com/openshift-pipelines/pipeline-service). They will fail to pass the [enterprise contract](../book/enterprise-contract.md).
```
arewm, line 173: Do we need to reword this as there isn't a really HACBS context anymore?

ADR/0004-component-image-location.md
```diff
@@ -0,0 +1,187 @@
...
+2. Create an OAuth Application in the Quay.io organization.
+3. Geneate a token for the OAuth Application. This token would act as the 'service account' using which Quay.io resources would be created. Important to note, the token acts on behalf of the user who is requesting it - but uses the explicit scopes specified at the time of token generation. 
+4. Allowlist user 'shbose' to be create organizations using non-user-tokens using the Quay.io API.
+
+| Syntax      | Description |
+| ----------- | ----------- |
+| Quay.io organization      | quay.io/redhat-user-workloads       |
+| OAuth Application  name | Created, name redacted        |
+| Account used to generate token | `shbose` |
+| Scope | Administer organizations, Adminster repositories, Create Repositories |
```
arewm, line 73: Can we align permissions granted to users for quay organizations/repositories to the Stonesoup roles as defined in https://github.com/redhat-appstudio/book/pull/42/files?

ADR/0004-component-image-location.md
```diff
@@ -0,0 +1,187 @@
...
+* Token generation: 
+    * Robot account token: At the moment, the controller responsible for generating the Quay.io resources would be responsible for rotating the tokens https://docs.quay.io/api/swagger/#!/robot/regenerateUserRobotToken
+    * Quay.io API token: No programmatic way to regenerate this token is known at this point of time. This would be a manual activity to begin with. 
+
+* Upon deletion of an `Application`/`Component` from StoneSoup, 
+    * The controller/finalizer would delete the the relevant Quay.io resources namely, the image repository and the robot account.
+    * The controller/finalizer would delete the linked `Secret` from the user's namespace. Most likely, this should be a mere `ownerReference`-based garbage collection.
+
+* Upon removal of a user from Stonesoup,
+    * The empty Quay.io organization associated with the user or the user's Space may not be deleted instantly, but would be scheduled for a delayed cleanup.
```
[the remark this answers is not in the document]
sbose78: >  Preventing others from being able to squat on it after cleanup?

Yes, this is what I had in mind.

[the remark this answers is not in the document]
arewm: Gotcha, I wasn't sure if you had any other reason in mind. It doesn't seem like a necessary feature to protect against the quay repo itself (and potentially introduce unnecessary complexity). I imagine that it would be good to have a "delete process" where all the content which will be deleted is itemized so that the user can make an informed decision.

I also assume that any creation of organizations will be able to handle "fuzzing" to resolve name conflicts so there wouldn't be an issue if someone did manage to squat on an organization..

Is there any planned differentiation between empty and nonempty quay organizations?

ADR/0004-component-image-location.md
```diff
@@ -0,0 +1,187 @@
...
+
+* Alexey Kazakov
+* Gorkem Ercan
+* Parag Dave
+* Ralph Bean
+
+## Context
+
+**Problem**
+StoneSoup does not have a internal registry where images could be pushed to as an intermediate step before being deployed as a container.
```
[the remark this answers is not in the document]
arewm: Well, now change that to `AppStudio`. :)

ADR/0004-component-image-location.md
sbose78: That would be a customer support ticket I guess :-)

ADR/0004-component-image-location.md
```diff
@@ -0,0 +1,187 @@
...
+* The quay.io repository created with use the format   quay.io/org-name/namespace-name/application-name/component-name. Example, https://quay.io/repository/redhat-user-workloads/shbose/applicationname/automation-repo
+
+#### Lifecycle of the Quay.io resources
+
+* Token generation: 
+    * Robot account token: At the moment, the controller responsible for generating the Quay.io resources would be responsible for rotating the tokens https://docs.quay.io/api/swagger/#!/robot/regenerateUserRobotToken
+    * Quay.io API token: No programmatic way to regenerate this token is known at this point of time. This would be a manual activity to begin with. 
+
+* Upon deletion of an `Application`/`Component` from StoneSoup, 
+    * The controller/finalizer would delete the the relevant Quay.io resources namely, the image repository and the robot account.
```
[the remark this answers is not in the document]
arewm: Revisiting --

If there was a deletion delay to enable the repository to be re-hooked up to, would it even be _possible_ to create a new component which will push to the same location? Based on my current understanding, this isn't really possible without a new feature to "restore a recently deleted component." Therefore, this functionality would also be relegated to its own feature/request if desired.

> Could we support component archiving where artifacts remain but all configuration is unhooked?

Component images that have not been mirrored to some other location (i.e. via release pipelines) can be treated as "developer" images in that they are not guaranteed to remain for a long point in time. While these images can be treated a deletable (the behavior MUST be documented), an archival state seems like it could be useful as well -- but this would be a different feature/request.

> What if we have multiple components being pushed to the same repository but different versions (i.e. different branches)?

This still seems like a potentially valid use case to me. Do we afford users with enough customization and flexibility to push multiple components to one repository? I feel like we _should_ be able to afford this customization. Based on the current flow, I assume not ... so again, it would have to be a new feature request.

> I assume that this would have to vary if we are pushing to a custom quay organization.

Likely so, but this is far outside of the current support that it seems fine to defer this question until (and if) we decide to get there

ADR/0004-component-image-location.md
Michkov: @sbose78 are you going to add @amfred to fulfill this request?

ADR/0004-component-image-location.md
```diff
@@ -0,0 +1,190 @@
...
+
+**Non-Goals**: 
+* Define the user experience for users bringing existing images from other image registry services.
+* Provide users an option to choose what the out-of-the-box location for images would be. 
+* Define the user experience for users who are willing to configure additional credentials for pushing to an image registry of their choice.
+
+**Design Goals**
+* Use Quay.io as the out-of-the-box image registry in order to avoid getting into the business of maintaining an internal registry.
+* Align user permissions in an StoneSoup workspace with those in Quay.io
+* Maintain the right levels of isolation between images being expected of a multi-tenant system.
```
tkdchen, line 43: What are the levels of isolation? Is isolation level between images explained anywhere else in any AppStudio document?

ADR/0004-component-image-location.md
```diff
@@ -0,0 +1,190 @@
...
+
+
+**Non-Goals**: 
+* Define the user experience for users bringing existing images from other image registry services.
+* Provide users an option to choose what the out-of-the-box location for images would be. 
+* Define the user experience for users who are willing to configure additional credentials for pushing to an image registry of their choice.
+
+**Design Goals**
+* Use Quay.io as the out-of-the-box image registry in order to avoid getting into the business of maintaining an internal registry.
+* Align user permissions in an StoneSoup workspace with those in Quay.io
```
tkdchen, line 42: What does it mean by "those"?

ADR/0004-component-image-location.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Decision
+
+### What
+
+* Per workspace called "david", setup a new org “quay.io/unique-org-david/…”
+* Per component, setup a new new repo “quay.io/unique-org-david/unique-component-repo”
+* Use User’s Quay.io API token to manage the org/repo. Short-term, we'll use a pre-configured Quay.io API token associated with StoneSoup to create the org/repo till we 
+ figure out how to determinstically map a user in StoneSoup to a user in Quay.io.
+* Generate a robot account token scoped to the relevant repository and persist it in the user's workspace for the image build and push process to consume.
```
tkdchen, line 58: Both "user's workspace" and "user's namespace" are used through this document. Do they mean the same thing? If yes, could we use single term consistently? If no, could we have a separate section to explain them as well as other terms used in this document?

ADR/0004-component-image-location.md
```diff
@@ -0,0 +1,190 @@
...
+    * Robot account token: At the moment, the controller responsible for generating the Quay.io resources would be responsible for rotating the tokens https://docs.quay.io/api/swagger/#!/robot/regenerateUserRobotToken
+    * Quay.io API token: No programmatic way to regenerate this token is known at this point of time. This would be a manual activity to begin with. 
+
+* Upon deletion of an `Application`/`Component` from StoneSoup, 
+    * The controller/finalizer would delete the the relevant Quay.io resources namely, the image repository and the robot account.
+    * The controller/finalizer would delete the linked `Secret` from the user's namespace. Most likely, this should be a mere `ownerReference`-based garbage collection.
+
+* Upon removal of a user from Stonesoup,
+    * The empty Quay.io organization associated with the user or the user's Space may not be deleted instantly, but would be scheduled for a delayed cleanup.
+* PR-based tags are to be deleted on a regular basis. Image tags associated with `main` may remain un-pruned for now.
```
tkdchen, line 112: Would it be good to replace `main` with "the merge target branch"?

sbose78:
Took this, I will address the remaining comments as a follow-up 👍 (appreciate you taking out time to review)

PR #24, opened 2022-11-29
ADR-0008 Test workstream - API contracts

hongweiliu17:
This PR comes from [HACBS Test workstream - API contracts doc](https://docs.google.com/document/d/1FJX7TH5wBTcWdU2VYuecmkY5O69DGpItN52A-JaoCtU/edit#heading=h.couaoxve6mi9)

https://github.com/hongweiliu17/book/blob/HACBS-1425/ADR/0008-integration-service-api-contracts.md can be referred to review this PR.

### DIFF
--- a/ADR/0008-integration-service-api-contracts.md
+++ b/ADR/0008-integration-service-api-contracts.md
@@ -1,4 +1,4 @@
-# 8. StoneSoup Test Stream - API contracts
+# 8. Stonesoup Test Stream - API contracts
 
 Date: 2023-01-[]
 
@@ -8,7 +8,7 @@ Accepted
 
 ## Context
 
-The Hybrid Application Cloud Build Service being developed aims to serve Red Hat teams but also partners and customers. This requires a level of adaptability to avoid recreating custom flows and Tasks for each stakeholder.
+The Stonesoup project being developed aims to serve Red Hat teams but also partners and customers. This requires a level of adaptability to avoid recreating custom flows and Tasks for each stakeholder.
 
 In this respect Tasks developed by StoneSoup test stream should allow swapping external systems to accommodate different environments. This swap should not induce the complete recreation of pipelines.
 
@@ -28,13 +28,11 @@ Related to [PLNSRVCE-41](https://issues.redhat.com/browse/PLNSRVCE-41) investiga
 
 ## Decision
 
-### Results and Records
+Related to [PLNSRVCE-41](https://issues.redhat.com/browse/PLNSRVCE-41) investigations in regards to [Tekton results](https://github.com/tektoncd/results). The investigations are documented [here](https://docs.google.com/document/d/1WGWx6MllVs-BwRLd0PP5QTQHgGp0fW2dOYRKRAKWiwo/edit#heading=h.bpyxfhyud4dv). The output of each Tekton task will be provided in two forms: **Tekton Task Results** and **Full Test Output JSON**.
 
-The output of each Tekton task will be provided in two forms:
-- As a minimized [Tekton result](https://tekton.dev/docs/pipelines/tasks/#emitting-results) in JSON format listing all test failures
-  - The name of the result will be **HACBS_TEST_OUTPUT**
-- As a Test output JSON file with full test information that’s saved in the Tekton Pipeline Workspace
-  - The name of the file will be in snake case and will be in the form of **test_name_output.json**
+### Tekton Task Results
+
+The output of each Tekton task will be provided in a minimized [Tekton result](https://tekton.dev/docs/pipelines/tasks/#emitting-results) in JSON format listing all test failures. The name of the result will be **HACBS_TEST_OUTPUT**.
 
 The maximum size of a [Task's Results](https://tekton.dev/vault/pipelines-v0.17.3/tasks/#emitting-results) is limited by the container [termination message](https://kubernetes.io/docs/tasks/debug/debug-application/determine-reason-pod-failure/#customizing-the-termination-message) feature of Kubernetes.
 
@@ -49,7 +47,7 @@ The output will provide the following information about the overall test result:
 - **namespace** - Optional, the rego namespace of the test policy
 - **timestamp** - An UNIX epoch timestamp of the test completion time
 - **successes** - The number of successful checks in the form of an integer
-- **note** - Optional, a short note providing additional information about the test
+- **note** - Optional, a short note provided by test workstream to provide additional information about the test
 - **failures** - The number of failed checks in the form of an integer
 - **warnings** - The number of warning checks in the form of an integer
 
@@ -60,8 +58,9 @@ Example contents of the test result output file (**HACBS_TEST_OUTPUT**) for a fa
     "namespace": "image_labels",
     "timestamp": "1649148140",
     "successes": 12, 
-    "note": "",
-    "failures": 2
+    "note": "Please check tekton task log for more detailed failures",
+    "failures": 2,
+    "warnings": 0
 }
 ```
 Example for a successful run:
@@ -71,67 +70,35 @@ Example for a successful run:
     "timestamp": "1649843611",
     "namespace": "required_checks",
     "successes": 16,
-    "note": "",
-    "failures": 0
+    "failures": 0,
+    "warnings": 0
 }
 ```
 Example for a skipped run:
 ```
 {
     "result": "SKIPPED",
-    "note": "<skipped message>",
-    "timestamp": "1649842004"
+    "note": "We found 0 supported files",
+    "timestamp": "1649842004",
+    "successes": 0,
+    "failures": 0,
+    "warnings": 0
 }
 ```
 Example for a run with an error:
 ```
 {
     "result": "ERROR",
-    "timestamp": "1649842004"
+    "timestamp": "1649842004",
+    "successes": 0,
+    "failures": 0,
+    "warnings": 0
 }
 ```
 
-##### Proposed Adapted Tekton Result Format for Xunit Tests Results
+### Detailed Conftest Output JSON
 
-- Existing Mapped:
-
-  - **Namespace** == **name** attribute of all of the **TestSuite** elements 
-
-  - **Failures** == **name** attribute for all the **TestCase** elements that have the attribute **status=Failed**
-
-- New Key:
-
-  - **Skipped** == the number of skipped test in integer format from the **TestSuite** attribute **skipped** or from **TestCase** attribute **skipped**
-
-Some Junit XML files generate a **TestSuites** xml element that contains N number of **TestSuite** elements. In this case it maybe simply best to augment our Task Result Schema to be a JSON List of Task Results 
-
-Example output
-```
-[
-    {
-        "result": "FAILURE",
-        "namespace": "Red Hat App Studio E2E tests",
-        "timestamp": "1649148140",
-        "successes": 75,
-        "skipped": 14,
-        "note": "",
-        "failures": [
-            "[It] [release-suite test-demo] Creation of the 'Happy path' resources Create an ApplicationSnapshot."
-        ]
-    },
-    {
-        "result": "SUCCESS",
-        "timestamp": "1649843611",
-        "namespace": "Demo Tests",
-        "successes": 12,
-        "skipped": 0,
-        "note": "",
-        "failures": []
-    }
-]
-```
-
-#### Full Test Output JSON
+Test output JSON file with detailed test information is saved in the Tekton Pipeline Workspace. The name of the file will be in snake case and will be in the form of **test_name_output.json**
 
 The Test Tekton tasks will use a standardized manner of displaying testing information. This information will be saved in the form of a JSON file. The contents of this file will relay results of validation by the [Open Policy Agent](https://www.openpolicyagent.org/) policies as executed by the [Conftest](https://www.conftest.dev/) tool and saved in its JSON output format.
 
@@ -150,7 +117,7 @@ The **failures** list of objects will provide the following information about ea
 - **description** - The explanation about why the check is significant (Why the label is used, why the vulnerabilities are checked etc.)
 - **url** - link to the further documentation on the individual check (or that type of check)
 
-Example full JSON test output for a Tekton task that tests the container image labels:
+Example detailed JSON test output for a Tekton task that tests the container image labels:
 ```
 [
     {
@@ -173,10 +140,20 @@ Example full JSON test output for a Tekton task that tests the container image l
 ]
 ```
 
+```
+[
+    {
+        "filename": "image_inspect.json",
+        "namespace": "fbc_checks",
+        "successes": 1
+    }
+]
+```
+
 ### Information injection
 
 Whenever possible, resources like ConfigMaps or Secrets will be used to inject configuration into Tasks. This is preferred to templates and patches as it fits well with Kubernetes declarative and GitOps approaches.
-tf
+
 ConfigMaps and Secrets will be mounted into the Task pod to inject file-based information like certificates.
 Environment variables may be injected from ConfigMaps.
 
@@ -193,31 +170,14 @@ Atomic information may be passed as a simple parameter. Non binary encoded compl
 
 #### Image references
 
-Tags are not uniquely identifying an image so that we cannot rely on them. They do not guarantee that the scanning is for the image built as part of the PipelineRun if there is a race with a parallel build. As such, image digests will be used for image references.
-
-#### Logging
-
-In containers, logs are usually written to the standard output and collected by the container engine. Then OpenShift Fluentd, Splunk agent or any other tool can forward them to a central aggregator. Using [structured logging](https://docs.openshift.com/container-platform/4.9/logging/cluster-logging-enabling-json-logging.html) with a common format would provide additional value.
-
-Common fields (draft):
-- Timestamp, for instance: "2022-02-20T00:28:47.125906354Z"
-- Workspace, for instance: "user-1"
-- Pipeline, for instance: "test"
-- PipelineRun, for instance: "test-wc05gl"
-- Task, for instance: "img-scan"
-- TaskRun, for instance: "img-scan-er02es"
-- CompName, for instance: "hco-bundle-registry-container"
-- CompVers, for instance: "v4.10.0"
-- CompRel, for instance: "709"
-
-Today NVR is largely used for tracking, for instance hco-bundle-registry-container-v4.10.0-709.
-
-The idea is that the common fields listed above would be populated for each log entry when the information is available.
-
-It is advisable to limit the number of structured fields as it impacts indexing. Fields that are common to multiple tasks should be preferred. Before a field specific to a task is added it is worth thinking whether it is really needed and whether the same information can be conveyed through a more generic field. Information that does not need to be “searchable” can be recorded in an unstructured way.
+Since tags can be moved from one image to another, they should not be relied on as a reference. In order to guarantee that any scanning is performed on an image built as part of a PipelineRun, the immutable image digest reference will be used instead.
 
 ## Consequences
 
+As a result of the decision here to summarize results in a **HACBS_TEST_OUTPUT** result and store the larger test output as a file named **test_name_output.json**, we should find that:
+* Other components in Stonesoup can leverage information exposed by TaskRuns - notably the UI (HAC), integration-service, and enterprise-contract - enabling features for the larger system that need to depend on some data from inside a variety of TaskRuns.
+* We'll be able to have PipelineRuns that _succeed_ and continue, even if they have tasks whose payloads fail and expose errors. This enables a progressive model where the user can get a build and a functional test and a deployment to their development Environment, even if a linter or scanner in their build pipeline emits an error.
+* By having chosen a convention that has "HACBS" in the name, we're going to have trouble integrating third-party Task providers in the future. In order to have its output respected by our system, a hypothetical vendor of a third-party scanner will need to add a "HACBS_TEST_OUTPUT" result on their Task, which is oddly specific to our system. At some point in the future, we should revise this decision to instead align to a common upstream convention that gains traction in the broader tekton ecosystem. See also [HACBS-1563](https://issues.redhat.com/browse/HACBS-1563).
 
 ## Additional Recommendations
 
--- a/ADR/0008-integration-service-api-contracts.md
+++ b/ADR/0008-integration-service-api-contracts.md
@@ -1,0 +1,256 @@
+# 8. HACBS Test Stream - API contracts
+
+Date: 2022-11-28
+
+## Status
+
+Accepted
+
+## Context
+
+The Hybrid Application Cloud Build Service being developed aims to serve Red Hat teams but also partners and customers. This requires a level of adaptability to avoid recreating custom flows and Tasks for each stakeholder.
+
+In this respect Tasks developed by HACBS test stream should allow swapping external systems to accommodate different environments. This swap should not induce the complete recreation of pipelines.
+
+This and the idea of providing a homogeneous experience, which is easier to comprehend and navigate complex systems, leads to the definition of `API contracts`. These contracts need to be understood as guidance that may evolve with time and experience while keeping the aim of building a flexible homogeneous system.
+
+Tasks from the test stream may exchange information through
+
+1. Parameters passed at the time of the trigger of a PipelineRun (that is user, git repository, etc)
+2. Output of upstream elements of the Pipeline which may be needed for performing the Task (that is build artifacts, container image, etc)
+3. Providing whether the Task was successful or not
+4. Output of the Task that may be needed for processing subsequent Tasks of the pipeline
+5. Output of the Task that need to be stored for auditing and troubleshooting (that is summarized results of a validation or a scan, etc)
+6. Communication with external systems that are needed for performing the Task (that is container image repository, image scanner, etc)
+7. Connection details and credentials that may be required for the above
+
+Related to [PLNSRVCE-41](https://issues.redhat.com/browse/PLNSRVCE-41) investigations in regards to [Tekton results](https://github.com/tektoncd/results). The investigations are documented [here](https://docs.google.com/document/d/1WGWx6MllVs-BwRLd0PP5QTQHgGp0fW2dOYRKRAKWiwo/edit#heading=h.bpyxfhyud4dv).
+
+## Decision
+
+### Results and Records
+
+The output of each Tekton task will be provided in two forms:
+- As a minimized [Tekton result](https://tekton.dev/docs/pipelines/tasks/#emitting-results) in JSON format listing all test failures
+  - The name of the result will be **HACBS_TEST_OUTPUT**
+- As a Test output JSON file with full test information that’s saved in the Tekton Pipeline Workspace
+  - The name of the file will be in snake case and will be in the form of **test_name_output.json**
+
+The maximum size of a [Task's Results](https://tekton.dev/vault/pipelines-v0.17.3/tasks/#emitting-results) is limited by the container [termination message](https://kubernetes.io/docs/tasks/debug/debug-application/determine-reason-pod-failure/#customizing-the-termination-message) feature of Kubernetes.
+
+App Studio builds are structured as [a shared Persistent Volume per App Studio Workspace](https://docs.google.com/document/d/1IPlihVjkJ4Kb9tdhsk7iz3bn5rkT_SvJCNQhyXzK3aI/edit#bookmark=id.gefgys3vno2). This allows teams to share builds, implement caching and other shared volumes. A single persistent volume is mapped to each default build pipeline. Builds are passed a directory specific to their builds.
+
+#### Tekton Result Format
+
+The Test output of the Tekton result will be a JSON object that includes the context about the test along with the list of check names for all failed checks
+
+The output will provide the following information about the overall test result:
+- **result** - The outcome of the testing task, can be `SUCCESS`, `FAILURE`, `WARNING`, `SKIPPED` or `ERROR`
+- **namespace** - Optional, the rego namespace of the test policy
+- **timestamp** - An UNIX epoch timestamp of the test completion time
+- **successes** - The number of successful checks in the form of an integer
+- **note** - Optional, a short note providing additional information about the test
+- **failures** - The number of failed checks in the form of an integer
+- **warnings** - The number of warning checks in the form of an integer
+
+Example contents of the test result output file (**HACBS_TEST_OUTPUT**) for a failed run:
+```
+{ 
+    "result": "FAILURE",
+    "namespace": "image_labels",
+    "timestamp": "1649148140",
+    "successes": 12, 
+    "note": "",
+    "failures": 2
+}
+```
+Example for a successful run:
+```
+{
+    "result": "SUCCESS",
+    "timestamp": "1649843611",
+    "namespace": "required_checks",
+    "successes": 16,
+    "note": "",
+    "failures": 0
+}
+```
+Example for a skipped run:
+```
+{
+    "result": "SKIPPED",
+    "note": "<skipped message>",
+    "timestamp": "1649842004"
+}
+```
+Example for a run with an error:
+```
+{
+    "result": "ERROR",
+    "timestamp": "1649842004"
+}
+```
+
+##### Proposed Adapted Tekton Result Format for Xunit Tests Results
+
+- Existing Mapped:
+
+  - **Namespace** == **name** attribute of all of the **TestSuite** elements 
+
+  - **Failures** == **name** attribute for all the **TestCase** elements that have the attribute **status=Failed**
+
+- New Key:
+
+  - **Skipped** == the number of skipped test in integer format from the **TestSuite** attribute **skipped** or from **TestCase** attribute **skipped**
+
+Some Junit XML files generate a **TestSuites** xml element that contains N number of **TestSuite** elements. In this case it maybe simply best to augment our Task Result Schema to be a JSON List of Task Results 
+
+Example output
+```
+[
+    {
+        "result": "FAILURE",
+        "namespace": "Red Hat App Studio E2E tests",
+        "timestamp": "1649148140",
+        "successes": 75,
+        "skipped": 14,
+        "note": "",
+        "failures": [
+            "[It] [release-suite test-demo] Creation of the 'Happy path' resources Create an ApplicationSnapshot."
+        ]
+    },
+    {
+        "result": "SUCCESS",
+        "timestamp": "1649843611",
+        "namespace": "Demo Tests",
+        "successes": 12,
+        "skipped": 0,
+        "note": "",
+        "failures": []
+    }
+]
+```
+
+#### Full Test Output JSON
+
+The Test Tekton tasks will use a standardized manner of displaying testing information. This information will be saved in the form of a JSON file. The contents of this file will relay results of validation by the [Open Policy Agent](https://www.openpolicyagent.org/) policies as executed by the [Conftest](https://www.conftest.dev/) tool and saved in its JSON output format.
+
+Each test will have a standardized short name in snake case, e.g. `release_label_required`, `architecture_label_required`, `architecture_label_deprecated` etc.
+
+The output will provide the following information about the overall test result:
+- **filename** - Name of the file that was inspected by the Tekton task
+  - This will reflect the data that the test was executed on, such as `image-inspect.json`, `clair-vulnerabilities.json` etc.
+- **namespace** - The rego namespace of the test policy. 
+- **successes** - The number of successful checks in the form of an integer
+- **failures** - A JSON list containing objects describing each failure
+
+The **failures** list of objects will provide the following information about each check:
+- **name** - Name of the individual check written in [snake case](https://en.wikipedia.org/wiki/Snake_case)
+- **msg** - Message about what went wrong with the check
+- **description** - The explanation about why the check is significant (Why the label is used, why the vulnerabilities are checked etc.)
+- **url** - link to the further documentation on the individual check (or that type of check)
+
+Example full JSON test output for a Tekton task that tests the container image labels:
+```
+[
+    {
+        "filename": "image-inspect.json",
+        "namespace": "image_labels",
+        "successes": 19,
+        "failures": [
+            {
+                "msg": "The 'architecture' label is required",
+                "metadata": {
+                    "details": {
+                        "description": "Architecture the software in the image should target.",
+                        "name": "architecture_label_required",
+                        "url": "https://source.redhat.com/groups/public/container-build-system/container_build_system_wiki/guide_to_layered_image_build_service_osbs#jive_content_id_Labels"
+                    }
+                }
+            }
+          ]
+    }
+]
+```
+
+### Information injection
+
+Whenever possible, resources like ConfigMaps or Secrets will be used to inject configuration into Tasks. This is preferred to templates and patches as it fits well with Kubernetes declarative and GitOps approaches.
+tf
+ConfigMaps and Secrets will be mounted into the Task pod to inject file-based information like certificates.
+Environment variables may be injected from ConfigMaps.
+
+Variables directly configured in the Pod definition are discouraged.
+
+Side note: Secrets and ConfigMaps should be made [immutable](https://kubernetes.io/docs/concepts/configuration/secret/#secret-immutable)
+
+
+### Information format
+
+[Clearly define the format of input parameters and results](https://github.com/tektoncd/catalog/blob/main/recommendations.md#clearly-define-the-format-of-input-parameters-and-results).
+
+Atomic information may be passed as a simple parameter. Non binary encoded complex information should be exchanged through [JSON format](https://github.com/tektoncd/catalog/blob/main/recommendations.md#use-composable-parameter-formats) between Tasks.
+
+#### Image references
+
+Tags are not uniquely identifying an image so that we cannot rely on them. They do not guarantee that the scanning is for the image built as part of the PipelineRun if there is a race with a parallel build. As such, image digests will be used for image references.
+
+#### Logging
+
+In containers, logs are usually written to the standard output and collected by the container engine. Then OpenShift Fluentd, Splunk agent or any other tool can forward them to a central aggregator. Using [structured logging](https://docs.openshift.com/container-platform/4.9/logging/cluster-logging-enabling-json-logging.html) with a common format would provide additional value.
+
+Common fields (draft):
+- Timestamp, for instance: "2022-02-20T00:28:47.125906354Z"
+- Workspace, for instance: "user-1"
+- Pipeline, for instance: "test"
+- PipelineRun, for instance: "test-wc05gl"
+- Task, for instance: "img-scan"
+- TaskRun, for instance: "img-scan-er02es"
+- CompName, for instance: "hco-bundle-registry-container"
+- CompVers, for instance: "v4.10.0"
+- CompRel, for instance: "709"
+
+Today NVR is largely used for tracking, for instance hco-bundle-registry-container-v4.10.0-709.
+
+The idea is that the common fields listed above would be populated for each log entry when the information is available.
+
+It is advisable to limit the number of structured fields as it impacts indexing. Fields that are common to multiple tasks should be preferred. Before a field specific to a task is added it is worth thinking whether it is really needed and whether the same information can be conveyed through a more generic field. Information that does not need to be “searchable” can be recorded in an unstructured way.
+
+## Consequences
+
+
+## Additional Recommendations
+
+### Convention over Configuration
+https://en.wikipedia.org/wiki/Convention_over_configuration
+
+Tasks should assume default locations for locating information, external systems or output storage. This aims to reduce the amount of information that needs to be configured for running a pipeline.
+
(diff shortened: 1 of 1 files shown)


ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,277 @@
...
+7. Connection details and credentials that may be required for the above
+
+## Decision
+
+### Convention over Configuration
+https://en.wikipedia.org/wiki/Convention_over_configuration
+
+Tasks should assume default locations for locating information, external systems or output storage. This aims to reduce the amount of information that needs to be configured for running a pipeline.
+
+TODO: Evaluate for the first tasks the information, which is required, and document here the defaults:
```
gbenhaim, line 36: I think that `TODOs` shouldn't commit to git. Usually, they just sit there and no one is doing anything with them.
Work that should be done needs to be in the issue tracker, in our case Jira.

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,277 @@
...
+- Environment variables: Following [Kubernetes](https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/) and bash practices environment variables should be defined in UPPER_CASE
+- Labels: Labels names will follow Kubernetes practices: prefix with dot `.` as separator and dash-case for names. Prefixes and names are separated by a slash `/`. In addition to the [labels automatically created by Tekton Kubernetes standard labels or custom labels](https://tekton.dev/vault/pipelines-v0.14.3/labels/#automatically-added-labels) could be leveraged to further qualify the resources. Possible groupings (TODO: if desired, define common names):
+  - Pipeline part: Build, Test, Release, etc.
+  - Application/Product being built: AMQ Streams, EAP, RHSSO, etc.
+
+### DRY
+
+Information, which relates to an environment like (7) connection details and credentials should be configured once at the environment level and not passed as parameters for every PipelineRun. On the other hand, information, which is specific to a run like a container image digest, may be passed as a parameter.
```
gbenhaim, line 58: What does it mean `configured ag the environment level` how the pipeline run will use those secrets/configmaps?

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,277 @@
...
+
+### Information injection
+
+Whenever possible, resources like ConfigMaps or Secrets will be used to inject configuration into Tasks. This is preferred to templates and patches as it fits well with Kubernetes declarative and GitOps approaches.
```
gbenhaim, line 62: When it comes to `ConfigMaps` how it will work with pipeline as code? I'm interested in the pre-merge case.
The problem is that in order to test changes to the ConfigMap, one should first apply them on the cluster, so we can test them in the CI.

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,277 @@
...
+- Task Parameters
+- Task Results
+- Tekton Results
+- PipelineResources (resources input / output)  are deprecated: don’t use
+
+
+The maximum size of a Task's Results is limited by the container termination message feature of Kubernetes.
+
+
+App Studio builds are structured as [a shared Persistent Volume per App Studio Workspace](https://docs.google.com/document/d/1IPlihVjkJ4Kb9tdhsk7iz3bn5rkT_SvJCNQhyXzK3aI/edit#). This allows teams to share builds, implement caching and other shared volumes. A single persistent volume is mapped to each default build pipeline. Builds are passed a directory specific to their builds.
```
gbenhaim, line 87: The link point to the header of the document, can you please change it to point to the section about storage?
In addition, this repository is public, so we need to attach only documents that are public.

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,277 @@
...
+Whenever possible, resources like ConfigMaps or Secrets will be used to inject configuration into Tasks. This is preferred to templates and patches as it fits well with Kubernetes declarative and GitOps approaches.
+
+ConfigMaps and Secrets will be mounted into the Task pod to inject file-based information like certificates.
+Environment variables may be injected from ConfigMaps.
+
+Variables directly configured in the Pod definition are discouraged.
+
+Side note: Secrets and ConfigMaps should be made [immutable](https://kubernetes.io/docs/concepts/configuration/secret/#secret-immutable)
+
+### Information vector
```
gbenhaim, line 71: What is the purpose of this section? seems like it repeats Tekton's documentation.

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,277 @@
...
+```
+Example for a run with an error:
+```
+{
+    "result":"ERROR",
+    "timestamp":"1649842004"
+}
+```
+
+###### Proposed Adapted Tekton Results for Xunit Tests Results
```
gbenhaim, line 185: Xunit test result has a well-knowen format that is understood by a lot of tools. I don't think we should transform it into our own format. We can just forward it as is.

ADR/0008-integration-service-api-contracts.md
hongweiliu17: Thanks for you advice. I will change it to point to the section about storage.

About attaching public doc, do you have any idea if we want to link that g-doc to make the users of this ADR know it? Maybe we can use this g-doc link and then update the link once the doc from build team is made public in the future?

ADR/0008-integration-service-api-contracts.md
hongweiliu17: I think listing them here is to highlight it and make them more visible.

ADR/0008-integration-service-api-contracts.md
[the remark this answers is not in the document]
hongweiliu17: Thanks! It is a leftover from previous doc, let's remove it.

ADR/0008-integration-service-api-contracts.md
hongweiliu17: I don't feel we will test changes to the ConfigMap, @dirgim could you help confirm it? Thanks!

dirgim: @hongweiliu17 you're correct, we will not test changes to the `ConfigMaps` currently. 
We're currently only using `Secrets` to inject configuration that allows tasks to access the testing environments and are not using `ConfigMaps` at this point, but have left them as an option for future enhancements.

ADR/0008-integration-service-api-contracts.md
dirgim: @hongweiliu17 yes, the connection details will be created per-environment and stored in a `Secret`. That `Secret` will be available to be mounted for individual pipeline tasks. The details for how to use that secret were worked out as part of [HACBS-1234](https://issues.redhat.com/browse/HACBS-1234) and will be further documented in [HACDOCS-179](https://issues.redhat.com/browse/HACDOCS-179).

hongweiliu17: Thanks Kruno.

ADR/0008-integration-service-api-contracts.md
hongweiliu17: Thanks Kruno.

ADR/0008-integration-service-api-contracts.md
Dannyb48: Yeah, the idea wasn't to completely convert it and display the whole junit/xunit results output. I agree it is a well understood format. So full details it can be viewed elsewhere. The idea was to provide a summarized of the results similar to what we did with the Sanity/SAST tasks for specific tekton task that was go to execute the test framework. Figured I would provide some context since I originally added that into the contract doc :)

ADR/0008-integration-service-api-contracts.md
hongweiliu17: I add the tekton link here as well so that user can read more details from Tekton's documentation by clicking the links.

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,277 @@
...
+
+
+The maximum size of a Task's Results is limited by the container termination message feature of Kubernetes.
+
+
+App Studio builds are structured as [a shared Persistent Volume per App Studio Workspace](https://docs.google.com/document/d/1IPlihVjkJ4Kb9tdhsk7iz3bn5rkT_SvJCNQhyXzK3aI/edit#). This allows teams to share builds, implement caching and other shared volumes. A single persistent volume is mapped to each default build pipeline. Builds are passed a directory specific to their builds.
+
+TODO: provide decision factors
+
+### Information format
```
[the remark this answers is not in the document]
hongweiliu17: Thanks, tekton doc links are provided here, and user can read more details from the links.

ADR/0008-integration-service-api-contracts.md
hongweiliu17: Thanks @Dannyb48, I feel it is a summary as well so we can get the points quickly.

ralphbean: Yeah - I think this is the crux of the ADR. We have to get results out of a task by way of HACBS_TEST_OUTPUT because of the size limitation on task results. Normal results (like full xunit results) just can't be exposed.

ralphbean (changes requested):
I'd like to see the document restructured to center around HACBS_TEST_OUTPUT (the decision), what caused us to introduce it (the context), and what are some of the benefits are drawbacks of this decision we made (the consequences).

There is lots of extra information here that I think is distracting from that point - the recopying of information from the tekton docs, for instance.

Things like the naming convention section and the DRY section don't actually form a "contract" - they're just a good practice. Perhaps those could be moved to a new "additional recommendations" section at the bottom of the doc if they're especially valuable to record - otherwise they could be removed.

Apologies - I know I requested this ADR to be just a transcription of the API Contract doc written earlier in the year. As I read it now in the context of this repo, I think it does need changes.

hongweiliu17:
> I'd like to see the document restructured to center around HACBS_TEST_OUTPUT (the decision), what caused us to introduce it (the context), and what are some of the benefits and drawbacks of this decision that we made (the consequences).
> 
> There is lots of extra information here that I think is distracting from that point - the recopying of information from the tekton docs, for instance.
> 
> Things like the naming convention section and the DRY section don't actually form a "contract" - they're just a good practice. Perhaps those could be moved to a new "additional recommendations" section at the bottom of the doc if they're especially valuable to record - otherwise they could be removed.
> 
> Apologies - I know I requested this ADR to be just a transcription of the API Contract doc written earlier in the year. As I read it now in the context of this repo, I think it does need changes.

Thanks for the feedback, let us restructure it.

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,264 @@
...
+
+The output of each Tekton task will be provided in two forms:
+- As a minimized [Tekton result](https://tekton.dev/docs/pipelines/tasks/#emitting-results) in JSON format listing all test failures
+  - The name of the result will be **HACBS_TEST_OUTPUT**
+- As a Test output JSON file with full test information that’s saved in the Tekton Pipeline Workspace
+  - The name of the file will be in snake case and will be in the form of **test_name_output.json**
+
+#### Tekton Results
+
+Each test will have a standardized short name in snake case, e.g. `release_label_required`, `architecture_label_deprecated` etc.
```
dirgim, line 41: Since we converted the failures to an integer field so it doesn't contain the test names in it anymore, I think we can move this line to the `Full Test Output JSON` section, probably somewhere around [line 135](https://github.com/redhat-appstudio/book/pull/24/files#diff-c480f27e8be91d3a13f9ed1429fff01b9790de544a57d7c812a25137a1898f61R135).

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,264 @@
...
+
+The Test output of the Tekton result will be a JSON object that includes the context about the test along with the list of check names for all failed checks
+
+The output will provide the following information about the overall test result:
+- **result** - The outcome of the testing task, can be `SUCCESS`, `FAILURE`, `WARNING`, `SKIPPED` or `ERROR`
+- **namespace** - The rego namespace of the test policy
```
dirgim, line 47: Maybe we can mark this as an optional field because it will not be needed for integration test use cases?

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,264 @@
...
+- **timestamp** - An UNIX epoch timestamp of the test completion time
+- **successes** - The number of successful checks in the form of an integer
+- **note** - A short note providing additional information about the test
```
dirgim, line 50: We can possibly mark this as an optional field too I think.

ADR/0008-integration-service-api-contracts.md
hongweiliu17: Thanks, it is moved to the new [line 134](https://github.com/redhat-appstudio/book/pull/24/files#diff-c480f27e8be91d3a13f9ed1429fff01b9790de544a57d7c812a25137a1898f61R134)

ADR/0008-integration-service-api-contracts.md
hongweiliu17: `Optional` is added [here](https://github.com/redhat-appstudio/book/pull/24/files#diff-c480f27e8be91d3a13f9ed1429fff01b9790de544a57d7c812a25137a1898f61R45)

ADR/0008-integration-service-api-contracts.md
hongweiliu17: `Optional` is added [here](https://github.com/redhat-appstudio/book/pull/24/files#diff-c480f27e8be91d3a13f9ed1429fff01b9790de544a57d7c812a25137a1898f61R48)

hongweiliu17:
> I'd like to see the document restructured to center around HACBS_TEST_OUTPUT (the decision), what caused us to introduce it (the context), and what are some of the benefits and drawbacks of this decision that we made (the consequences).
> 
> There is lots of extra information here that I think is distracting from that point - the recopying of information from the tekton docs, for instance.
> 
> Things like the naming convention section and the DRY section don't actually form a "contract" - they're just a good practice. Perhaps those could be moved to a new "additional recommendations" section at the bottom of the doc if they're especially valuable to record - otherwise they could be removed.
> 
> Apologies - I know I requested this ADR to be just a transcription of the API Contract doc written earlier in the year. As I read it now in the context of this repo, I think it does need changes.

The ADR is restructured and can be reviewed again. Thanks!

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,266 @@
...
+4. Output of the Task that may be needed for processing subsequent Tasks of the pipeline
+5. Output of the Task that need to be stored for auditing and troubleshooting (that is summarized results of a validation or a scan, etc)
+6. Communication with external systems that are needed for performing the Task (that is container image repository, image scanner, etc)
+7. Connection details and credentials that may be required for the above
+
+## Decision
+
+### Results and Records
+
+Related to [PLNSRVCE-41](https://issues.redhat.com/browse/PLNSRVCE-41) investigations in regards to [Tekton results](https://github.com/tektoncd/results). The investigations are documented [here](https://docs.google.com/document/d/1WGWx6MllVs-BwRLd0PP5QTQHgGp0fW2dOYRKRAKWiwo/edit#heading=h.bpyxfhyud4dv).
```
ralphbean, line 31: Move the information in this sentence to the `Context` section please. It provides background on why and how we came to the `Decision`.

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,266 @@
...
+
+The output of each Tekton task will be provided in two forms:
+- As a minimized [Tekton result](https://tekton.dev/docs/pipelines/tasks/#emitting-results) in JSON format listing all test failures
+  - The name of the result will be **HACBS_TEST_OUTPUT**
+- As a Test output JSON file with full test information that’s saved in the Tekton Pipeline Workspace
+  - The name of the file will be in snake case and will be in the form of **test_name_output.json**
+
+#### Tekton Results
```
ralphbean, line 39:
```suggestion
#### Tekton Result Format
```

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,266 @@
...
+
+The Test output of the Tekton result will be a JSON object that includes the context about the test along with the list of check names for all failed checks
+
+The output will provide the following information about the overall test result:
```
ralphbean, line 43: Do we happen to have this format encoded as [json schema](https://json-schema.org/) anywhere? That could be nice to link to or copy here directly if we have it. Nicely unambiguous.

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,266 @@
...
+```
+Example for a run with an error:
+```
+{
+    "result": "ERROR",
+    "timestamp": "1649842004"
+}
+```
+
+##### Proposed Adapted Tekton Results for Xunit Tests Results
```
ralphbean, line 90:
```suggestion
##### Proposed Adapted Tekton Result Format for Xunit Tests Results
```

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,266 @@
...
+                        "url": "https://source.redhat.com/groups/public/container-build-system/container_build_system_wiki/guide_to_layered_image_build_service_osbs#jive_content_id_Labels"
+                    }
+                }
+            }
+          ]
+    }
+]
+```
+
+### Information injection
```
ralphbean, line 171: This section is too open ended.

I don't think I see an example in [build-definitions](https://github.com/redhat-appstudio/build-definitions/) of a `ConfigMap` being used. I do see that `tasks/sast-snyk-check.yaml` uses a Secret.

Is the ConfigMap part relevant?

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,266 @@
...
+Whenever possible, resources like ConfigMaps or Secrets will be used to inject configuration into Tasks. This is preferred to templates and patches as it fits well with Kubernetes declarative and GitOps approaches.
+
+ConfigMaps and Secrets will be mounted into the Task pod to inject file-based information like certificates.
+Environment variables may be injected from ConfigMaps.
+
+Variables directly configured in the Pod definition are discouraged.
+
+Side note: Secrets and ConfigMaps should be made [immutable](https://kubernetes.io/docs/concepts/configuration/secret/#secret-immutable)
+
+### Information vector
```
ralphbean, line 182: I don't understand the purpose of this Information vector section.

There's a list of things, with no indication of when to use them. There's a (correct) note about the limmitations on task result sizes - which is actually more relevant for the section at the top of the Decision (it is why we need a minimal HACBS_TEST_OUTPUT). The sentence about the volume is also correct, but not sure why it is here.

Delete this section?

hongweiliu17: Let's delete the list and move the limitation to the top of the `Decision`

ADR/0008-integration-service-api-contracts.md
hongweiliu17: Currently we use `Secrets` to inject configuration into Tasks only, will we use `ConfigMaps` next maybe?

ADR/0008-integration-service-api-contracts.md
hongweiliu17: Thanks!

ADR/0008-integration-service-api-contracts.md
hongweiliu17: I don't think we have json schema for it, but we generate it with `jq`. @dirgim Could you help confirm it?

ADR/0008-integration-service-api-contracts.md
hongweiliu17: Done. Thanks.

hongweiliu17:
@ralphbean I revised the file, could you take a look at this PR again? Thank you!

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,256 @@
...
+
+- Existing Mapped:
+
+  - **Namespace** == **name** attribute of all of the **TestSuite** elements 
+
+  - **Failures** == **name** attribute for all the **TestCase** elements that have the attribute **status=Failed**
+
+- New Key:
+
+  - **Skipped** == the number of skipped test in integer format from the **TestSuite** attribute **skipped** or from **TestCase** attribute **skipped**
```
arewm, line 104: Why is skipped here and not above?

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,256 @@
...
+    "result": "ERROR",
+    "timestamp": "1649842004"
+}
+```
+
+##### Proposed Adapted Tekton Result Format for Xunit Tests Results
+
+- Existing Mapped:
+
+  - **Namespace** == **name** attribute of all of the **TestSuite** elements
```
arewm, line 98: Keys are not capitalized above. Should we be consistent and not capitalize here?

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,256 @@
...
+```
+[
+    {
+        "result": "FAILURE",
+        "namespace": "Red Hat App Studio E2E tests",
+        "timestamp": "1649148140",
+        "successes": 75,
+        "skipped": 14,
+        "note": "",
+        "failures": [
```
arewm, line 118: Why are you changing failures to a list and not successes and skipped? I feel like the minimized result should always have the number of tests and leave these details to the volume-stored result.

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,256 @@
...
+#### Tekton Result Format
+
+The Test output of the Tekton result will be a JSON object that includes the context about the test along with the list of check names for all failed checks
+
+The output will provide the following information about the overall test result:
+- **result** - The outcome of the testing task, can be `SUCCESS`, `FAILURE`, `WARNING`, `SKIPPED` or `ERROR`
+- **namespace** - Optional, the rego namespace of the test policy
+- **timestamp** - An UNIX epoch timestamp of the test completion time
+- **successes** - The number of successful checks in the form of an integer
+- **note** - Optional, a short note providing additional information about the test
```
arewm, line 52: Is the suggested use of this to describe the test itself or to pass information about the state of the test? The latter feels like it might have more utility but can also be used to fulfil the former (and stated) purpose.

All examples just have this as an empty string, but I would expect that to be omitted in those cases to save emitted result space.

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,256 @@
...
+- **successes** - The number of successful checks in the form of an integer
+- **failures** - A JSON list containing objects describing each failure
+
+The **failures** list of objects will provide the following information about each check:
+- **name** - Name of the individual check written in [snake case](https://en.wikipedia.org/wiki/Snake_case)
+- **msg** - Message about what went wrong with the check
+- **description** - The explanation about why the check is significant (Why the label is used, why the vulnerabilities are checked etc.)
+- **url** - link to the further documentation on the individual check (or that type of check)
+
+Example full JSON test output for a Tekton task that tests the container image labels:
```
arewm, line 153: Do you not expect skips, error, and warnings to be represented here?

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,256 @@
...
+            }
+          ]
+    }
+]
+```
+
+### Information injection
+
+Whenever possible, resources like ConfigMaps or Secrets will be used to inject configuration into Tasks. This is preferred to templates and patches as it fits well with Kubernetes declarative and GitOps approaches.
+tf
```
arewm, line 179: accidental characters?

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,256 @@
...
+
+### Information format
+
+[Clearly define the format of input parameters and results](https://github.com/tektoncd/catalog/blob/main/recommendations.md#clearly-define-the-format-of-input-parameters-and-results).
+
+Atomic information may be passed as a simple parameter. Non binary encoded complex information should be exchanged through [JSON format](https://github.com/tektoncd/catalog/blob/main/recommendations.md#use-composable-parameter-formats) between Tasks.
+
+#### Image references
+
+Tags are not uniquely identifying an image so that we cannot rely on them. They do not guarantee that the scanning is for the image built as part of the PipelineRun if there is a race with a parallel build. As such, image digests will be used for image references.
```
arewm, line 196: Digests can be rendered invalid if there is no tag pointing to them, so if you are concerned about the tag moving to a different digest, then the digest references will still end up being invalid.

I am not disagreeing with the use of digest, just disagreeing on the logic to defend the choice.

ADR/0008-integration-service-api-contracts.md
hongweiliu17: Updated. Thanks.

ADR/0008-integration-service-api-contracts.md
hongweiliu17: Yes, thanks.

ADR/0008-integration-service-api-contracts.md
hongweiliu17: Thanks!

ADR/0008-integration-service-api-contracts.md
hongweiliu17: Thanks! I update it as below, is it ok?
```
Tags are not uniquely identifying an image so that we cannot rely on them. They do not guarantee that the scanning is for the image built as part of the PipelineRun if there is a race with a parallel build. As such, image digests will be used for image references then we can get an error once the original image doesn't exist any more.
```

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,256 @@
...
+{
+    "result": "SKIPPED",
+    "note": "<skipped message>",
+    "timestamp": "1649842004"
+}
+```
+Example for a run with an error:
+```
+{
+    "result": "ERROR",
```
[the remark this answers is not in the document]
dirgim: Yes, adding a note for users to check the task logs sounds like a good idea. 
In the future, we expect to have persistent task logs and other artifacts, but currently I think that having such a message is the way to go.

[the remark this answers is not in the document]
arewm: Thanks for the background. I am fine with that approach for `note` as long as the error message is actionable unlike its current form. Is there a way that you can clearly, concisely, and precisely show the user how to check the logs?

Can we still report the `namespace` of the error?

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,256 @@
...
+    "namespace": "required_checks",
+    "successes": 16,
+    "note": "",
+    "failures": 0
+}
+```
+Example for a skipped run:
+```
+{
+    "result": "SKIPPED",
```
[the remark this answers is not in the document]
arewm: Is the `namespace` not populated when a test is skipped?

ADR/0008-integration-service-api-contracts.md
arewm: The `note` is never populated by the test itself, but always populated by the test stream, correct? If only used for `SKIPPED` results (and possibly for `ERROR` as discussed), maybe this would be more accurate so say that it is additional information about the test result?

If the field is indeed optional, I would expect it to be absent instead of having an empty string value. If some results have an empty string returned, I would expect all to have a minimum empty string. In this latter scenario, the key would not be optional, but additional context within the key is still optional.

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,256 @@
...
+- As a minimized [Tekton result](https://tekton.dev/docs/pipelines/tasks/#emitting-results) in JSON format listing all test failures
+  - The name of the result will be **HACBS_TEST_OUTPUT**
+- As a Test output JSON file with full test information that’s saved in the Tekton Pipeline Workspace
+  - The name of the file will be in snake case and will be in the form of **test_name_output.json**
+
+The maximum size of a [Task's Results](https://tekton.dev/vault/pipelines-v0.17.3/tasks/#emitting-results) is limited by the container [termination message](https://kubernetes.io/docs/tasks/debug/debug-application/determine-reason-pod-failure/#customizing-the-termination-message) feature of Kubernetes.
+
+App Studio builds are structured as [a shared Persistent Volume per App Studio Workspace](https://docs.google.com/document/d/1IPlihVjkJ4Kb9tdhsk7iz3bn5rkT_SvJCNQhyXzK3aI/edit#bookmark=id.gefgys3vno2). This allows teams to share builds, implement caching and other shared volumes. A single persistent volume is mapped to each default build pipeline. Builds are passed a directory specific to their builds.
+
+#### Tekton Result Format
```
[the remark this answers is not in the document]
arewm: This is splitting into the following two sections, correct?

```
### Tekton Task Results
### Full Test Output JSON
```

That is good -- it avoids that confusion. I do, however, miss the introduction that you had summarizing why the two types of results are being used. Am I just missing this in the rewrite?

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,256 @@
...
+
+The Test output of the Tekton result will be a JSON object that includes the context about the test along with the list of check names for all failed checks
+
+The output will provide the following information about the overall test result:
+- **result** - The outcome of the testing task, can be `SUCCESS`, `FAILURE`, `WARNING`, `SKIPPED` or `ERROR`
```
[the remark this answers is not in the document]
arewm: Thanks for that added detail. Does that information need to be added -- that a test is only classified as `SKIPPED` or `ERROR` as a whole; that specific checks will not be further described?

How does `WARNING` behave? Is it considered a pass or a failure if some number of checks has this status? I am interpreting the precedence as `FAILURE` > `WARNING` > `SUCCESS` (when discrete checks are actually reported). If any check has a result of a higher precedence status than that status will be the overall test result. Is that accurate?

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,248 @@
...
+- **namespace** - Optional, the rego namespace of the test policy
+- **timestamp** - An UNIX epoch timestamp of the test completion time
+- **successes** - The number of successful checks in the form of an integer
+- **note** - Optional, a short note providing additional information about the test
+- **failures** - The number of failed checks in the form of an integer
+- **warnings** - The number of warning checks in the form of an integer
```
arewm, line 50: **successes**, **failures**, and **warnings** are implied non-optional but the examples below do not all include these fields. Are these required or optional?

ADR/0008-integration-service-api-contracts.md
arewm: This mean that it is possible for the shortened result to have more information than the full test output? If we are reporting these to the result, I would expect it to be reported to the full record even if no additional information/context is given.

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,248 @@
...
+
+Date: 2023-01-[]
+
+## Status
+
+Accepted
+
+## Context
+
+The Hybrid Application Cloud Build Service being developed aims to serve Red Hat teams but also partners and customers. This requires a level of adaptability to avoid recreating custom flows and Tasks for each stakeholder.
```
arewm, line 11: Hybrid Application Cloud Build Service == HACBS but I don't know what the new way of saying this would be. :-/

ADR/0008-integration-service-api-contracts.md
arewm: How about something like

> Since tags can be moved from one image to another, they should not be relied on as a reference. In order to guarantee that any scanning is performed on an image built as part of a PipelineRun, the immutable image digest reference will be used instead.

The fact that digests can become invalid if all tags are removed is an implementation detail of the registry. Hopefully that won't happen to people, but I don't think it has any bearing on the test workstream if it does.

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,256 @@
+# 8. HACBS Test Stream - API contracts
```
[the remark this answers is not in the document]
hongweiliu17: Thanks! I will update it.

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,256 @@
...
+7. Connection details and credentials that may be required for the above
+
+Related to [PLNSRVCE-41](https://issues.redhat.com/browse/PLNSRVCE-41) investigations in regards to [Tekton results](https://github.com/tektoncd/results). The investigations are documented [here](https://docs.google.com/document/d/1WGWx6MllVs-BwRLd0PP5QTQHgGp0fW2dOYRKRAKWiwo/edit#heading=h.bpyxfhyud4dv).
+
+## Decision
+
+### Results and Records
+
+The output of each Tekton task will be provided in two forms:
+- As a minimized [Tekton result](https://tekton.dev/docs/pipelines/tasks/#emitting-results) in JSON format listing all test failures
```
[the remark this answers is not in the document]
hongweiliu17: Split `Results and Records` to `Tekton Task Results` and `Full Test Output JSON` and add the clarification to each of them.

ADR/0008-integration-service-api-contracts.md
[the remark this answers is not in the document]
hongweiliu17: We have not had `WARNING` status so far, but I think it should be considered as pass.
These results are final result of tekton tasks, each of task can have only one of them.

jira task [HACBS-1598](https://issues.redhat.com/browse/HACBS-1598) is filed to add more informative message to note of result.

ADR/0008-integration-service-api-contracts.md
hongweiliu17: Yes, the content in `note` is provided by test stream. I remove the empty note from the examples
and change the description of `note` to `Optional, a short note provided by test workstream to provide additional information about the test`.

As mention above, [HACBS-1598](https://issues.redhat.com/browse/HACBS-1598) is filed to add a short message to [ERROR, SKIPPED, FAILURE] status result

ADR/0008-integration-service-api-contracts.md
hongweiliu17: The shortened result will have only the number of successful/failed/warning and a short message because the maximum size of a [Task's Results](https://tekton.dev/vault/pipelines-v0.17.3/tasks/#emitting-results is limited so the information is less than the full json result. The failures items will not shown in shortened tekton task result but they are logged out in tekton task log.

ADR/0008-integration-service-api-contracts.md
hongweiliu17: Thanks!

ADR/0008-integration-service-api-contracts.md
hongweiliu17: So let's use `The Stonesoup project` now, and we may update it if it has a different name. Thanks.

ADR/0008-integration-service-api-contracts.md
hongweiliu17: Thanks, I think we use `required` because they are required for specific status, such as  `failures` is required for the status is `FAILURE`. But it may be better to list all fields of `successes`, `failures`, and `warnings` for all result `SUCCESS`, `FAILURE`, `WARNING`, `SKIPPED` or `ERROR`.  Is it? @dirgim

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,256 @@
...
+  - The name of the result will be **HACBS_TEST_OUTPUT**
+- As a Test output JSON file with full test information that’s saved in the Tekton Pipeline Workspace
+  - The name of the file will be in snake case and will be in the form of **test_name_output.json**
+
+The maximum size of a [Task's Results](https://tekton.dev/vault/pipelines-v0.17.3/tasks/#emitting-results) is limited by the container [termination message](https://kubernetes.io/docs/tasks/debug/debug-application/determine-reason-pod-failure/#customizing-the-termination-message) feature of Kubernetes.
```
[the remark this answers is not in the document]
hongweiliu17: No, user can not input message for `note` now. The `skip` often means the test is not run because there is not supported files, such as https://github.com/redhat-appstudio/build-definitions/blob/main/task/sast-snyk-check/0.1/sast-snyk-check.yaml#L48 and https://github.com/redhat-appstudio/build-definitions/blob/main/task/sast-java-sec-check/0.1/sast-java-sec-check.yaml#L66

If the result is longer than 4096 Bytes, the task will meet error `Termination message is above max allowed size 4096, caused by large task result`.

Is a longer namespace meaningful for the team(EC?) consuming these fields? We may confirm with them.

ADR/0008-integration-service-api-contracts.md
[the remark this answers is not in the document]
hongweiliu17: I remove it, let me add it back like this, it is ok?
```
Related to [PLNSRVCE-41](https://issues.redhat.com/browse/PLNSRVCE-41) investigations in regards to [Tekton results](https://github.com/tektoncd/results). The investigations are documented [here](https://docs.google.com/document/d/1WGWx6MllVs-BwRLd0PP5QTQHgGp0fW2dOYRKRAKWiwo/edit#heading=h.bpyxfhyud4dv). The output of each Tekton task will be provided in two forms: Tekton Task Results and Full Test Output JSON.
```

ADR/0008-integration-service-api-contracts.md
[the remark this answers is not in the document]
hongweiliu17: Yes, skipped.

ADR/0008-integration-service-api-contracts.md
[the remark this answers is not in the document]
hongweiliu17: Ok, let's add it. Tracked in [HACBS-1598](https://issues.redhat.com/browse/HACBS-1598)

ADR/0008-integration-service-api-contracts.md
dirgim: Yes, I agree that we can have those fields as required to show up for all statuses.

ADR/0008-integration-service-api-contracts.md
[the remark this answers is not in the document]
arewm: I assume that a long namespace field would just be there if the test suite had a long name. I don't think there is any inherent (dis)advantage to a long or short name.

ADR/0008-integration-service-api-contracts.md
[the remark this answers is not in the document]
arewm: I think that helps, thanks! Sorry for the iteration. :)

ADR/0008-integration-service-api-contracts.md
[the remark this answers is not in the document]
arewm: Would it be up to a specific test case/suite to raise a warning?

Is it within the purview of this ADR to determine whether a warning is a pass or a fail -- to help the classification of other components and tests? If so, we can just note which types of statuses are considered success-like and which ones are considered fail-like (likely not with those terms though?).

ADR/0008-integration-service-api-contracts.md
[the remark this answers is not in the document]
arewm: and the namespace will continue being empty under these cases. Thanks --

ADR/0008-integration-service-api-contracts.md
arewm: I know that we are deferring some decisions (i.e. around `note` content) to HACBS-1598. Can we at least also consider that issue for adding detailed output for these same cases (well, plus WARNING which isn't currently in the issue)? If some information is being populated to the shortened results, I would expect to be able to glean at least that much information from the full results. In this case, it would at least be the presence/count of skips, errors, and warnings. Preferably with more context/information added in -- if possible.

ADR/0008-integration-service-api-contracts.md
arewm: Thanks for filing that issue. I'll watch it!

ADR/0008-integration-service-api-contracts.md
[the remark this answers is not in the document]
hongweiliu17: Thanks, the normal namespace(The package name of Policy https://github.com/redhat-appstudio/hacbs-test/blob/main/policies/clair/vulnerabilities-check.rego#L1) is supported and will not exceed the limit.

ADR/0008-integration-service-api-contracts.md
hongweiliu17: I updated the AC of https://issues.redhat.com/browse/HACBS-1598, could you take a look? Thanks so much for your ideas!

arewm: HACBS-1598 only takes into account the short results. This thread is focused on ensuring that the long results have at least as much detail as the short ones, namely that skips, error, and warnings are represented here as well. Would it be best to create a new issue for this or is it something that can be modified in this ADR?

arewm: Apologies for the repetition. I might have gotten lost in all the threads I opened up. I think that my hang-up is coming from the fact that we are calling it the full json output. It is not the full output unless we include all possible output including if there are any skips and errors. As it is now, it is more like a detailed successful and failure output.

ADR/0008-integration-service-api-contracts.md
ralphbean: The way I'm thinking about it now is that I'd rather link to a private doc than not link at all. As we open up this doc and others, over time we should wind up with only public-facing links. The ADRs are historical documents, documenting decisions at a point in time. The reality today is that most of those docs are private, but here we are working to change that, one document at a time. :)

ADR/0008-integration-service-api-contracts.md
ralphbean: All good - just curious. Thanks!

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,223 @@
...
+
+[Clearly define the format of input parameters and results](https://github.com/tektoncd/catalog/blob/main/recommendations.md#clearly-define-the-format-of-input-parameters-and-results).
+
+Atomic information may be passed as a simple parameter. Non binary encoded complex information should be exchanged through [JSON format](https://github.com/tektoncd/catalog/blob/main/recommendations.md#use-composable-parameter-formats) between Tasks.
+
+#### Image references
+
+Since tags can be moved from one image to another, they should not be relied on as a reference. In order to guarantee that any scanning is performed on an image built as part of a PipelineRun, the immutable image digest reference will be used instead.
+
+#### Logging
```
ralphbean, line 165: I suggest deleting this whole `#### Logging` section. I think it is duplicated by https://github.com/redhat-appstudio/book/pull/21.

ADR/0008-integration-service-api-contracts.md
dirgim: I can see the confusion now. I think that renaming the `Full JSON output` to `Detailed Conftest output` should clear up the confusion here, as that's a more accurate name for it.

arewm: That would work! I am glad that I found the proper words to convey what was in my head. :D

ADR/0008-integration-service-api-contracts.md
```diff
@@ -0,0 +1,223 @@
...
+- CompRel, for instance: "709"
+
+Today NVR is largely used for tracking, for instance hco-bundle-registry-container-v4.10.0-709.
+
+The idea is that the common fields listed above would be populated for each log entry when the information is available.
+
+It is advisable to limit the number of structured fields as it impacts indexing. Fields that are common to multiple tasks should be preferred. Before a field specific to a task is added it is worth thinking whether it is really needed and whether the same information can be conveyed through a more generic field. Information that does not need to be “searchable” can be recorded in an unstructured way.
+
+## Consequences
+
```
ralphbean, line 187:
```suggestion
As a result of the decision here to summarize results in a **HACBS_TEST_OUTPUT** result and store the larger test output as a file named **test_name_output.json**, we should find that:

* Other components in Stonesoup can leverage information exposed by TaskRuns - notably the UI (HAC), integration-service, and enterprise-contract - enabling features for the larger system that need to depend on some data from inside a variety of TaskRuns.
* We'll be able to have PipelineRuns that _succeed_ and continue, even if they have tasks whose payloads fail and expose errors. This enables a progressive model where the user can get a build and a functional test and a deployment to their development Environment, even if a linter or scanner in their build pipeline emits an error.
* By having chosen a convention that has "HACBS" in the name, we're going to have trouble integrating third-party Task providers in the future. In order to have its output respected by our system, a hypothetical vendor of a third-party scanner will need to add a "HACBS_TEST_OUTPUT" result on their Task, which is oddly specific to our system. At some point in the future, we should revise this decision to instead align to a common upstream convention that gains traction in the broader tekton ecosystem. See also [HACBS-1563](https://issues.redhat.com/browse/HACBS-1563).
```

hongweiliu17: Updated, thanks!

ADR/0008-integration-service-api-contracts.md
hongweiliu17: Removed. Thanks!

ADR/0008-integration-service-api-contracts.md
hongweiliu17: Updated. Thanks!

ralphbean (approved):

arewm (approved):
Looks good to me but we will need to fix the ADR conflict as 8 is taken and it needs a date as well.

PR #26, opened 2022-12-01
ADR-0009: Pipeline Service via Operator

adambkaplan:
Deploy "Pipeline Service" using the stock OpenShift Pipelines operator, ensuring that the "latest" code is deployed and that all configuration is driven by the operator.


adambkaplan:
@sbose78 @vdemeester @concaf I hope this captures the high-level goals of the "service first" conversations we were having via email and Slack. I am purposefully omitting the details of how we get there.

concaf (commented):
Well put 🙃 
The product will (and should) eventually ship its operator to Pipeline Serivce via StoneSoup once we can get an operator build from it, in the meantime let's leverage what we currently have to get moving with the Service First initiative at least.

ADR/0009-pipeline-service-via-operator.md
```diff
@@ -0,0 +1,40 @@
...
+[infra-deployments](https://github.com/redhat-appstudio/infra-deployments) as
+an ArgoCD application.
+
+## Consequences
+
+- Tekton Results will need to be productized and incorporated into the
+OpenShift Pipelines operator.
+- Tekton Triggers should be disabled using appropriate operator configuration.
+- The pruner for Tekton needs to be disabled using appropriate operator configuration.
+- Pipelines as Code should use an appropriate URL to the Hybrid Application
```
chmouel, line 32: when you say appropriate, i am guessing you mean a public route accessible from GitHub ?

book/pipeline-service.md
```diff
@@ -1,29 +1,31 @@
...
-- `Task`
-- `Pipeline`
-- `PipelineRun`
-- `Repository` (for [Pipelines as Code](https://pipelinesascode.com))
+The deployment of the OpenShift Pipelines operator will have the following
+notable configurations:
 
-The following features are also provided:
+- Tekton Triggers will be disabled entirely.
+- The pruner (provided by the Pipelines operator) will be disabled in favor of
```
chmouel, line 13: just wondering, is max-keep-run pac feature will not used?

ADR/0009-pipeline-service-via-operator.md
```diff
@@ -0,0 +1,40 @@
...
+kcp is no longer being used as a control plane for StoneSoup. This means that
+"Pipeline Service" cannot be deployed as an independent service. For our
+initial MVP, all Tekton APIs need to be deployed onto a standard OpenShift
+cluster (specifically OpenShift Dedicated).
+
+## Decision
+
+All Tekton APIs will be provided using the stock OpenShift Pipelines operator.
+In the spirit of developing in a "service first" manner, StoneSoup will deploy
+a candidate "nightly" release of the operator. The service will be defined in
```
Michkov, line 20: that's good for testing environments but for production?

book/pipeline-service.md
```diff
@@ -1,29 +1,31 @@
...
+  pruning via Tekton Results.
+- Pipelines as Code will link the results of pipeline tasks to an appropriate
+  [Hybrid Application Console (HAC)](./hybrid-application-console.md) URL.
 
-- Signing and attestation of `TaskRuns` with [Tekton Chains](https://tekton.dev/docs/chains/)
```
bnallapeta, line 13: We still provide Tekton Results and Chains right? Any reason to omit that information from this file?

book/pipeline-service.md
adambkaplan: Results has the ability to delete PipelineRuns/TaskRuns once they are archived. We want pipeline data to live off cluster as much as possible, and use Results to gate retention policies.

book/pipeline-service.md
adambkaplan: This was a mistake, I will restore it (along with the info about Results).

ADR/0009-pipeline-service-via-operator.md
adambkaplan: Yes. StoneSoup has an independent interface (Hybrid Application Console) that runs the user experience and lives outside of the cluster. The OpenShift web console won't be accessible from the external internet.

ADR/0009-pipeline-service-via-operator.md
adambkaplan: We'll want to employ some level of quality gates on our staging environment before we release to production. We must aspire to merge production-grade code, or we will fail in this endeavor. The time from code merge to code deploy _in production_ needs to get to 4 hours, not 4 days (or 4 months, which is the time it takes for new features to be released today).

ADR/0009-pipeline-service-via-operator.md
```diff
@@ -0,0 +1,53 @@
...
+[infra-deployments](https://github.com/redhat-appstudio/infra-deployments) as
+an ArgoCD application.
+
+Configurations that are specific to StoneSoup must be made available through
+the Pipelines operator CRDs. The following changes are specific to StoneSoup:
+
+- Disable Tekton Triggers. Pipelines as Code will be the favored mechanism for
+  event-based triggering of pipelines for now. This decision can be revisited
+  in the future based on need.
+- Disable the pruner that comes with the operator. Tekton Results will be used
```
vdemeester, line 32: Do we consider this as "temporary" — because the pruner doesn't fit the bill today — or a finite decision ?
Asking because the requirements that Pipeline Service / StoneSoup has/will have are most likely also going to be very useful for our customers.

ADR/0009-pipeline-service-via-operator.md
sbose78: > all of which are developed by, possibly, a different set of people with a different set of rules,

We'll need to influence the projects/repos to have CI tests which give us production-like quality such that a merge-to-main commit is deployable with reasonable guarantees. 

> So when we talk about nightly here, is it about “nightly” code from the pipeline service, from the additional, owned feature of OpenShift Pipelines or from upstream components ?

In my opinion, 

Long-term, everything.
Short-term, (a) The core Tekton APIs could move slow and steady (b) The non-core Tekton APIs could be cutting edge.

sbose78 (approved):

ADR/0009-pipeline-service-via-operator.md
```diff
@@ -0,0 +1,67 @@
...
+  to prune `PipelineRun` and `TaskRun` data off cluster, thereby ensuring data
+  is archived before removal.
+- Direct Pipelines as Code to use a URL pattern that displays the `PipelineRun`
```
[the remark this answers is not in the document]
adambkaplan: Yes. We should call out that users should not have access to the compute cluster(s).

ADR/0009-pipeline-service-via-operator.md
adambkaplan: Agree here - at the current juncture we are still learning, so we need maximum flexibility with respect to how we configure things for the service. Once we land on something that we believe works at scale, we can advocate for those pieces to be default/primary experiences in the operator.

ADR/0009-pipeline-service-via-operator.md
arewm: > We'll need to influence the projects/repos to have CI tests which give us production-like quality such that a merge-to-main commit is deployable with reasonable guarantees.

Even so, we should not necessarily be relying entirely on upstream CI as that can easily be changed outside of our control. We should still work to have our own quality gates in place, even if that is as minimal as ensuring that the content of the checks hasn't been degraded and that the tests passed on the commits (i.e. not force pushed/merged).

ralphbean: It looks like currently, we have a [Subscription](https://github.com/redhat-appstudio/infra-deployments/blob/main/components/pipeline-service/openshift-pipelines/subscription.yaml) to the redhat product version of openshift pipelines. The proposal is to use a different `source` for that Subscription, and a different `channel`.

The channel on its own is problematic, because each night when the nightly build is produced - I think OLM would automatically deploy it, without quality checks on our end.

If we can configure the CatalogSource in infra-deployments and pin that to an image digest, that will give us an ability to deliberately increment to the next nightly build, forcing our CI to run before rolling it out.

adambkaplan:
/close

@Roming22 @prietyc123 I think you two should re-draft/update this ADR, since you have made a lot of progress with the Pipelines product team and have a better understanding of how this will work.

PR #27, opened 2022-12-03
[WIP] Secret Management for Workloads

gorkem:
Explains the user worklaod secret management.


ADR/adr-0008-secret-mgmt-for-user-workloads.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Context
+
+When user workloads are deployed to environments, the system should be able to provide a way to inject values that is specific to the environment. Today, this is done through environment variables that are managed as overlays on the GitOps repository for the application. However, this method does not provide a good way to manage secrets. This ADR addresses the secret management of user workloads for different environments.
+
+## Decision
+
+### Architecture Overview
+
+All user workloads are synced through GitOps service to their environments. StoneSoup's GitOps service uses the [Argo CD Vault Plugin (AVP)](https://github.com/argoproj-labs/argocd-vault-plugin) to replace secret values from the supported secret management backends.
```
jannfis, line 17: I have a concern with using a plugin to handle secrets management. Unfortunately, plugins (including ACP) are not really suitable for a multi-tenant environment. This is due to the fact that the repository server, where the plugin is executed (either within the repository server's pod itself, or in a sidecar), is not tenant-aware. So basically, any secret that is made available to the repository server, is potentially accessible by all tenants.

This is one of the current Argo CD design limitations. There's a [proposal for Argo CD](https://github.com/argoproj/argo-cd/pull/12123) to change the way how Argo CD handles manifest generation. One of its goals is that multi-tenant systems requiring the repository server to process secret information can securely do so.

However, even with the proposal accepted and the appropriate changes incorporated into the code, we'd some heavy architectural lifting on our side to leverage that.

ADR/adr-0008-secret-mgmt-for-user-workloads.md
```diff
@@ -0,0 +1,60 @@
...
+![](../diagrams/secret-mgmt.excalidraw.svg)
+
+#### Secret management APIs
+An API is provided to create and update secrets on the default backend. This is a REST-based API similar to [SPI token management API](https://github.com/redhat-appstudio/service-provider-integration-operator/blob/main/docs/USER.md#post-tokennamespacename). The secrets are per component and can be varied per environment. Only workspace owners can create and update secrets.
+
+#### The default secret management backend
+**TODO:** Add the default backend and its deployment shape.
+
+#### Bring Your Own Secret Management
+Applications can use secret management backends other than the default secret management backend. However, this requires the GitOps service to provide single-tenant ArgoCD instances. The supported backends will be limited to those supported by [AVP as documented](https://argocd-vault-plugin.readthedocs.io/en/stable/backends/). Creating and updating secrets through StoneSoup secret management REST APIs is not supported for custom backends.
```
jannfis, line 29: A concern that I have with using a plugin such as AVP in combination with BYO-Secret-Manager is, that the users leveraging their own secrets manager would a) have to give us credentials for read access to their secrets management system and b) entrust us with the knowledge about the secrets during the processing of manifests.

ADR/adr-0008-secret-mgmt-for-user-workloads.md
gorkem: What do you think about the method described in [here](https://argocd-vault-plugin.readthedocs.io/en/stable/config/#multitenancy)?

ADR/adr-0008-secret-mgmt-for-user-workloads.md
gorkem: Bringing your own is not an immediate requirement, it is more of a thought exercise. We can remove it for now.

ADR/adr-0008-secret-mgmt-for-user-workloads.md
jannfis: I think that's more of a placebo multi-tenancy setup, to be honest. It _might_ be ok when the teams trust each other to a certain extent, however, we're in the fully untrusted domain here. Also, I was wondering what would prevent anyone from setting the `AVP_SECRET` in their `Application` resource to the name of another secret (considering that people usually have kind of control over their own `Application` resources).

Another factor is, the repository server itself should not have access to Kubernetes API in the first place for security reasons (that is why in the upstream's [default configuration](https://github.com/argoproj/argo-cd/blob/9fb7847f4e054b05b222a6cfabe4cd821d90f6e5/manifests/base/repo-server/argocd-repo-server-deployment.yaml#L19), access to Kubernetes for the repository server is disabled). We should avoid making it a requirement for the repository server to be able to talk to Kubernetes.

Also, the repository server does not have any concept of multi-tenancy, it just executes potentially untrusted processes on behalf of the application controller to render manifests and stream the results back for consumption. Now once we let users bring their own GitOps repositories, we won't have control over the repositories' contents anymore, leading to all kind of attack surfaces (Kustomize reading a file on the repository server or the sidecar, Helm reading a resource from Kubernetes, etc).

IMHO, we should not have our users' crown jewels handled by a third-party plugin. The plugin approach is fine for personal or non-critical purposes, but for large scale production environments with untrusted tenants such as ours, I'd suggest using an externalized secrets processing tool such as [external-secrets](https://github.com/external-secrets/external-secrets).

ADR/adr-0008-secret-mgmt-for-user-workloads.md
```diff
@@ -0,0 +1,101 @@
...
+    linkedTo:
+    - serviceAccount:
+        reference:
+          name: sa-star
+```
+
+
+### Security and Access Control
+
+* Access to the secret management backend should be restricted to authorized users and components.
```
arewm, line 90: Are there any criteria for who/what is an authorized user or component?

ADR/adr-0008-secret-mgmt-for-user-workloads.md
```diff
@@ -0,0 +1,101 @@
...
+* The default backend should use authentication and authorization mechanisms provided by the underlying service, such as Kubernetes or AWS IAM.
+* Workspace maintainers should have the ability to create and update secrets for their workspace on the default secret management backend.
```
arewm, line 92: Are these the maintainers of the individual workspaces within AppStudio? Would any user within AppStudio need to have accounts created for them on the `SecretStorage` provider or would the maintainers just create the resources with k8s and the system would populate them in the backend?

Based on the consequence below, I assume "no" to the latter that the users would just create the secrets in the UI and then the system would do all of the pushing.

ADR/adr-0008-secret-mgmt-for-user-workloads.md
```diff
@@ -0,0 +1,105 @@
...
+
+* When user workloads are deployed to environments, the system should be able to provide a way to inject values that are specific to the environment. Currently, this is done through environment variables that are managed as overlays on the GitOps repository for the application. However, this method does not provide a good way to manage `Secret`. This ADR addresses the secret management of user workloads for different environments.
+* As a StoneSoup component, I expect to have the ability to securely upload a secret, associate it with the component, and propagate it to the target environment.
+
+## Decision
+
+### Architecture Overview
+#### Terminology
+
+- `UploadSecret`: A short-lived Kubernetes secret used to deliver confidential data to permanent storage and link it to the RemoteSecret CR.
```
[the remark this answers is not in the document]
ralphbean: When I first heard your proposal, I was temporarily confused and thought that the UploadSecret was a new CR/CRD.

My suggestion: remove use of backticks for it to make it look less like a CR/CRD and more like a concept in bold.

ADR/adr-0008-secret-mgmt-for-user-workloads.md
```diff
@@ -0,0 +1,105 @@
...
+
+#### Example: If the destination is `Environment`
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: jdbc-connection-parameter
+  labels:
+    spi.appstudio.redhat.com/upload-secret: secret
+    spi.appstudio.redhat.com/target-environment: prod
```
ralphbean, line 40: I think this needs one more label, to be the `name` of the `Secret` that is put in place in the deployment target as a result of the `RemoteSecret`.

To see why, consider this problem:

* User has two environments (stage and prod) and has a password secret that needs to be available in both.
* The user's app expects to mount the secret from a secret called `password-secret`.
* In the user's workspace, they create an **Upload Secret** with target-environment set to stage. The upload secret has name "password-secret-stage".
* This is reconciled and a RemoteSecret is created, also with metadata.name set to "stage-password" and also with spec.secret.name set to "password-secret-stage".

When the user's app is deployed in the stage environment, it will try to mount a secret called "password-secret", but the secret (in the stage environment, put there by way of the RemoteSecret) will be named "password-secret-stage".

```suggestion
  labels:
    spi.appstudio.redhat.com/upload-secret: secret
    spi.appstudio.redhat.com/target-environment: prod
    spi.appstudio.redhat.com/target-secret-name: password-secret
```

ADR/adr-0008-secret-mgmt-for-user-workloads.md
ralphbean: I think we are moving ahead *without* the plugin - exploring new designs now. Closing this.

ADR/adr-0008-secret-mgmt-for-user-workloads.md
ralphbean: I think we are moving ahead *without* the plugin - exploring new designs now. Closing this.

ADR/adr-0008-secret-mgmt-for-user-workloads.md
arewm: Can we document this? Maybe it would be good to pull that table out of the ADR and put it elsewhere in the book so that it can be continued to be updated as role permission/authorization continues to be touched.

ADR/adr-0008-secret-mgmt-for-user-workloads.md
arewm: Thanks. Can we just note that here as well so that we don't have to look down to the consequences to understand?

ADR/adr-0008-secret-mgmt-for-user-workloads.md
metlos: If I read https://github.com/redhat-appstudio/book/blob/main/ADR/0011-roles-and-permissions.md correctly, maintainers don't have any permission to work with secrets.

metlos (commented):
Do we need to distinguish between users that are able to provide secrets to different environments?

gorkem:
> Do we need to distinguish between users that are able to provide secrets to different environments?

We do not have roles that are granular to environments. This is hard to implement with RBAC and is not a requirement

ADR/adr-0008-secret-mgmt-for-user-workloads.md
```diff
@@ -21,7 +21,7 @@ Accepted
 - `RemoteSecret`: A CRD that appears during upload and links `SecretData` + `DeploymentTarget(s)` + K8s `Secret`. `RemoteSecret` is linked to one (or zero) SecretData and manages its deleting/updating.
 - K8s `Secret`: What appears at the output and is used by consumers.
 - `SecretId`: A unique identifier of SecretData in permanent SecretStorage.
-- `SecretStorage`: A database eligible for storing `SecretData` (such as HashiCorp Vault, AWS Secret Manager).
+- `SecretStorage`: A database eligible for storing `SecretData` (such as HashiCorp Vault, AWS Secret Manager). That is an internal mechanism. Only spi-operator will be able to access it directly.
```
metlos, line 24: Technically, spi-operator needs full crud and oauth-service needs create permission in this storage. Not sure we need to be this detailed here though.

ralphbean (approved):

PR #28, opened 2022-12-06
StoneSoup workspace layout in post-KCP environment

alexeykazakov:


sbose78:
Would you be able to add some textual commentary which interprets the diagram, please?

ralphbean (approved):
Some text to describe this would be nice - but, it feels like we need an ADR titled "no more kcp", and this is the diagram that supports that ADR.

It's okay with me if we submit that separately.

alexeykazakov:
Yeah.. this PR is just to replace the outdate diagram. We sill need a proper ADR which I would do in a separate PR.

MatousJobanek (approved):
:+1: for phase 2
do we want to mention somewhere also the plan for SP (summit) - aka phase 1 or 1.5?

MatousJobanek:
> This looks good, but I think it's missing the Sandbox Host and Member Operators?

This is abstracted by the host & member clusters. We need to hide some of the details, otherwise, it would be hard to understand the diagram.

amfred:
> > This looks good, but I think it's missing the Sandbox Host and Member Operators?
> 
> This is abstracted by the host & member clusters. We need to hide some of the details, otherwise, it would be hard to understand the diagram.

Those operators are not any less important than the other operators in the diagram. In fact, they're quite important and they come up in conversations often.

PR #29, opened 2022-12-09
ADR-0010 Namespace metadata we will need

amfred:
The namespace labels are to help us in quickly searching logs across clusters and namespaces.
The metadata will help us meet this Red Hat requirement: SEC-NET-REQ-5 Monitor Outgoing Internet Traffic
Story: [ASC-186](https://issues.redhat.com/browse/ASC-186)

### DIFF
--- a/ADR/adr-0010-namespace-metadata
+++ b/ADR/adr-0010-namespace-metadata
@@ -15,21 +15,24 @@ Accepted
 
 ## Context
 
-We need metadata on our namespaces to make Stone Soup easier to operate and maintain. Standardizing namespace metadata will make it easier for us to search our logs and metrics across clusters. It will also allow us to enable logging for outgoing network traffic, one of our security requirements.
+We need metadata on our namespaces to make Stonesoup easier to operate and maintain. Standardizing namespace metadata will make it easier for us to search our logs and metrics across clusters. It will also allow us to enable logging for outgoing network traffic, one of our security requirements.
 
 ## Namespace labels
 
-We will apply the following labels to Stone Soup namespaces, to make them easier to find.
+We will apply the following labels to Stonesoup namespaces, to make them easier to identify programmatically:
 
-- `stonesoup.redhat.com/namespacetype: "controller"` for namespaces containing Stone Soup controllers
-- `stonesoup.redhat.com/namespacetype: "user-workspace-data"` for User workspaces where Applications, Components, and so on are defined
+- `stonesoup.redhat.com/namespacetype: "controller"` for namespaces containing controllers developed for Stonesoup. For example, we would annotate the `gitops-service-argocd` namespace but not the `openshift-gitops` namespace.
+- `stonesoup.redhat.com/namespacetype: "user-workspace-data"` for User workspaces where Applications, Components, and so on are stored
 - `stonesoup.redhat.com/namespacetype: "user-deployments"` for the namespaces where GitOps deploys applications for users
-- `stonesoup.redhat.com/namespacetype: "user-builds"` for the namespaces where the Pipeline Service has PipelineRuns
+- `stonesoup.redhat.com/namespacetype: "user-builds"` for the namespaces where the Pipeline Service manages users' PipelineRun resources
+
+The following labels are used for billing and telemetry. Values can be left blank if they are not defined yet:
+
 - `stonesoup.redhat.com/workspace_id: 12345678` a globally unique identifier for the workspace
-- `stonesoup.redhat.com/workspace_name: "test_workspace"` name for the workspace (need not be unique)
+- `stonesoup.redhat.com/workspace_name: "test_workspace"` a name for the workspace (need not be unique)
 - `stonesoup.redhat.com/external_organization: 11868048` the Red Hat orgId
 - `stonesoup.redhat.com/product_variant: "Stonesoup"` identifier for the type of product (allows tracking multiple variants in the same metric)
-- `stonesoup.redhat.com/sku_account_support: "Standard"` Standard, Premium, or Self-Support. Must match the value from the SKU
+- `stonesoup.redhat.com/sku_account_support: "Standard"` Standard, Premium, or Self-Support. Must match the value from the SKU.
 - `stonesoup.redhat.com/sku_env_usage: "Production"` Development/Test, Production, or Disaster Recovery. Must match the value from the SKU
 - `stonesoup.redhat.com/billing_model: "marketplace"` must be set to marketplace to indicate this service instance is billed through marketplace. This allows you to mark some instances as billed via marketplace (and some as not billed through marketplace)
 - `stonesoup.redhat.com/billing_marketplace: "aws"` which marketplace is used for billing
@@ -37,7 +40,7 @@ We will apply the following labels to Stone Soup namespaces, to make them easier
 
 ## Namespace annotations
 
-We will apply the following annotation to Stone Soup namespaces, which will enable OVN network logging to log outgoing network traffic:
+We will apply the following annotation to namespaces installed and maintained by Stonesoup on the clusters that Red Hat manages.  This will enable OVN network logging to log outgoing network traffic:
 
 metadata:
   annotations:
@@ -45,4 +48,4 @@ metadata:
 
 ## Consequences
 
-TBD
+We might have to migrate `stonesoup.redhat.com` to another product name in the future.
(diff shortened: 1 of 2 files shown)


ADR/adr-0010-namespace-metadata
```diff
@@ -0,0 +1,39 @@
...
+We need metadata on our namespaces to make Stone Soup easier to operate and maintain. Standardizing namespace metadata will make it easier for us to search our logs and metrics across clusters. It will also allow us to enable logging for outgoing network traffic, one of our security requirements.
+
+## Namespace labels
+
+We will apply the following labels to Stone Soup namespaces, to make them easier to find.
+
+- `stonesoup.redhat.com/namespacetype: "controller"` for namespaces containing Stone Soup controllers
+- `stonesoup.redhat.com/namespacetype: "user-workspace-data"` for User workspaces where Applications, Components, and so on are defined
+- `stonesoup.redhat.com/namespacetype: "user-deployments"` for the namespaces where GitOps deploys applications for users
+- `stonesoup.redhat.com/namespacetype: "user-builds"` for the namespaces where the Pipeline Service has PipelineRuns
```
MatousJobanek, line 27: As pointed out by @Michkov in this [slack thread](https://coreos.slack.com/archives/C02CTEB3MMF/p1670591996653489) - let's not introduce another group name, mainly when we know that the name is just an internal and temporal one. 
We already use `appstudio.redhat.com` all over the place, so let's stay with it.

ralphbean: +1 to keep it as stonesoup.redhat.com

Mention it in the Consequences that we're going to have to rename it again in the future.

ralphbean: Also, for Phase 1 - aren't user builds and the workspace data in the same namespace?

gorkem: Shared namespace does not change the labeling, both labels should be applied I suppose.

ADR/adr-0010-namespace-metadata
```diff
@@ -0,0 +1,39 @@
...
+
+## Reviewers
+
+## Context
+
+We need metadata on our namespaces to make Stone Soup easier to operate and maintain. Standardizing namespace metadata will make it easier for us to search our logs and metrics across clusters. It will also allow us to enable logging for outgoing network traffic, one of our security requirements.
+
+## Namespace labels
+
+We will apply the following labels to Stone Soup namespaces, to make them easier to find.
```
gorkem, line 22: Needs a definition of what is a "StoneSoup namespace"? For instance does that also include namespaces that we deploy to on BYOC clusters.

gorkem (changes requested):
Can you also add the 'Namespace labels' section to `book/index.md`? In case other ADRs introduces new ones It would be good to have a  single place to see them.

ADR/adr-0010-namespace-metadata
MatousJobanek: @ralphbean do you really want to change the labels (and group name) every time when the product/project name changes? A proper migration from one name to another can be very painful (even if we do it only for labels). Using a mix of group names within one project can be confusing, so I would rather stay with one group name, even if it's not perfect and can become outdated. 
In an ideal world, we would use a name that is not bound to a project/product name, so any future changes won't affect us.

ADR/adr-0010-namespace-metadata
```diff
@@ -0,0 +1,39 @@
...
+
+- `stonesoup.redhat.com/namespacetype: "controller"` for namespaces containing Stone Soup controllers
```
MatousJobanek, line 24: could you please define what you mean by "StoneSoup controller"? Do count also controllers of the OpenShift Pipeline operator (Tekton)? How about ArgoCD and Sandbox controllers?

ADR/adr-0010-namespace-metadata
ralphbean: > do you really want to change the labels (and group name)

No, no I do not.

> A proper migration from one name to another can be very painful ... Using a mix of group names within one project can be confusing,

I agree.

>  In an ideal world, we would use a name that is not bound to a project/product name, so any future changes won't affect us.

Oo, why not pursue this now?  Could we use `cloud.redhat.com` or `apps.redhat.com` or `apps.cloud.redhat.com` or `cicd.redhat.com`?

ADR/adr-0010-namespace-metadata
ralphbean: I think it means tekton, yes. I think it does _not_ mean the `openshift-gitops` namespace, but it does mean the `gitops-service-argocd` namespace.

ADR/adr-0010-namespace-metadata
amfred: Because this went back and forth I'm not inclined to change it right now.

ADR/adr-0010-namespace-metadata
amfred: > 

OK I added a "Naming conventions" section to the index.
I also added more details on which namespaces need the annotations.

ADR/adr-0010-namespace-metadata
amfred: I added additional details.

amisstea:
@alexeykazakov @ralphbean, as [we discussed](https://redhat-internal.slack.com/archives/C02CTEB3MMF/p1675176475445669), billing metadata has been added to this ADR. Please take a look.

ADR/adr-0010-namespace-metadata
```diff
@@ -0,0 +1,54 @@
...
+We need metadata on our namespaces to make Stonesoup easier to operate and maintain. Standardizing namespace metadata will make it easier for us to search our logs and metrics across clusters. It will also allow us to enable logging for outgoing network traffic, one of our security requirements.
+
+## Namespace labels
+
+We will apply the following labels to Stonesoup namespaces, to make them easier to identify programmatically:
+
+- `appstudio.redhat.com/namespacetype: "controller"` for namespaces containing controllers developed for Stonesoup. For example, we would annotate the `gitops-service-argocd` namespace but not the `openshift-gitops` namespace.
+- `appstudio.redhat.com/namespacetype: "user-workspace-data"` for User workspaces where Applications, Components, and so on are stored
+- `appstudio.redhat.com/namespacetype: "user-deployments"` for the namespaces where GitOps deploys applications for users
+- `appstudio.redhat.com/namespacetype: "user-builds"` for the namespaces where the Pipeline Service manages users' PipelineRun resources
```
alexeykazakov, line 30: Let's explicitly mention that one namespace can have multiple types.

alexeykazakov (commented):
I can't really comment on the billing labels because I miss the context. For example how exactly the workspace_id is going to be used? It would help to understand the meaning of those labels.

But the namespace types look fine to me. If a single namespace can have multiple types.

ADR/adr-0010-namespace-metadata
amfred: Added.  I will ask @hmariset to come back and add a little more context on how the billing team will use the labels.  I assume they will be the ones to apply them, or that they will contribute the code to do so.

hmariset: stonesoup-o11y team will copy these labels and attach them to all the data we aggregate for billing.
The SKUs data probably needs to be propagated from the UI during the user onboarding.

ADR/adr-0010-namespace-metadata
```diff
@@ -0,0 +1,55 @@
...
+
+The following labels are used for billing and telemetry. Values can be left blank if they are not defined yet:
+
+- `appstudio.redhat.com/workspace_id: 12345678` a globally unique identifier for the workspace
+- `appstudio.redhat.com/workspace_name: "test_workspace"` a name for the workspace (need not be unique)
+- `appstudio.redhat.com/external_organization: 11868048` the Red Hat orgId
```
ralphbean, line 37:
```suggestion
- `appstudio.redhat.com/external_organization: 11868048` the Red Hat orgId of the account that created the workspace
```

Is this what is meant?

ralphbean (approved):

ADR/adr-0010-namespace-metadata
hmariset: Yes

alexeykazakov (approved):
Overall looks good.
We might want to update the billing labels when we have more clarity of how we are going to use them but we can do it in separate PRs.

ADR/adr-0010-namespace-metadata
amfred: OK let's assume that actually populating the labels will have its own feature and epics, as opposed to designing it on the fly here.

PR #30, opened 2022-12-09
added some missing indexes

amfred:


ADR/index.md
```diff
@@ -0,0 +1,48 @@
+# Architecture Decision Records
```
ralphbean, line 1: We currently get an index automatically generated for this by the script: https://redhat-appstudio.github.io/book/ADR/index.html

We need to fix up the names and titles of the ADRs so that it looks clean, but if we do then we won't have to maintain a manual index.md file.

ralphbean: The script is in https://github.com/redhat-appstudio/book/blob/main/.github/workflows/generate-adr-index.yml

gorkem (changes requested):
Can we keep depending on the generated ADR index and not create new maintenance?

ADR/index.md
ralphbean: See also https://github.com/redhat-appstudio/book/pull/32

amfred:
OK, this is superceded by https://github.com/redhat-appstudio/book/pull/39 and one other.

PR #31, opened 2022-12-14
Environment Provisioning

ralphbean:
For https://issues.redhat.com/browse/STONE-174


ADR/adr-0008-environment-provisioning.md
```diff
@@ -0,0 +1,428 @@
...
+
+![](../diagrams/ADR-0008/provision-loop.jpg)
+
+##### Deprovision Loop (DT Reconciliation)
+
+![](../diagrams/ADR-0008/deprovision-loop.jpg)
+
+#### EnvironmentController
+
+TODO - anything to write about the Environment controller?
```
ralphbean, line 187: @gbenhaim I think this section got omitted in the original doc.

ADR/adr-0008-environment-provisioning.md
```diff
@@ -0,0 +1,428 @@
...
+- [Discussion recording, 2022-12-07](https://drive.google.com/file/d/1UWIefLzFb2rYnz8mvvWkNKJyuWy-_nhm/view)
+- [Original drafting in googledoc](https://docs.google.com/document/d/1vFD5lDbxek_Q2RZyKdFuZ8z0xYubOA1i8cu2K7KNxaQ/edit#heading=h.k2fng7c0dm7k)
+- [Original miro](https://miro.com/app/board/uXjVP77ztI4=)
+- [Issue tracking creation of this ADR (STONE-174)](https://issues.redhat.com/browse/STONE-174)
+
+[Environment]: ../ref/application-environment-api.md#environment
+[Environments]: ../ref/application-environment-api.md#environment
+[DT]: ../ref/application-environment-api.md#deploymenttarget
+[DTs]: ../ref/application-environment-api.md#deploymenttarget
+[DeploymentTarget]: ../ref/application-environment-api.md#deploymenttarget
```
ralphbean, line 421: This url anchor won't be valid until something actually gets implemented in the application-api repo.

ADR/adr-0008-environment-provisioning.md
```diff
@@ -0,0 +1,428 @@
...
+- At minimum, create the CRDs and make them available, with no binder or provisioner controllers behind them.
+- Next, modify the Environment CRD and teach the gitops service how to navigate from the linked [DTC] to the [DT] in order to find the Secret that it needs for Argo.
+
+Minimal functionality looks like the BYOC case: No provisioners
+necessary - and even the binder can be omitted in this first pass.
+
+- On workspace initialization, have the NsTemplateTier:
+  - Create Spaces with fixed -dev and -stage suffixes manually as a part of the appstudio tier
+  - Create serviceaccounts with rolebindings to administer the namespaces
+  - Create [DeploymentTargets] to match and reference Secrets that will be created to contains tokens for the serviceaccounts in the previous steps.
```
ralphbean, line 276: Note, we need to resolve this blocker around Phase 0 here. It doesn't appear to be possible using only static tier templates to get the token from one namespace and make it available as a Secret in another namespace.

ADR/adr-0008-environment-provisioning.md
```diff
@@ -0,0 +1,428 @@
+# ADR-0008 Environment Provisioning
+
+Date: 2022-12-14
+
+## Status
+
+Accepted
```
ralphbean, line 7: Not actually accepted yet. Pending review in this PR.

ADR/adr-0008-environment-provisioning.md
```diff
@@ -0,0 +1,428 @@
...
+
+## Approvers
```
ralphbean, line 9: These people haven't actually approved yet. Will seek their approval before merging.

ADR/adr-0008-environment-provisioning.md
```diff
@@ -0,0 +1,428 @@
...
+
+```yaml
+apiVersion: appstudio.redhat.com/v1alpha1
+kind: DeploymentTarget
+metadata:
+    name: prod-dt
+spec:
+    deploymentTargetClassName: isolation-level-namespace
+    kubernetesCredentials:
+        defaultNamespace: team-a--prod-dtc
```
amfred, line 70: Just out of curiosity, what does this naming convention mean? I assume team-a is just an imaginary org; what's the rest?

ADR/adr-0008-environment-provisioning.md
```diff
@@ -0,0 +1,428 @@
...
+
+#### DeploymentTargetBinder
+
+Binds [DeploymentTargetClaim] to a [DeploymentTarget] that satisfies its requirements.
+
+It watches for the creation of new [DTC] and tries to find a matching [DT] for each one of them.
+
+A [DT] that we created dynamically for a specific [DTC] will always be attached to it.
+
+[DT] and [DTC] have one to one bindings.
```
amfred, line 162: What would this look like if we had multiple workspaces using the same cluster? Would you create multiple deployment targets for the same cluster?
And if we had multiple environments using the same cluster within a workspace, would we also have multiple deployment targets for that cluster?

ADR/adr-0008-environment-provisioning.md
```diff
@@ -0,0 +1,428 @@
...
+### DeploymentTarget(DT) and DeploymentTargetClaim(DTC) Lifecycle
+
+![](../diagrams/ADR-0008/dt-dtc-lifecycle.jpg)
+
+### Access Matrix
+
+| *User,Controller/CRD* | **DT**                | **DTC**               | **DTCLS** | **Environment**      |
+|-----------------------|-----------------------|-----------------------|-----------|----------------------|
+| **Binder**            | watch/list/get/update | watch/list/get/update |           |                      |
+| **Provionser**        | create/delete         | watch                 | get       |                      |
```
amfred, line 198: What has the ability to create/delete/update DTCLS?

amfred: typo: Provisioner

ADR/adr-0008-environment-provisioning.md
amfred: This is something I was starting to get at above.  It feels like we need something else that's storing the cluster data so we can re-use clusters across multiple workspaces.  For the new-user workspace initialization, we need to copy the information about the Sandbox cluster into the various DeploymentTargets.

Later when a tenant can have multiple workspaces, the workspaces for a single tenant will often share clusters. We shouldn't require users to re-enter BYOC information for every workspace.  We need to be able to define the cluster information once at the tenant level and copy or share that information across multiple workspaces.

ADR/adr-0008-environment-provisioning.md
ralphbean: Yeah, the convention isn't important, in that nothing depends on parsing this. It could be a uuid or something similar. It just needs to be unique.

But, as the doc was coming together, I was assuming that this would be put together from the `<name of the user's appstudio namespace>--<name of the deploymenttargetclaim in the user's appstudio namespace that caused the new deployment namespace to be created>`

ADR/adr-0008-environment-provisioning.md
ralphbean: In terms of the user's appstudio namespace where the DT and the DTC live, there is also a Secret that contains the kubeconfig-formatted serviceaccount token to admin the deployment namespace. It is no more persistent than an ordinary Secret in k8s.

ADR/adr-0008-environment-provisioning.md
ralphbean: Only we (appstudio administrators) would create new DTCLS resources when we want to make a new kind of deployment target available for automatic provisioning.

ADR/adr-0008-environment-provisioning.md
gbenhaim: I assume that a `tenant` equals a user. If so, in order to allow sharing BYOC clusters between workspaces, we will have to associate them with the user that added them to Stonesoup, otherwise, BYOC clusters can leak between workspaces. For example:

Alice is part of workspace A
Bob is part of workspaces, A and B
Alice adds cluster C to workspace A
Can Bob use cluster C in workspace B?

ADR/adr-0008-environment-provisioning.md
ralphbean: > typo: Provisioner

Fixed.

ADR/adr-0008-environment-provisioning.md
```diff
@@ -0,0 +1,428 @@
...
+|--------------------|-------------------------------------------|-------------|------------|-------------------------------------------------------------------------------|
+| finalizerDT        | provionser.appstudio.redhat.com/finalizer | Provisioner | DT         | Delays the deletion of a DT so the provisioner can free up external resources |
+
+### Use Case Descriptions
+
+- **During onboarding** - when a user requests a new appstudio tier namespace, the tier template includes two [Environments], and two [DTCs]. The [Environments] reference the [DTCs]. The [DTCs] bear a request for the “devsandbox” [DTCLS]. The devsandbox provisioner responds to that request and generates a SpaceRequest, ultimately resulting in a new namespace for each environment. The SpaceRequest is marked ready by the spacerequest controller. The devsandbox deployment target provisioner controller sees that and marks the devsandbox [DT] as ready. The deployment target binder sees that, and attaches the new [DTs] to the [DTCs]. The environment controller sees this and marks the [Environments] as ready.
```
gbenhaim, line 220: do we need to elaborate on what it means to mark the environment as ready?

ralphbean: I think we went back and forth about whether or not to propagate things from the DTC and the DT back to the Environment. What entity will be waiting for that status on the Environment? I'm not sure we have a consumer for that. If no consumer, then perhaps we can drop this sentence.

ralphbean: Updated in a4349a4fb0fb28ece4c2a1a5401c91c29b568065

ADR/adr-0008-environment-provisioning.md
alexeykazakov: For tenants, orgs etc we will have to introduce workspace hierarchy when we can store shared resources like DTC/DT on the parent workspace level.

@ralphbean the problem with creating DTs with a secret reference via tier templates is a pretty big problem. I'm not sure we can work it around without introducing very ugly stonesoup specific hacks in the sandbox core code for MVP.

ADR/adr-0008-environment-provisioning.md
```diff
@@ -0,0 +1,458 @@
...
+### Phase 0 (Path to MVP)
+
+- At minimum, create the CRDs and make them available, with no binder or provisioner controllers behind them.
+- Next, modify the Environment CRD and teach the gitops service how to navigate from the linked [DTC] to the [DT] in order to find the Secret that it needs for Argo.
+
+Minimal functionality looks like the BYOC case: No provisioners
+necessary - and even the binder can be omitted in this first pass.
+
+- On workspace initialization, have the NsTemplateTier:
+  - Create Spaces with fixed -dev and -stage suffixes manually as a part of the appstudio tier
```
alexeykazakov, line 304:
```suggestion
  - Create Namespaces with fixed -dev and -stage suffixes manually as a part of the appstudio tier. So every provisioned workspace/space has three namespaces. The main user namespace and two -dev and -stage for environments.
```

ADR/adr-0008-environment-provisioning.md
ralphbean: On the problem of creating the secret reference in the appstudio namespace, to me, the choice is between these two:

* Do we invent and accept some hack to make Phase 0 happen?
* Do we ignore Phase 0 and go straight for Phase 1? We can avoid spending time on a temporary hack, but will delay MVP.

To resolve that, we need some estimate of how long the SpaceRequest implementation will take - which looks like the biggest item in Phase 1.

gorkem (changes requested):
A few typos and a couple of clarifications that I needed. Otherwise looks like it is ready. Phase 3 probably needs more thought but it is something that we need to do anyway.

ADR/adr-0008-environment-provisioning.md
ralphbean: OK - an update from conversation today with @alexeykazakov and @MatousJobanek

* What we have right now for the Environment API technically works for static environments for MVP target.
* Argo has a privileged service account that it can use to manage any namespace on the cluster.
* We can defer rolling out this new Environments API until after MVP. We will essentially combine the steps from Phase 0 and Phase 1 into a larger Phase 1.

ADR/adr-0008-environment-provisioning.md
```diff
@@ -0,0 +1,458 @@
...
+apiVersion: appstudio.redhat.com/v1alpha1
+kind: DeploymentTarget
+metadata:
+    name: prod-dt
+spec:
+    deploymentTargetClassName: isolation-level-namespace
+    kubernetesCredentials:
+        defaultNamespace: team-a--prod-dtc
+        apiURL: …
+        clusterCredentialsSecret: team-a--prod-dtc--secret
```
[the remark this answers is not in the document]
ralphbean: A reference to a Secret, containing a single `kubeconfig` key, value pair.

ADR/adr-0008-environment-provisioning.md
```diff
@@ -0,0 +1,458 @@
...
+- Adjusting the integration service controller to create and delete a [DTC]
+
+### Phase 2 (need more grooming and clear requirement from the PM)
+
+The result of this phase is the ability to specify parametes that are
+needed from the deployment target, such has memory, CPU, CPU
+architecture and the number of nodes.
+
+- The [DT], [DTC] an [DTCLS] would need to extended to support the new parameters.
+  - **DTCLS - allowTargetResourceExpansion** - represents whether or not the underlying provisioner allows targets to be resized. I.e., for a whole cluster provided by the "hypershift" ocm provider - can it have its number or size of nodes increased without having to delete it and create a new one.
```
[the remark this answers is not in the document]
ralphbean: It's not important at all for early Phases. We suspect it might be useful in future phases, in the same way that it's useful to be able to scale up a PVC requested size for storageClasses that support it. However - there are a lot more kinds of parameters here for "size" than for storage. (Think limits vs requests vs node count vs node size and more..). We originally had a sketch of how that might work but it needs its own followup ADR to explore it fully. "Needs more grooming and input from PM."

ADR/adr-0008-environment-provisioning.md
```diff
@@ -0,0 +1,458 @@
...
+**Immutable object**: no
+
+**Scope**: namespace
+
+```yaml
+apiVersion: appstudio.redhat.com/v1alpha1
+kind: DeploymentTargetClaim
+metadata:
+    name: prod-dtc
+spec:
```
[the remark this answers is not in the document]
ralphbean: No `selector` support as it is currently designed. As designed:

* A StorageClass references exactly one *provider*. Multiple StorageClasses can reference the same provider.
* A DeploymentTargetClaim references exactly one StorageClass, by name in the `deploymentTargetClassName`.

With selection, I suppose you could claim a target matching _any_ storage class or (as you suggest) claim a target matching any of the storage classes that provide cluster-level isolation (but not those that provide only namespace isolation).

---

In the current design, the binding controller sets an annotation on the DTC with the provisioner it found on the associated StorageClass. Provisioners only process DTCs annotated for them.

If we added a `storageClassSelector` field later in addition to the `storageClassName, we could add support for processing that selection in the binding controller. I hesitate to add it now, to avoid adding any more complexity to the initial design than necessary.

ADR/adr-0008-environment-provisioning.md
ralphbean: Updated in 5469aeb.

gorkem (approved):
LGTM

ADR/adr-0008-environment-provisioning.md
```diff
@@ -0,0 +1,464 @@
...
+
+In our old KCP architecture, we had [a
+design](https://docs.google.com/document/d/1WKd1FVHAxaNQKCIzIW-vUQRgsoOP9T-8rYozAMDpYc0/edit#) for
+provisioning a new deployment target in support of new [Environments].  This design was to be
+implemented in [GITOPSRVCE-228](https://issues.redhat.com/browse/GITOPSRVCE-228) by an environment
+controller that would create and manage sub-workspaces of the user’s main AppStudio workspace, and
+that would provide a serviceaccount to Argo in order to deploy the user’s application to those
+sub-workspaces. Now, without KCP, we need a new design.
+
+The [Environment] CR serves two purposes:
```
sbose78, line 35: Is `Environment` supposed to be a link ?

jgwest (approved):
LGTM, great stuff @ralphbean and @gbenhaim (et al), thanks for all your hard work in really nailing down the complex technical details around DT/DTC/DTClass  :heart:

ADR/adr-0008-environment-provisioning.md
```diff
@@ -0,0 +1,464 @@
...
+
+### Use Case Descriptions
+
+- **During onboarding** - when a user requests a new appstudio tier namespace, the tier template includes two [Environments], and two [DTCs]. The [Environments] reference the [DTCs]. The [DTCs] bear a request for the `devsandbox` [DTCLS]. The devsandbox provisioner responds to that request and generates a `SpaceRequest`, ultimately resulting in a new namespace for each environment. The spacerequest controller provisions a serviceaccount in the new namespace and places a kubeconfig including the token for this serviceaccount in the origin namespace. The `SpaceRequest` in the origin namespace is updated to reference that `Secret` and is marked ready by the `SpaceRequest` controller. The devsandbox deployment target provisioner controller sees that and creates the devsandbox [DT] referencing the `Secret` and marks the devsandbox [DT] as ready. The deployment target binder sees that, and attaches the new [DTs] to the [DTCs]. The environment controller sees this and creates a [GitOpsDeploymentManagedEnvironment] that references the `Secret` found from traversing from the [Environment] to the [DTC] to the [DT].
```
sbose78, line 250: Instead of introducing a devsandbox provisioner and the SpaceRequest API/Controller : 
Could we implement a "free-tier-osd-provisioner" controller which takes a pool of OSDs and gets us namespaces ? The same could be used against Sandbox OSD clusters too, what do you think ?

ADR/adr-0008-environment-provisioning.md
```diff
@@ -0,0 +1,464 @@
...
+
+**Pending:** DT is not yet available for binding.
+
+**Available:** DT waits for a Claim to be bound to.
+
+**Bound:** The DT was bounded to a DTC.
+
+**Released:** The DT was previously bound to a DTC which got deleted. external resources were not freed.
+
+**Failed:** DT was released from its claim, but there was a failure during the release of external resources.
```
elsony, line 95: Is the failure to release the only case that will cause it in the failed phase? What about the failed case where DT failed during creation in the first place, e.g. secret is incorrect?

ADR/adr-0008-environment-provisioning.md
```diff
@@ -0,0 +1,464 @@
...
+
+#### DeploymentTargetBinder
+
+Binds [DeploymentTargetClaim] to a [DeploymentTarget] that satisfies its requirements.
+
+It watches for [DTC] and objects and tries to find a matching [DT] for each one of them. In addition, it's responsible for updating the `Status` sections of the [DT] and [DTC] (it also watches [DTC] objects). It marks [DTC] objects that requires dynamic provisioning.
+
+A [DT] that was created dynamically for a specific [DTC] will always be attached to it.
+
+[DT] and [DTC] have one to one bindings.
```
elsony, line 190: I think it make sense to have DTC only maps to one DT. Is there any case where we can have more than one DTC maps to the same DT?

ADR/adr-0008-environment-provisioning.md
```diff
@@ -0,0 +1,464 @@
...
+- **For manual creation of new Environments** - a user submits a form in HAC which creates a new Environment CR and a new [DTC] CR. The Environment CR references the [DTC] CR, which is reconciled as in the previous bullet.
+- **For automated testing in ephemeral environments** - a user specifies an IntegrationTestScenario CR with an existing Environment to clone. After a build completes, but before it executes tests, the integration-service creates a new Environment CR and a new [DTC] CR with the devsandbox [DTCLS] as above, and references the [DTC] from the Environment. The integration-service should delete the [DTC] once the environment isn’t needed anymore for the test.
+- **BYO cluster** - A user creates a [DT] and a [DTC] and `Secret`. The [DT] has the details and a reference to the secret used to connect to his/hers cluster. In addition, it contains the name of the [DTC] it should be bounded to. The user then refer to the [DTC] from the Environment that should use it.
```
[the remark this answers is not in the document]
ralphbean: Only to satisfy API requirements - because the Environment expects a DTC (just like a pod expects a pvc). In a way, it is API overhead.

ADR/adr-0008-environment-provisioning.md
ralphbean: It is. The link references are at the _very_ end of the file.

ADR/adr-0008-environment-provisioning.md
```diff
@@ -0,0 +1,464 @@
...
+
+**Scope**: namespace
+
+```yaml
+apiVersion: appstudio.redhat.com/v1alpha1
+kind: DeploymentTarget
+metadata:
+    name: prod-dt
+spec:
+    deploymentTargetClassName: isolation-level-namespace
```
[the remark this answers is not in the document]
ralphbean: It is because in the general case, the `binder` controller's job is to find matching DTs for DTCs, and bind them together. In theory, a DT could be pre-created but sitting around idle. You should be able to create a DTC that requests a certain kind of DT. If there is an existing DT that matches and if it is not yet bound to any other DTCs, then the binder controller should bind _it_ to the DTC.

In order to do that matching, DTs need to declare what storage class they are created from. I.e., if the DTC wants `isolation-level-namespace`, binding controller needs to look for any unbound DTs that provide `isolation-level-namespace`.

ADR/adr-0008-environment-provisioning.md
```diff
@@ -0,0 +1,464 @@
...
+
+### Phase 1
+
+The result of this phase is automation that binds [DT] and [DTC], and automatically provisions
+an environment on the Sandbox cluster (using the
+[SpaceRequest
+API](https://docs.google.com/document/d/1uqgghk1lN9dyoBLsvn5YD443TwKrsKHU6fvfUuo84Hs/edit#)) and
+creates a [DT].
+
+- Ceate the [DT], [DTC], and [StorageClass] CRDs and make them available.
```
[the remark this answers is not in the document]
ralphbean: Yes! Fixing.

ADR/adr-0008-environment-provisioning.md
ralphbean: Yes, although the provisioner needs to be coded to manage specific kinds of external resources. The devsandbox provisioner described here knows specifically how to create and manage `SpaceRequests`. If we want to get namespaces from another pool of OSD clusters that are _not_ managed by codeready toolchain operators, then we'll need a provisioner that knows how to do that _without_ `SpaceRequests`. The proposal gives us freedom to implement one or both, exposing each as different `DeploymentTargetClasses`.

FWIW, we already have devsandbox member clusters available and assuming that the SpaceRequest API really will be implemented - I think we should build against it for the first provisioner implementation.

ADR/adr-0008-environment-provisioning.md
ralphbean: I think in that case, the DT should not have been created by the provisioner at all. Instead, the provisioner should emit an `Event` complaining that it cannot create a DT that is being requested by a DTC for some reason - it should describe the reason in the Event.

ADR/adr-0008-environment-provisioning.md
alexeykazakov: > Could we implement a "free-tier-osd-provisioner" controller which takes a pool of OSDs and gets us namespaces 

This is basically what Dev Sandbox does in nutshell :)
So, you are asking if instead of using Dev Sandbox we could re-implement dev sandbox in the "free-tier-osd-provisioner". Of course we could. But why? :)

ADR/adr-0008-environment-provisioning.md
ralphbean: I don't have a concrete use case at the moment. We can perhaps imagine a use case where there is some single fixed cluster that we want to make available to multiple teams. We might hope to model the cluster as one DT and then let multiple DTCs be bound to it.

Or, instead maybe the single cluster should get multiple DTs created for it and the multiple DTCs should bind exclusively to each of those. This would at least encourage a model where the multiple DTCs do not share a Secret reference (of course, I guess they could all use the same Secret name - but at least the model encourages that kind of separation).

gbenhaim: At this point, I don't see a reason to have a one-to-many relationship between a DT and a DTC. If we want to use a different namespace on a target cluster, each namespace can be represented with its own DT.

ADR/adr-0008-environment-provisioning.md
```diff
@@ -0,0 +1,464 @@
...
+
+##### Deprovision Loop (DT Reconciliation)
+
+![](../diagrams/ADR-0008/deprovision-loop.jpg)
+
+#### EnvironmentController
+
+Watch for [Environment] objects. If the [Environment] references a [DTC] that is `Bound` and that
+[DTC] references a [DT] that contains a `spec.kubernetesCredentials.clusterCredentialsSecret` field,
+then ensure that there exists a [GitOpsDeploymentManagedEnvironment] resource that also references
```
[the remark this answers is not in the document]
gbenhaim: I think it's better to have two different API's with a controller that bridges between them rather than a single API that will create a coupling between the controllers. It's expected that in the future the DT will contain more details about the cluster/namespace it represents, for example, its CPU architecture, and I don't think that this information should interest the GitopsService.

ADR/adr-0008-environment-provisioning.md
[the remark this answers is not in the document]
sbose78: Got it, sounds very reasonable to me.

ADR/adr-0008-environment-provisioning.md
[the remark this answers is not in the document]
sbose78: Thank you @gbenhaim, fair point.

The reason I feel having many APIs which wrap cluster credentials is that there is a long-term API-maintenance overhead even though short-term it is quicker to get things done. 

The description of a Kubernetes deployment target (Cluster properties and cluster credentials) would always be the same - with a subset of properties being optional.

ADR/adr-0008-environment-provisioning.md
jduimovich: > we could re-implement dev sandbox in the "free-tier-osd-provisioner". Of course we could. But why? :) 

So customers could create private "devsandboxes" the same way WE do as part of a BYOC cluster pool they want to control ? Otherwise our devsandbox is provisioned/configured in a special way and differently than the perhaps customers would try to create their own free-tier thing.  Why do we need two ways to do it ?

sbose78: > if we want to get namespaces from another pool of OSD clusters that are not managed by codeready toolchain operators, then we'll need a provisioner that knows how to do that without SpaceRequests.

Yes, this ^

My reframed question is, should we implement a provisioner that is generic enough to be used for provisioning namespaces from devsandbox OSDs / non-devsandbox OSDs ( or even non-openshift clusters ). How significant would that effort be ? We'll need to do that anyway sooner than later.

sbose78 (approved):
This looks great! A lot of good work done in a very short amount of time. 💪

There are a couple of areas which could be revisited in the future :

1. We'd need to implement a more generic provisioner for provisioning namespaces from devsandbox OSDs and non-devsandbox OSDs. Therefore, we should consider revisiting the amount of investment in the devsandbox-specific provisioner.

2. `DeploymentTarget` and `GitOpsDeploymentManagedEnvironment` are very similar with respect to the information being wrapped. Bringing down the API surface by merging similar APIs might be a worthwhile effort in the future.

elsony (approved):
LGTM

ADR/adr-0008-environment-provisioning.md
alexeykazakov: Sandbox already provide a lot of useful functionality which such a provisioner would need to re-implement. SpaceRequest API is our next step towards making Sandbox is even more generic and flexible so it can be used as "Namespace as a Service" in more use cases, such as the namespace provisioner to create/manage namespaces in different types of clusters (control plane clusters, workload clusters, etc).

It's not only about provisioning a namespace. We need a service which would provide and *manage* namespaces in a fleet of clusters. It's not that trivial.

ralphbean:
Thanks for the review and iteration everyone!

PR #37, opened 2022-12-22
Update integration-service.md content in book

srivickynesh:
This PR comes from [HACBS Test workstream - Integration Service](https://docs.google.com/document/d/1ZggNV3wTcZBFkbMs49eeQUG3vdJJXDDH3fIvfjhVPS8/edit#)

https://github.com/srivickynesh/book/blob/HACBS-1371/book/integration-service.md can be referred to review this PR.

Signed-off-by: Sri Vignesh <sselvan@redhat.com>

### DIFF
--- a/book/integration-service.md
+++ b/book/integration-service.md
@@ -1,0 +1,213 @@
+# Integration Service
+
+
+## Overview
+
+The Integration Service is composed of controllers that facilitate automated testing of content produced by the build pipelines. It is mainly responsible for integration testing capabilities.
+
+The Integration service uses the pipeline, snapshot, and environment controllers to watch for Component builds, trigger and manage testing Tekton Pipelines, and create releases based on the testing outcome(s).
+
+### Goals
+
+- The Integration service need to be able to deploy a specific set of images to an environment and record the results of integration testing.
+- Given a specific set of images, Integration service should be able to tell if they passed application validation tests.
+- The Integration service should be able to update one component image at a time in order to test and deploy individual component builds.
+- The Integration service should have a capability to automatically promote sets of passed images.
+- To be SLSA compliant, Integration service should also be able to tell what version of the build and test pipelines were used to create each component image.
+
+### High Level Workflow
+
+- When a build pipeline completes, Integration service creates a Snapshot CR representing a new collection of components that should be tested together.
+- When Integration Service sees a new Snapshot CR (created either by itself or by a user), it coordinates deployment of the application for testing. It does this by creating new “ephemeral” StoneSoup Environment CRs which trigger the GitOps Service to provision the ephemeral test environment.
+- After the environment is ready and the application snapshot has been deployed to it, Integration Service tests and validates the application according to user-provided configuration. It does this by executing Tekton PipelineRuns against the ephemeral test Environment.
+- When all of the test Tekton PipelineRuns have passed, Integration service triggers a deployment of the application to all Environments which are not labeled as "ephemeral" and which do not have a `parentEnvironment` set. For most users, this will result in a deployment to their "development" environment (or to their "staging" environment, if they do not have a dev environment).
+- Finally, if any automatic ReleasePlans have been created by the user in the workspace, it will create a Release CR signalling intent to release the tested content, to be carried out by the [release-service](./release-service.md).
+
+The diagram below shows the interaction of the integration service and other services.
+
+![](../diagrams/hacbs-data-flow.jpg)
+
+### Dependencies
+
+The [Integration Service](./integration-service.md) is dependent on the following services:
+- [Pipeline Service](./pipeline-service.md)
+  - Pipeline execution, Pipeline logging
+- [GitOps Service](./gitops-service.md)
+  - Provides the facility to create
+    - Snapshots defining sets of Builds to test
+    - Environment to test the Application on
+    - SnapshotEnvironmentBindings to have the Snapshot of the Application deployed to a specific Environment
+- [Hybrid Application Service](./hybrid-application-service.md)
+  - Provides the Application and Component model. Integration Service updates the pullspec reference on the Component CR when a component passes its required testing.
+- [Release Service](./release-service.md)
+  - Provides the ReleasePlan that will be used to determine if integration-service should create a Release
+- [Enterprise Contract Service](./enterprise-contract.md)
+  - Provides facilities to validate whether content has passed the Enterprise Contract.
+
+
+## Terminology
+
+**Component Pipeline** - This is the test pipeline run for the Component of an Application that gets triggered by a completed Build pipeline.
+
+**Composite Pipeline** - This is the pipeline run that runs if two or more Components have completed and passed their Component pipeline concurrently so that they should be tested together.
+
+**Global Candidate List** - The list of all Component digest that have passed individual component testing. This can be retrieved from the Application to see what Components it is made of and then querying each of the Components `spec.containerImage`. 
+
+**Snapshot** -  The custom resource that contains the list of all Components of an Application with their Component Image digests. Once created, the list of Components with their images is immutable. The Integration service updates the status of the resource to reflect the testing outcome.
+
+**IntegrationTestScenario** - The custom resource that describes separate PipelineRuns that are supposed to be run by the snapshot controller. The IntegrationTestScenarios can be marked as optional, in which case they will not be taken into account when determining if the Snapshot has passed testing. Any number of IntegrationTestScenarios can be set by the user and also the user can specify the contexts where they will be applied (Component or Composite stage - or both).
+
+**Ephemeral Environment** - The ephemeral copy of an existing environment that is created by the Integration service for use during a specific Integration pipeline run.
+
+**SLSA** - [SLSA](http://slsa.dev/) is a new PSSC initiative. Current goal is to reach [SLSA](http://slsa.dev/) level 4 or above for all services.
+
+## Outdated Terminology
+
+**Prelim Pipeline** - This is the same as the Component pipeline. Going forward the Preliminary Pipeline will be referred as Component Pipeline.
+
+**Final Pipeline** -  This is the same as the Composite Pipeline. Going forward the Final Pipeline will be referred as Composite Pipeline.
+
+## Resources 
+Below are the list of CRs that the integration service is responsible for interacting and performing operations:
+
+### CREATE
+
+| Custom Resource | When? | Why? |
+|---|---|---|
+| Snapshot | Post Component build PipelineRun completes.  | Used as the source input for the test gate pipeline |
+| PipelineRuns | Post Snapshot creation  | To test the Snapshot  |
+| Release | Post PipelineRun if there is a ReleasePlan for the Application, if the auto-release flag is set on the ReleasePlan | To signal to the Release Service to automatically push the Snapshot to the Production Environment |
+| SnapshotEnvironmentBinding | When deploying the Snapshot to an Environment | Create a new SnapshotEnvironmentBinding for the Dev/Stage Environment if a matching SnapshotEnvironmentBinding doesn’t already exist |
+
+### READ
+
+| Custom Resources | When? | Why? |
+|---|---|---|
+| Application & Component | Before creating the Snapshots | To know which ImageSpecs to use in the Snapshot |
+| IntegrationTestScenario | Before creating the Component/Composite PipelineRun(s) | To get the information for the PipelineRun - which Tekton bundles and Environments will be used |
+| ReleasePlan | Before creating the Release | To signal the Release Service for next environment promotion (prod) |
+
+### UPDATE
+
+| Custom Resource | When? | Why? |
+|---|---|---|
+| SnapshotEnvironmentBinding | Post Test PipelineRun | To promote the Snapshot to the next Environment (Dev/Stage) |
+| Snapshot | Post Test PipelineRun | To mark the snapshot validated so that it can be promoted to the next environment |
+| Component | Post Test PipelineRun | To update the Component with the image spec, in order to update the global candidates list |
+
+### WATCH
+
+| Custom Resource | When? | Why? |
+|---|---|---|
+| PipelineRun | Post Build PipelineRun | Read the component and the output image to create the Snapshot and start the Preliminary PipelineRun |
+| PipelineRun | Post Component PipelineRun | Check the result of all Component PipelineRuns and mark the linked Snapshot as it either passed all tests or failed them |
+| PipelineRun | Post Composite PipelineRun | Check the result of all Composite PipelineRuns and mark the linked Snapshot as it either passed all tests or failed them |
+| Snapshot | Upon Snapshot creation | Check the Snapshot details and start running Integration PipelineRuns for it |
+| Snapshot | Upon Snapshot status being updated as passed all tests | Follow up on the global candidate list state to possibly create the Composite Snapshot. Otherwise check ReleasePlan and create a Release. Update SnapshotEnvironmentBindings if the Dev/Stage Environments exist |
+| SnapshotEnvironmentBinding (To be implemented) | Upon the Snapshot being deployed to the ephemeral Environment | Start the Integration PipelineRun to run tests on the Snapshot that was deployed to the ephemeral Environment |
+
+### Pipeline
+
+The test pipeline will consist of a single definition that could be executed as two logically distinct pipeline runs by the Integration Service. The logical pipeline runs are:
+
+- Component pipeline run
+- Composite pipeline run
+
+The Component pipeline run is always triggered for each Component of an Application and is triggered by a completed build of a component.  
+
+The Composite pipeline run is only triggered when it has been determined by the Integration Service that two or more Components of an Application have been successfully built concurrently by their respective Component pipeline and need to be tested together.
+
+### PipelineRun
+
+The Integration Service will generate a PipelineRun that can reference the user provided tekton pipeline.
+
+#### Pipeline Reference
+The pipeline reference identifies the target Tekton Pipeline that the PipelineRun will run. An application team should be able to offer its Tekton bundle at a well-known place, according to the pipeline reference. The pipeline packaging will be retrieved when the IntegrationTestScenario is executed by the Integration service.
+
+#### Annotations/Labels
+
+Following the [annotation guidelines](https://docs.google.com/document/d/1gyXM3pkKFFfxHZnopBi_53vREFWhwA0pFUuIhopDuEo/edit#), the Integration Service sets the below annotations on PipelineRuns.
+```
+“pipelines.appstudio.openshift.io/type": "test",
+"test.appstudio.openshift.io/test":      “<component|composite>”
+“test.appstudio.openshift.io/scenario":  “<IntegrationTestScenario name>”
+"appstudio.openshift.io/snapshot":       “<snapshot name>”
+“appstudio.openshift.io/component":      “<component name>”
+“appstudio.openshift.io/application":    “<Application name>”
+```
+The Integration service will copy the annotations and labels from the Build PipelineRun and append those to the Test PipelineRuns for traceability across the system per [Labels and Annotations for StoneSoup pipelines](https://docs.google.com/document/d/1fJq4LDakLfcAPvOOoxxZNWJ_cuQ1ew9jfBjWa-fEGLE/edit#) and [StoneSoup builds and tests PRs](https://docs.google.com/document/d/113XTplEWRM63aIzk7WwgLruUBu2O7xVy-Zd_U6yjYr0/edit#)
+
+#### Workspace
+
+The Integration service will provide Tekton workspaces to the individual PipelineRuns to supply additional resources required for testing, such as secure credentials for accessing test Environments.
+
+#### Persistent Volume
+
+ A persistent volume that the pipeline can store test artifacts generated from the test execution.
+
+#### Secrets
+
+The Integration service needs secrets mounted so that the `Environment Provisioner` Task running in the workspace has the correct permissions to create AppStudio Application Environments resources within the Applications/Component’s namespace.
+
+## Detailed Workflow
+1. Watch for Build PipelineRuns of `type: build`
+    - Extract  Component Name, Application Name, and Image from pipeline annotations
+2. Query the Application 
+    - for each Component extract the `spec.containerImage`
+3. Create a Snapshot
+    - Populate the `spec.components` list with the component name and the `spec.containerImage` with information from Step 1 and 2, replacing the container image for the built component with the one from the build PipelineRun
+4. Create PipelineRuns for each IntegrationTestScenario
+    - Fetch the IntegrationTestScenario for the application/component to get the Tekton Bundle and Environment information
+    - Assign annotations of
+        ```
+        "test.appstudio.openshift.io/test":   component
+        "test.appstudio.openshift.io/snapshot":   “<snapshot name>”
+        “test.appstudio.openshift.io/component":   “<component name>”
+        “test.appstudio.openshift.io/application":   “<Application name>”
+        ```
+    - Pass in the Snapshot json representation as a parameter
+    - Optional: Create the ephemeral Environment if the IntegrationTestScenario specifies one. After the Environment is ready, link it to the pipelineRun and start it
+5. Watch the PipelineRun of `test: component` and `component: <component name>`
+    - When all PipelineRuns complete
+        - Check if all the PipelineRuns associated with the snapshot have passed successfully
+        - If all PipelineRuns passed, mark the Snapshot as validated by setting its status condition `HACBSTestsSucceeded` as true
+        - If not all PipelineRuns passed, mark the Snapshot as not validated by setting its status condition `HACBSTestsSucceeded` as false, end the Integration testing process
+6. Update the Component `spec.containerImage` 
+7. Query the Application and find all of its components
+    - For each Component extract the `spec.containerImage`
+8. Prepare a new Snapshot
+    - Populate the `spec.components` list and the `spec.containerImage` with information from Step 5.
+9. Compare new Snapshot from step 8 to step 3
+    - If determined the same, proceed to step 10
+    - If determined not the same, proceed to step 14
+10. Update the spec.Snapshot of SnapshotEnvironmentBinding(s) for lowest environments with the `<snapshot name>`
+11. Query ReleasePlan `spec.application` for the specific application in question
+    - Check the auto-release flag on the ReleasePlan too
+    - If ReleasePlan for Application found proceed to step 12
+    - If ReleasePlan for Application NOT found proceed to step 13
+12. Create a Release with the `spec.ReleasPlan` and `spec.Snapshot`
+13. Done - Repeat from Step 1 again
+14. If not the same, repeat step 4
+    - Assign the same annotations except 
+        Test: composite
+        Snapshot: name from step 8
+    - Pass in the Snapshot name as a parameter
+15. Watch the PipelineRun of `test: composite` and `snapshot: <snapshot name>`
+    - When PipelineRuns complete
+        - If all PipelineRuns passed, mark the Snapshot as validated by setting its status condition HACBSTestsSucceeded as true
+16. Repeat step 10
+17. Repeat step 11-13
+
+
+## Appendix
+
+- [StoneSoup Promotion & Environment API](https://docs.google.com/document/d/14LaXAmQEW73kIr3a6TvPswT-zSdBsuaaxLF77HJ3gX4/edit#)
+- [v3 - Environment API draft: Component-scoped with native Snapshots](https://docs.google.com/document/d/1-_rWLgALd5pdSlqNNcQ5FSrD00fZb0l_exU-_FiL68o/edit#heading=h.vf66svd4o6xr)
+- [StoneSoup builds and tests PRs](https://docs.google.com/document/d/113XTplEWRM63aIzk7WwgLruUBu2O7xVy-Zd_U6yjYr0/edit#)
+- [Labels and Annotations for StoneSoup pipelines](https://docs.google.com/document/d/1fJq4LDakLfcAPvOOoxxZNWJ_cuQ1ew9jfBjWa-fEGLE/edit#)
+- [Implementation design and rules for pipeline customization](https://docs.google.com/document/d/1PXkpFHKrnq1Sg1giTgeXdYzNVf7CTRgwowykr7YPM2I/edit#)
+
+
+## References
+
+Originally drafted in a [google document](https://docs.google.com/document/d/1ZggNV3wTcZBFkbMs49eeQUG3vdJJXDDH3fIvfjhVPS8/edit#)


book/integration-service.md
```diff
@@ -0,0 +1,271 @@
...
+  - Provides the Application and Component model. Integration Service updates the pullspec reference on the Component CR when a component passes its required testing.
+- [Release Service](./release-service.md)
+  - Provides the ReleasePlan that will be used to determine if integration-service should create a Release
+- [Enterprise Contract Service](./enterprise-contract.md)
+  - Provides facilities to validate whether content has passed the Enterprise Contract.
+
+
+## Terminology
+
+**Prelim Pipeline** - This is the test pipeline run for the Component of an Application that gets triggered by a completed Build pipeline. Deprecated name, refer to the Component Pipeline
```
ralphbean, line 49: A suggestion: maybe reorganize this into a `## Terminology` section that contains only current terminology, but then add a section just below it titled `### Outdated Terminology` that mentions the old Prelim and Final terms.

(As a reader - I gotmomentarilly confused when the _first_ term I encountered was deprecated.)

book/integration-service.md
```diff
@@ -0,0 +1,271 @@
...
+|---|---|---|
+| PipelineRun | Post Build PipelineRun | Read the component and the output image to create the Snapshot and start the Preliminary PipelineRun |
+| PipelineRun | Post Preliminary PipelineRun | Check the result of the Preliminary PipelineRun and  follow up on the global candidate list state to either trigger the Final PipelineRun or check ReleaseLink and create a Release |
+| PipelineRun | Post Final PipelineRun | Check the result of the Final PipelineRun, check ReleaseLink and create a Release |
+
+### Pipeline
+
+The test pipeline will consist of a single definition that could be executed as two logically distinct pipeline runs by the Integration Service. The logical pipeline runs are:
+
+- Component (Prelim) pipeline run
```
ralphbean, line 109: Remove the deprecated terminology here. It is old enough that we don't need to keep mentioning it down here.

ralphbean (approved):

book/integration-service.md
```diff
@@ -0,0 +1,271 @@
+# Integration Service
+
+
+## Overview
+
+The Integration Service is composed of a couple controllers that facilitate automated testing of content produced by the HACBS build pipelines. It is responsible for integration testing capabilities.
```
arewm, line 6: Is it literally a couple of controllers -- i.e. two controllers? I think we can just remove `a couple`.

book/integration-service.md
```diff
@@ -0,0 +1,271 @@
...
+
+To support this capability, it needs to do a couple of things.
+
+- When a build pipeline completes, it creates an Snapshot CR representing a new collection of components that should be tested together.
```
arewm, line 10: creates `a` Snapshot CR

book/integration-service.md
```diff
@@ -0,0 +1,271 @@
+# Integration Service
+
+
+## Overview
+
+The Integration Service is composed of a couple controllers that facilitate automated testing of content produced by the HACBS build pipelines. It is responsible for integration testing capabilities.
+
+To support this capability, it needs to do a couple of things.
```
arewm, line 8: You use `it` many times in the bullets. Can we change this one to the integration service?

book/integration-service.md
```diff
@@ -0,0 +1,271 @@
...
+
+- When a build pipeline completes, it creates an Snapshot CR representing a new collection of components that should be tested together.
+- When it sees a new Snapshot CR (created either by itself or by a user), it coordinates deployment of the application for testing. It does this by creating a new “ephemeral” StoneSoup Environment CRs which should in turn trigger the GitOps Service to provision environment (and a sub-workspace in CPS)
```
arewm, line 11: How many CRs are created? If one, change to `StoneSoup Environment CR`. If multiple, change to `creating new "ephemeral"`.

Remove `should in turn`.

What is `CPS`? I do not see it defined.

book/integration-service.md
```diff
@@ -0,0 +1,271 @@
...
+- When it sees the environment is ready, it triggers Tekton PipelineRuns supplied by the user that run application tests and validation against the app running in that ephemeral Environment.
```
arewm, line 12: Change the beginning to `After the environment is ready`. 

Does this wording inaccurately represent the service or fail to clarify its role? 
```
it deploys, tests, and validates the application in that ephemeral Environment according to user-provided configuration```

File: book/integration-service.md [hunk 8]
```diff
@@ -0,0 +1,271 @@
...
+- When it sees that all of the test PipelineRuns that it created have passed, it finishes its work by triggering a deployment of the application to the lowest known “static” environment defined by the user, usually staging.
```
arewm, line 13: ```After all test and validation PipelineRuns pass, the integration service will then trigger an application deployment to the lowest defined "static" environment (for example, staging).```

arewm: Is it important/pivotal that the tests and validation are PipelineRuns? I really only see references to that in this list (and my suggestion on the previous item removed the reference). If this is important to highlight, I feel like we should be making a bigger call-out of it elsewhere.

book/integration-service.md
```diff
@@ -0,0 +1,271 @@
...
+- Finally, if any automatic ReleasePlans have been created by the user in the workspace, it will create a Release CR signalling intent to release the tested content, to be carried out by the release-service (separate threat model).
```
arewm, line 14: Change `(separate thread model)` to a link to the release service.

book/integration-service.md
```diff
@@ -0,0 +1,271 @@
...
+
+
+### Goals
```
arewm, line 17: These feel like they are trying to be user stories.
Can the pronouns (we/you) be changed to specific personas?

book/integration-service.md
```diff
@@ -0,0 +1,271 @@
...
+
+
+### Dependencies
+
+The [Integration Service](./integration-service.md) is dependent on the following services:
+- [Pipeline Service](./pipeline-service.md)
+  - Pipeline execution, Pipeline logging
+- [GitOps Service](./gitops-service.md)
+  - Provides the facility to create
+    - Snapshots used to test Builds
```
arewm, line 36: I feel like it is more appropriate to say that Snapshots define a set of builds to test

book/integration-service.md
```diff
@@ -0,0 +1,271 @@
...
+    - Environment to test the Application on
+    - SnapshotEnvironmentBindings to have the Snapshot deployed to a specific Environment
```
arewm, line 38: Similarly, is the snapshot deployed to the environment or would it be more appropriate to say that the applications defined in the snapshot are deployed?

book/integration-service.md
arewm: I feel like I would go further. Since deprecated names are only really used in the early days, I would imagine that they would be most likely to only be looked up by searching. I think we would be able to remove them from the terminology and just indicate under the replacement terminology what other names it has been called in the past. If no references to these deprecated names can be found, I wouldn't include them in the terminology.

```
**Component Pipeline** - This is the test pipeline run for the Component of an Application that gets triggered by a completed Build pipeline. This is the replacement for the deprecated **Prelim Pipeline** which should no longer be used.
```

book/integration-service.md
```diff
@@ -0,0 +1,271 @@
...
+
+**Final Pipeline** - This is the pipeline run that runs if two or more Components have completed and passed their Component pipeline concurrently so that they should be tested together. Deprecated name, refer to Composite Pipeline.
+
+**Composite Pipeline** - This is the same as the Final Pipeline. Going forward the Final Pipeline will be referred to as Composite Pipeline
+
+**Global Candidate List** - The list of all Component digest that have passed individual component testing. This can be retrieved from the Application to see what Components it is made of and then querying each of the Components `spec.containerImage`. 
+
+**Snapshot** - The list that maps all Components of an Application to their Component Image digests. Once created, the snapshot is immutable. 
+
+**Snasphot Digest** - The snapshot itself should have an identifier constructed from its contents - using a digest of the contents likely makes sense.
```
arewm, line 61: Can you tell us what the digest is instead of what should happen or what makes sense?

book/integration-service.md
```diff
@@ -0,0 +1,271 @@
...
+
+**Integration Service** - The service that uses the pipeline, snapshot and environment controllers to watch for Component builds, trigger and manage the testing pipelines and create releases based on the testing outcome(s).
```
arewm, line 63: I like the directness of this definition. While I understand why a different language style might have been used at the beginning of this document, can we use some of this up there? If we define it properly at the beginning, do we need to also define it in the terminology section?

book/integration-service.md
```diff
@@ -0,0 +1,271 @@
...
+
+**IntegrationTestScenario** - The custom resource that describes separate PipelineRuns that are supposed to be run by the snapshot controller. Any number of them can be set by the user and each of them includes information about the Tekton pipeline that will be executed - Tekton bundle, pipeline name and parameters, and the Environment details. The Integration Test Scenarios can also specify the contexts where they will be applied - this can be used to define when the scenario will run (Component or Composite stage - or both). If more than one Integration Test Scenario has the same context, their respective PipelineRuns will be executed in parallel.
```
arewm, line 65: Is this the main way that users will interface with/configure the integration service? You say that any number of "them" (I assume you mean the defined PipelineRuns, but it would be good to clarify this) can be set by the user -- is this the only customization that a user will have when interacting with the service?

book/integration-service.md
```diff
@@ -0,0 +1,271 @@
...
+
+**Ephemeral Environment** - The ephemeral copy of an existing environment that’s created by the Integration service for use during a specific Integration pipeline run. After the environment is copied and marked as ephemeral, the Integration service creates a new SnapshotEnvironmentBinding linking the Snapshot under test with the newly created Environment. This triggers a deployment of the Snapshot on the new Environment, handled by the GitOps service. Finally, the Integration service links the credentials for the Environment to the Integration Tekton pipeline run and starts the test run.
```
arewm, line 67: The snapshot is just defined as a list above. We are not deploying a list to the environments but an immutable set of components.

book/integration-service.md
arewm: Ah, I might not be accurate according to the words in the ephemeral environment. Integration service doesn't do a deployment, but the GitOps service deploys a set of application components as defined by a snapshot.

book/integration-service.md
arewm: I think that you have included important details in this bullet, but I feel like most of it would be better to include elsewhere -- potentially in the steps carried out by integration service. This should be a definition of an ephemeral environment not a description of what actions are being applied to the environment after creation. If too much detail for the overview, would some place like the generalized workflow be better?

book/integration-service.md
arewm: Are we getting into too much detail for an overview? Would it make sense to simplify this and further break out the details for the various actions elsewhere?

book/integration-service.md
```diff
@@ -0,0 +1,271 @@
...
+
+
+## Resources 
+Below are the list of CR’s that the integration service is responsible for interacting and performing operations:
```
arewm, line 71: CR's or CRs?

book/integration-service.md
```diff
@@ -0,0 +1,271 @@
...
+| Application & Component | Before creating the Snapshots | To know which build.ImageSpecs to use in the Snapshot |
+| IntegrationTestScenario | Before creating the Component/Composite PipelineRun(s) | To get the information for the PipelineRun - which Tekton bundles and Environments will be used |
+| ReleaseLink | Before creating the Release | To signal the Release Service for next environment promotion (prod) |
+
+### UPDATE
+
+| Custom Resource | When? | Why? |
+|---|---|---|
+| SnapshotEnvironmentBinding | Post Test PipelineRun | To promote the Snapshot to the next environment (Stage/Stable) |
+| Snapshot | Post Test PipelineRun | To mark the snapshot validated so that it can be promoted to the next environment |
```
arewm, line 94: Isn't the snapshot was immutable once created?

book/integration-service.md
```diff
@@ -0,0 +1,271 @@
...
+| Component | Post Test PipelineRun | To update the Component with the image spec, in order to update the global candidates list |
+
+### WATCH
+
+| Custom Resource | When? | Why? |
```
arewm, line 99: Is there an Environment CR that is being watched? That is the impression that I had from the overview.

book/integration-service.md
```diff
@@ -0,0 +1,271 @@
...
+The test pipeline will consist of a single definition that could be executed as two logically distinct pipeline runs by the Integration Service. The logical pipeline runs are:
+
+- Component (Prelim) pipeline run
+- Composite (Final) pipeline run
+
+The Component pipeline run is always triggered for each Component of an Application and is triggered by a completed build of a component.  
+
+The Composite run is only triggered when it has been determined by the Integration Service that two or more Components of an Application have been successfully built concurrently by their respective Component pipeline and need to be tested together. 
+
+The two stage model is roughly modeled on the 'fast loop / slow loop' model that OpenShift uses.  ART is the slow loop.  The global candidates list at some point in time is the set of artifacts that the Composite pipeline will test and mark as a 'shippable' snapshot.
```
arewm, line 116: I don't think this comparison with OpenShift/ART will be understood by many. What is the fast loop in this example? If we will stick with the reference, can you define ART?

book/integration-service.md
```diff
@@ -0,0 +1,271 @@
...
+“pipelines.appstudio.openshift.io/type": "test",
+"test.appstudio.openshift.io/test":      “<component|composite>”
+“test.appstudio.openshift.io/scenario":  “<IntegrationTestScenario name>”
+"appstudio.openshift.io/snapshot":       “<snapshot name>”
+“appstudio.openshift.io/component":      “<component name>”
+“appstudio.openshift.io/application":    “<Application name>”
+```
+We must also copy the annotations and labels from the Build PipelineRun and append those to the Test PipelineRuns for traceability across the system per [Labels and Annotations for StoneSoup pipelines](https://docs.google.com/document/d/1fJq4LDakLfcAPvOOoxxZNWJ_cuQ1ew9jfBjWa-fEGLE/edit#) and [StoneSoup builds and tests PRs](https://docs.google.com/document/d/113XTplEWRM63aIzk7WwgLruUBu2O7xVy-Zd_U6yjYr0/edit#)
+
+#### Workspace
```
arewm, line 140: Did you mean to include anything under workspace?

book/integration-service.md
```diff
@@ -0,0 +1,271 @@
...
+
+The Composite run is only triggered when it has been determined by the Integration Service that two or more Components of an Application have been successfully built concurrently by their respective Component pipeline and need to be tested together. 
+
+The two stage model is roughly modeled on the 'fast loop / slow loop' model that OpenShift uses.  ART is the slow loop.  The global candidates list at some point in time is the set of artifacts that the Composite pipeline will test and mark as a 'shippable' snapshot.
+
+[Draft: StoneSoup-4 Pipeline & Task Design](https://docs.google.com/document/d/1ucQI3WJzkLpueYLuzHEXpKNvYQafDl4gNBZryvi2d5o/edit#)
+
+### PipelineRun
+
+The Integration Service will generate a PipelineRun that can reference the user provided tekton pipeline.
```
arewm, line 122: Are the subcategories all properties that the integration will set? Will need access/permissions to? Is any of it user facing/customizable or is it just a description of the service's interface?

book/integration-service.md
```diff
@@ -0,0 +1,271 @@
...
+[Draft: StoneSoup-4 Pipeline & Task Design](https://docs.google.com/document/d/1ucQI3WJzkLpueYLuzHEXpKNvYQafDl4gNBZryvi2d5o/edit#)
+
+### PipelineRun
+
+The Integration Service will generate a PipelineRun that can reference the user provided tekton pipeline. 
+
+#### Pipeline Reference
+The pipeline reference should point to a well known location that an Application team can provide their tekton bundle. The pipeline packaging will be retrieved the IntegrationScenario CR
+
+#### Annotations/Labels
```
arewm, line 127: I assume that this section is just stating what annotations/labels the service is setting. Instead of stating what "we" can or must do, just simply state what is being done.

book/integration-service.md
```diff
@@ -0,0 +1,271 @@
...
+        Test: composite
+        Snapshot: name from step 8
+    - Pass in the Snapshot name as a parameter
+15. Watch the PipelineRun of `test: composite` and `snapshot: <snapshot name>`
+    - When PipelineRun completes
+        - Mark the Snapshot `spec.artifacts.attribute` as validated
+16. Repeat step 10
+17. Repeat step 11-13
+
+Recover
```
arewm, line 196: Is this an accident?

book/integration-service.md
```diff
@@ -0,0 +1,271 @@
...
+
+![](../diagrams/hacbs-data-flow.jpg)
+
+
+## Open Questions
+### Determining Test Pass/Failure
+#### Current
+Current miro design board shows the integration controller determining test failure status. It looks like it’s required to gate in the following scenarios:
+  - Triggering of Composite (final) PipelineRun
```
arewm, line 205: remove deprecated terminology here as well.

book/integration-service.md
dirgim: While the number of controllers is currently two, we are in the (ongoing) process of adding additional ones (one for scenario management and one for environment deployments), so a couple might be the most appropriate at this time.

book/integration-service.md
dirgim: This is a common issue in our docs unfortunately, since initially it was supposed to be only one CR. Based on the current design, multiple CRs will be created, so I agree with the proposal to drop the `a` from the `a new "ephemeral"` part.

book/integration-service.md
dirgim: It's a Snapshot of the single application. This sentence admittedly used to make more sense when the CR was called ApplicationSnapshot. Maybe we can change it to `Snapshot of the Application`?

book/integration-service.md
dirgim: Yes, configuring Integration test scenarios is the only direct way to interact with the service from the user's side.
We can change the `Any number of them` to `Any number of IntegrationTestScenarios` to clarify the statement.

book/integration-service.md
dirgim: The list of components and their digests is immutable, the status of the Snapshot CR is what's updated here.
Maybe we can clarify that by changing `To mark the snapshot validated` to `To update the Snapshot's status to validated`.

book/integration-service.md
dirgim: This part is in flux as the Environment APIs are under design currently and being expanded.
We are currently experimenting with watching `SnapshotEnvironmentBindings` instead of `Environments` to determine if the `Snapshot` was deployed to a given `Environment`, but this is not yet finalized as far as we can tell.

book/integration-service.md
dirgim: That's a good point. We have a separate [HACBS-1426](https://issues.redhat.com/browse/HACBS-1426) ticket to document the two stage model, so it would probably be best to provide a very brief description here and have a link to the resulting document.

book/integration-service.md
dirgim: The `(and a sub-workspace in CPS)` part can be removed entirely. It was referring to the KCP-related part of the design and I overlooked it when refactoring.

book/integration-service.md
dirgim: Yes, I think that this point is accurate in its current state, we can change the beginning to be `After the environment is ready` and leave the rest as-is.

book/integration-service.md
dirgim: It is important because the tests and validations are executed by running Tekton PipelineRuns so it shapes how the tests are defined and configured by the user.

book/integration-service.md
dirgim: This is actually a proposed item that we've been designing to optimize finding existing snapshots on the cluster.
It's not implemented yet, so I think we can drop this line entirely until we implement it.

book/integration-service.md
dirgim: Yes, this is correct - all of this information is set by the Integration service.
This section is mostly directly useful to other Stone Soup workflows - some of this information will be bubbled up to the users by way of UI, and some power users might directly fetch it in the future.

book/integration-service.md
```diff
@@ -0,0 +1,271 @@
...
+Following the [guidelines](https://docs.google.com/document/d/1gyXM3pkKFFfxHZnopBi_53vREFWhwA0pFUuIhopDuEo/edit#) in we can define some basic ones to start
+```
+“pipelines.appstudio.openshift.io/type": "test",
+"test.appstudio.openshift.io/test":      “<component|composite>”
+“test.appstudio.openshift.io/scenario":  “<IntegrationTestScenario name>”
+"appstudio.openshift.io/snapshot":       “<snapshot name>”
+“appstudio.openshift.io/component":      “<component name>”
+“appstudio.openshift.io/application":    “<Application name>”
+```
+We must also copy the annotations and labels from the Build PipelineRun and append those to the Test PipelineRuns for traceability across the system per [Labels and Annotations for StoneSoup pipelines](https://docs.google.com/document/d/1fJq4LDakLfcAPvOOoxxZNWJ_cuQ1ew9jfBjWa-fEGLE/edit#) and [StoneSoup builds and tests PRs](https://docs.google.com/document/d/113XTplEWRM63aIzk7WwgLruUBu2O7xVy-Zd_U6yjYr0/edit#)
```
dirgim, line 138: We can change `We must also copy` to `The Integration service will copy`.

book/integration-service.md
```diff
@@ -0,0 +1,271 @@
...
+
+#### Workspace
+#### Persistent Volume
+ A persistent volume that the pipeline can store test artifacts generated from the test execution.
+
+#### Secrets
+
+We may need secrets mounted so that the `Environment Provisioner` Task running in the workspace has the correct permissions to create AppStudio Application Environments resources within the Applications/Component’s namespace
```
dirgim, line 146: We can change `We may need secrets mounted` to `The Integration service needs secrets mounted`.

book/integration-service.md
dirgim: I addressed the specific usages of `we` in separate comments where appropriate.

book/integration-service.md
dirgim: I think we can add a general sentence to the tune of:
```
The Integration service will supply Tekton workspaces to the individual PipelineRuns to provide additional resources required for testing, such as secure credentials for accessing test Environments.
```
We will need to flesh out this section in the future as more details get implemented.

book/integration-service.md
arewm: I still think it is simpler/clearer to just remove that as it is redundant since `controllers` indicates more than one controller.

book/integration-service.md
arewm: ack. I guess just something to keep in mind to come back to.

book/integration-service.md
arewm: Yeah, that would work.

book/integration-service.md
arewm: I would also suggest changing the following line

```
Following the [guidelines](https://docs.google.com/document/d/1gyXM3pkKFFfxHZnopBi_53vREFWhwA0pFUuIhopDuEo/edit#) in we can define some basic ones to start
```
to something like
```
Following the [annotation guidelines](https://docs.google.com/document/d/1gyXM3pkKFFfxHZnopBi_53vREFWhwA0pFUuIhopDuEo/edit#), the Integration Service sets the following annotations on PipelineRuns
```

Too bad that these google documents are not ADRs to be able to link to directly here. :-p

book/integration-service.md
arewm: It is an annoying balance to try to maintain between being specific enough for clarity but not too specific so that your documentation gets outdated with the slightest changes...

book/integration-service.md
arewm: Trying again to improve clarity here. How about something like this?
> After the environment is ready and the application snapshot has been deployed to it, the integration service tests and validates the application according to user-provided configuration.

book/integration-service.md
arewm: Cool. If Tekton PipelineRuns are a pivotal part of the technology, I think we should at least call it out in the initial part of the overview.

book/integration-service.md
amfred: It would indeed be helpful to replace "we" with more specific things like "a developer" or "the build system".

book/integration-service.md
```diff
@@ -0,0 +1,264 @@
...
+- We need to be able to deploy a specific set of images to an environment and record the results of integration testing.
+- Given a specific set of images, we should be able to tell if they passed application validation tests.
+- We should be able to update one component image at a time in order to support 'slow loop' application validation testing.
+- We should have a capability to automatically promote sets of passed images.
+- To be SLSA compliant, we should also be able to tell what version of the build and test pipelines were used to create each component image.
+
+### High Level Workflow
+
+- When a build pipeline completes, Integration service creates a Snapshot CR representing a new collection of components that should be tested together.
+- When Integration service sees a new Snapshot CR (created either by itself or by a user), it coordinates deployment of the application for testing. It does this by creating new “ephemeral” StoneSoup Environment CRs which trigger the GitOps Service to provision environment.
```
amfred, line 21:
```suggestion
- When Integration service sees a new Snapshot CR (created either by itself or by a user), it coordinates deployment of the application for testing. It does this by creating new “ephemeral” StoneSoup Environment CRs which trigger the GitOps Service to provision the ephemeral test environment.
```

book/integration-service.md
```diff
@@ -0,0 +1,264 @@
...
+- After the environment is ready and the application snapshot has been deployed to it, the integration service tests and validates the application according to user-provided configuration. It does this by executing Tekton Pipelineruns against the ephemeral Environment.
```
amfred, line 22:
```suggestion
- After the environment is ready and the application snapshot has been deployed to it, the integration service tests and validates the application according to user-provided configuration. It does this by executing Tekton PipelineRuns against the ephemeral test environment.
```

book/integration-service.md
```diff
@@ -0,0 +1,264 @@
...
+- When all of the test Tekton PipelineRuns have passed, Integration service triggers a deployment of the application to the lowest known “static” environment defined by the user, usually staging.
```
amfred, line 23: What does "static" mean and why isn't this the development environment?

book/integration-service.md
```diff
@@ -0,0 +1,264 @@
...
+- Finally, if any automatic ReleasePlans have been created by the user in the workspace, it will create a Release CR signalling intent to release the tested content, to be carried out by the [release-service](./release-service.md).
+
```
amfred, line 25: Is there another step before this where the integration service de-provisions the ephemeral test environment? Is the test environment left up and running for some period of time or is it de-provisioned immediately? 
Where do the logs go (especially if the test environment is quickly de-provisioned)?

book/integration-service.md
```diff
@@ -0,0 +1,264 @@
...
+### Proposal
+We should be consistent and continue to design the Test pipelines so that the tasks don’t fail and continue to produce the HACBS_TEST_OUTPUT.
+
+#### Option 1
+leverage the enterprise contract task to run the test validation rego policy against this pipeline.
+
+#### Option 2
+
+If being a part of the Enterprise Contract is NOT possible and not clear. We should still implement the same technology and mechanism we do now with EC. What we can do is extend the IntegrationTestScenario spec to have a testPolicyWhitelist field that a user defines per Scenario. This Policy will actually be a git repo to a data file that can be used by a conftest/opa rego file.
```
amfred, line 228: Red Hat is trying to replace blacklist/whitelist with blocklist/allowlist instead. Hopefully it's not too late to change the field name?

book/integration-service.md
```diff
@@ -0,0 +1,264 @@
...
+
+We would need to
+  - The Test Task Generates the HACBS_TEST_OUPTUT
+    - We may need to expand on the Task Output to accommodate more complex use case than the existing Sanity Tests
+
+  - We define a default test [policy](https://github.com/hacbs-contract/ec-policies/blob/main/policy/release/test.rego) for the Integration Service similar to how EC did. 
+  - We define a data policy format to use with this policy
+
+We submit a TaskRun post pipeline run to validate the Pipeline Run output that includes the HACBS_TEST_OUTPUT
+
```
amfred, line 238: There's that sneaky "we" again - is the integration service handling all of this?

book/integration-service.md
```diff
@@ -0,0 +1,264 @@
+# Integration Service
+
+
+## Overview
+
+The Integration Service is composed of controllers that facilitate automated testing of content produced by the build pipelines. It is mainly responsible for integration testing capabilities.
+
+The Integration service uses the pipeline, snapshot and environment controllers to watch for Component builds, trigger and manage the testing Tekton Pipelines and create releases based on the testing outcome(s).
```
arewm, line 8:
```suggestion
The Integration service uses the pipeline, snapshot, and environment controllers to watch for Component builds, trigger and manage the testing Tekton Pipelines, and create releases based on the testing outcome(s).
```

book/integration-service.md
```diff
@@ -0,0 +1,264 @@
...
+
+## Terminology
+
+**Component Pipeline** - This is the test pipeline run for the Component of an Application that gets triggered by a completed Build pipeline.
+
+**Composite Pipeline** - This is the pipeline run that runs if two or more Components have completed and passed their Component pipeline concurrently so that they should be tested together.
+
+**Global Candidate List** - The list of all Component digest that have passed individual component testing. This can be retrieved from the Application to see what Components it is made of and then querying each of the Components `spec.containerImage`. 
+
+**Snapshot** - The list that maps all Components of an Application to their Component Image digests. Once created, the snapshot is immutable.
```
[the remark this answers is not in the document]
dirgim: Would `Once created, the list of Components with their images is immutable.` work?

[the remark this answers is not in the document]
arewm: I just get hung up where we state that it is generally immutable here and then later mutate part of it that we never mention until that point.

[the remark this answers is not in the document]
dirgim: Hm, the fact that the `Snapshot` is an k8s custom resource may be missing here to explain the fact that it can have a status field even if the contents of its spec are immutable. 

Something like this could clear that up:
```
**Snapshot** - The custom resource that contains the list of all Components of an Application with their Component Image digests. Once created, the list of Components with their images is immutable. The Integration service updates the status of the resource to reflect the testing outcome.
```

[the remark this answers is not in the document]
arewm: That sounds clear to me even if you are not familiar with custom resources having a status!

book/integration-service.md
srivickynesh: Done

book/integration-service.md
srivickynesh: Modified overview section by adding additional sections.

book/integration-service.md
srivickynesh: removed and added this under the overview section.

book/integration-service.md
srivickynesh: Done

book/integration-service.md
srivickynesh: done

book/integration-service.md
```diff
@@ -0,0 +1,264 @@
...
+| Snapshot | Post Test PipelineRun | To update the Snapshot's status to validated, so that it can be promoted to the next environment |
+| Component | Post Test PipelineRun | To update the Component with the image spec, in order to update the global candidates list |
+
+### WATCH
+
+| Custom Resource | When? | Why? |
+|---|---|---|
+| PipelineRun | Post Build PipelineRun | Read the component and the output image to create the Snapshot and start the Preliminary PipelineRun |
+| PipelineRun | Post Preliminary PipelineRun | Check the result of the Preliminary PipelineRun and  follow up on the global candidate list state to either trigger the Final PipelineRun or check ReleaseLink and create a Release |
+| PipelineRun | Post Final PipelineRun | Check the result of the Final PipelineRun, check ReleaseLink and create a Release |
```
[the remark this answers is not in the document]
srivickynesh: Yes. I will correct it. Thanks for pointing it out.

book/integration-service.md
srivickynesh: The updated overview should have Tekton mentioned in it

book/integration-service.md
dirgim: Yes, in the current design the Integration service will de-provision the ephemeral test environment after the testing is done.
We've been looking into providing an option to the users of leaving the environment running for a set amount of time, but that will require a new feature for gracefully handling the cleanup and related actions, both in functional and billing sense.

book/integration-service.md
dirgim: I don't think there is a set terminology on this yet, but what this basically means is that the Integration service will deploy to all Environments that are defined in the user's workspace that are:
- Not ephemeral
- Do not have a `parentEnvironment` defined

In most cases this will apply to development or staging environments.

book/integration-service.md
dirgim: The entire `open questions` section is actually an enhancement proposal and hasn't been implemented as such.
@srivickynesh since a lot of use cases that this proposal addresses has been covered with the other parts of the design already (such as IntegrationTestScenario having contexts and marking them as optional/required), my proposal is to drop this section.

book/integration-service.md
```diff
@@ -0,0 +1,264 @@
...
+
+**Global Candidate List** - The list of all Component digest that have passed individual component testing. This can be retrieved from the Application to see what Components it is made of and then querying each of the Components `spec.containerImage`. 
+
+**Snapshot** - The list that maps all Components of an Application to their Component Image digests. Once created, the snapshot is immutable. 
+
+**IntegrationTestScenario** - The custom resource that describes separate PipelineRuns that are supposed to be run by the snapshot controller. Any number of IntegrationTestScenarios can be set by the user and each of them includes information about the Tekton pipeline that will be executed - Tekton bundle, pipeline name and parameters, and the Environment details. The Integration Test Scenarios can also specify the contexts where they will be applied - this can be used to define when the scenario will run (Component or Composite stage - or both). If more than one Integration Test Scenario has the same context, their respective PipelineRuns will be executed in parallel.
```
dirgim, line 58: We can also add this sentence at the end:
```
The IntegrationTestScenarios can be marked as optional, in which case they will not be taken into account when determining if the Snapshot has passed testing.
```

book/integration-service.md
srivickynesh: Done

book/integration-service.md
```diff
@@ -0,0 +1,264 @@
...
+
+**Ephemeral Environment** - The ephemeral copy of an existing environment that’s created by the Integration service for use during a specific Integration pipeline run. After the environment is copied and marked as ephemeral, the Integration service creates a new SnapshotEnvironmentBinding linking the Snapshot under test with the newly created Environment. This triggers a deployment of the Snapshot on the new Environment, handled by the GitOps service. Finally, the Integration service links the credentials for the Environment to the Integration Tekton pipeline run and starts the test run.
+
+## Outdated Terminology
+
+**Prelim Pipeline** - This is the same as the Component pipeline. Going forward the Prelim Pipeline will be referred as Component Pipeline.
```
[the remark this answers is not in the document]
srivickynesh: Done

book/integration-service.md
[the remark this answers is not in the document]
srivickynesh: Done

book/integration-service.md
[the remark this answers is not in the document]
srivickynesh: Done

book/integration-service.md
srivickynesh: Done

book/integration-service.md
srivickynesh: Done

book/integration-service.md
srivickynesh: Done

book/integration-service.md
srivickynesh: Done

book/integration-service.md
dirgim: After giving this some thought, I think we can replace "we" with "The Integration service" in this whole section. Most - if not all - items refer to multiple personas who are using the Integration service, but it's the service that provides these functionalities in the end.

srivickynesh: Done

book/integration-service.md
srivickynesh: Done

dirgim: Thanks, looks good

book/integration-service.md
dirgim: Thanks @srivickynesh 
@amfred we'll certainly keep in mind to use blocklists/allowlists, thanks for pointing that out.

book/integration-service.md
```diff
@@ -0,0 +1,207 @@
...
+
+**Global Candidate List** - The list of all Component digest that have passed individual component testing. This can be retrieved from the Application to see what Components it is made of and then querying each of the Components `spec.containerImage`. 
+
+**Snapshot** -  The custom resource that contains the list of all Components of an Application with their Component Image digests. Once created, the list of Components with their images is immutable. The Integration service updates the status of the resource to reflect the testing outcome.
+
+**IntegrationTestScenario** - The custom resource that describes separate PipelineRuns that are supposed to be run by the snapshot controller. The IntegrationTestScenarios can be marked as optional, in which case they will not be taken into account when determining if the Snapshot has passed testing.
+
+**Ephemeral Environment** - The ephemeral copy of an existing environment that’s created by the Integration service for use during a specific Integration pipeline run.
```
arewm, line 60:
```suggestion
**Ephemeral Environment** - The ephemeral copy of an existing environment that is created by the Integration service for use during a specific Integration pipeline run.

```

book/integration-service.md
```diff
@@ -0,0 +1,207 @@
...
+
+The Integration Service is composed of controllers that facilitate automated testing of content produced by the build pipelines. It is mainly responsible for integration testing capabilities.
+
+The Integration service uses the pipeline, snapshot, and environment controllers to watch for Component builds, trigger and manage testing Tekton Pipelines, and create releases based on the testing outcome(s).
+
+### Goals
+
+- The Integration service need to be able to deploy a specific set of images to an environment and record the results of integration testing.
+- Given a specific set of images, Integration service should be able to tell if they passed application validation tests.
+- The Integration service should be able to update one component image at a time in order to support 'slow loop' application validation testing.
```
arewm, line 14: You don't explain what slow loop is. I know we removed descriptions of this earlier, but I think we need some explanation if you are going to use the terminology.

book/integration-service.md
```diff
@@ -0,0 +1,207 @@
...
+- The Integration service should have a capability to automatically promote sets of passed images.
+- To be SLSA compliant, Integration service should also be able to tell what version of the build and test pipelines were used to create each component image.
```
arewm, line 16: You are using SLSA without defining it. Can you at least link to some reference on SLSA compliance?

book/integration-service.md
arewm: Would it be reasonable to call this out somewhere as a reminder to come back to it ... or a notice to readers about how it is not resolved yet? It might provide some confusions when reading through as is.

book/integration-service.md
```diff
@@ -0,0 +1,207 @@
...
+
+**Component Pipeline** - This is the test pipeline run for the Component of an Application that gets triggered by a completed Build pipeline.
+
+**Composite Pipeline** - This is the pipeline run that runs if two or more Components have completed and passed their Component pipeline concurrently so that they should be tested together.
+
+**Global Candidate List** - The list of all Component digest that have passed individual component testing. This can be retrieved from the Application to see what Components it is made of and then querying each of the Components `spec.containerImage`. 
+
+**Snapshot** -  The custom resource that contains the list of all Components of an Application with their Component Image digests. Once created, the list of Components with their images is immutable. The Integration service updates the status of the resource to reflect the testing outcome.
+
+**IntegrationTestScenario** - The custom resource that describes separate PipelineRuns that are supposed to be run by the snapshot controller. The IntegrationTestScenarios can be marked as optional, in which case they will not be taken into account when determining if the Snapshot has passed testing.
```
arewm, line 58: Can you add something to the effect that this is the way users can interact with the integration service?

book/integration-service.md
srivickynesh: Since these Open Questions are already implemented/covered in the above definitions.

book/integration-service.md
dirgim: We can change this to `The Integration service should be able to update one component image at a time in order to test and deploy individual component builds.`
We will later update this with a bit more details or references as part of [HACBS-1426](https://issues.redhat.com/browse/HACBS-1426).

book/integration-service.md
srivickynesh: could you reiterate this question?

book/integration-service.md
srivickynesh: Done

book/integration-service.md
srivickynesh: Done

book/integration-service.md
srivickynesh: Done

book/integration-service.md
srivickynesh: Done

book/integration-service.md
srivickynesh: Done

book/integration-service.md
srivickynesh: This section is no longer needed. Done

book/integration-service.md
arewm: You have a parent heading with child headings/sections. You do not explicitly describe the relationship between the child headings and a `PipelineRun`. I was trying to suggest ways to tie all the sections together.  If you can have a short description about how the children are related to the parent, that would be great.

For example, are these properties of the PipelineRun that is specified? Are these properties that the integration service controls/required/sets? Are they pre-requisites that the user needs to configure or customize? 

Reading it again, you might be doing a good enough job here for now. Just take another look at it to see if any additional clarification is needed. If you don't think anything else is needed, feel free to resolve this thread.

srivickynesh: Updated. Done

arewm (approved):
Thanks for all the hard work on this summary!

book/integration-service.md
dirgim: As a quick update, I just got confirmation that the deprovisioning will be fully implemented as part of the [STONE-114](https://issues.redhat.com/browse/STONE-114) feature, so I propose we add the deprovisioning step information to this document as part of that effort.

book/integration-service.md
ralphbean:
```suggestion
- When all of the test Tekton PipelineRuns have passed, Integration service triggers a deployment of the application to all Environments which are not labeled as "ephemeral" and which do not have a `parentEnvironment` set. For most users, this will result in a deployment to their "development" environment (or to their "staging" environment, if they do not have a dev environment.
```

srivickynesh: Updated

dirgim (approved):
+1

PR #39, opened 2023-01-04
Update ADR title format to be consistent

ralphbean:
I aligned them here with the output of the `adr new` command, so it should be easier to keep them more consistent in the future.


ralphbean:
Should be visible rendered on my fork at https://ralphbean.github.io/book/ADR/index.html, compare with https://redhat-appstudio.github.io/book/ADR/index.html

amfred (approved):
Much better.

gorkem (approved):

PR #44, opened 2023-01-13
add personal data diagram

amfred:
How about we add this in addition to #4 and call it a day?


ralphbean (approved):

gorkem (approved):
We should also close #4

PR #47, opened 2023-01-31
Document decision to let pipelines proceed

ralphbean:


ADR/0014-let-pipelines-proceed.md
```diff
@@ -0,0 +1,53 @@
...
+All scanning and linting TaskRuns should *succeed* even if they find problems in the content they
+are evaluating.
+
+Use the `HACBS_TEST_OUTPUT` result convention from [ADR-0013] to expose those results and render them
+for users ([STONE-459]).
+
+## Consequences
+
+* Users should find that even if their scanners find problems, they can still build, test, and
+  deploy to lower Environments.
```
arewm, line 41: What is considered a `lower Environment` here?

ADR/0014-let-pipelines-proceed.md
```diff
@@ -0,0 +1,53 @@
...
+stopping them from running integration tests, and stopping them from deploying candidate builds to
+their lower [Environments]?
+
+## Decision
+
+All scanning and linting TaskRuns should *succeed* even if they find problems in the content they
+are evaluating.
+
+Use the `HACBS_TEST_OUTPUT` result convention from [ADR-0013] to expose those results and render them
+for users ([STONE-459]).
```
arewm, line 36: Would it be up to both the UI and PaC to render the output to users? I would expect that at least some GitHub users/organizations would want to enforce checks to pass before merging. How might this be able to be accomplished?

Instead of just having the UI/PaC go through all TaskRuns, would it be reasonable to have a "finalizer" Task which can aggregate and summarize the result of the previous scanning/linting TaskRuns. 

I don't _really_ like stating a further reliance on `HACBS_TEST_OUTPUT` as I think that we should try to come to some other generic pattern which doesn't propagate a tie-down to an old product name.

ADR/0014-let-pipelines-proceed.md
```diff
@@ -0,0 +1,53 @@
...
+A problem we face: in Tekton, a failing TaskRun causes the whole PipelineRun to fail. If the purpose
+of our linting and scanning tasks is to find problems - which usually looks like failure with
+a non-zero exit code - how do they do their job without constantly breaking the user's build,
+stopping them from running integration tests, and stopping them from deploying candidate builds to
+their lower [Environments]?
+
+## Decision
+
+All scanning and linting TaskRuns should *succeed* even if they find problems in the content they
+are evaluating.
```
arewm, line 33: Are there any types of Tasks which should explicitly _not_ always succeed?

arewm (commented):
I see the footnote about this being a verbal decision that has been made already. If this ADR should be approved as-is with that in mind, then I can just leave my comments as unresolved as needed, coming back around to them at some point in time later.

ADR/0014-let-pipelines-proceed.md
ralphbean: Left intentionally vague for the purposes of the ADR. Anything that is "not production". Here, we're still assuming that the user uses the managed-workspace release model via the release service.

ADR/0014-let-pipelines-proceed.md
ralphbean: > I don't really like stating a further reliance on HACBS_TEST_OUTPUT as I think that we should try to come to some other generic pattern which doesn't propagate a tie-down to an old product name.

Agreed. It is one of the reasons we've been holding off on moving forwards with [STONE-459](https://issues.redhat.com/browse/STONE-459), but it's getting too late now and I don't want to wait for [TKNECO-12](https://issues.redhat.com/browse/TKNECO-12). Once we get a recommendation there, we should land new decisions overturning this ADR and ADR 13.

> Would it be up to both the UI and PaC to render the output to users? I would expect that at least some GitHub users/organizations would want to enforce checks to pass before merging. How might this be able to be accomplished?

Yeah, that's the idea in STONE-459. PaC wouldn't really render the output into a UI element - it would render the output into a call to the GitHub Checks API, which would make a successful TaskRun with "errors" in its HACBS_TEST_OUTPUT end up being represented as a "failing" check in the GitHub UI.

> Instead of just having the UI/PaC go through all TaskRuns, would it be reasonable to have a "finalizer" Task which can aggregate and summarize the result of the previous scanning/linting TaskRuns.

This is possible, but PaC and the UI both already need to deal with per-task stuff anyways. PaC submits a check result for each task (in GitHub you can see a table of all ~dozen steps performed in a pipelinerun) and in the UI we render the whole pipeline run with the ability to click through into the logs of a task. Having both of those systems be able to render errors on a per-task level directs the user to exactly where they need to click to go see the stdout from scanner $foo.

ADR/0014-let-pipelines-proceed.md
ralphbean: Not clear to me at this time.

The git clone task should fail if it fails (because it can't provide output to the next stages of the pipelinerun). The buildah task should fail if it fails (because it can't produce the image... which is needed by subsequent stages of the pipelinerun?)

ralphbean:
> I see the footnote about this being a verbal decision that has been made already. If this ADR should be approved as-is with that in mind, then I can just leave my comments as unresolved as needed, coming back around to them at some point in time later.

I think the substance should be approved. I'm happy to integrate feedback on anything that's unclear or unrepresentative of the existing direction. If we disagree about the direction forwards, that's a good case to followup with an ADR that overrides this one - documenting a new point-in-time decision.

sbose78 (approved):

sbose78:
Seems fair to me.

hongweiliu17 (approved):
Just a few minor comment. LGTM.

ADR/0013-integration-service-api-contracts.md
```diff
@@ -6,6 +6,8 @@ Date: 2023-01-30
 
 Accepted
 
+Relates to [ADR 14. Let Pipelines Proceed](0014-let-pipelines-proceed.html)
```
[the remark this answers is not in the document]
ralphbean: Yes, it should get rendered as html at https://redhat-appstudio.github.io/book/ADR/ after merge.

arewm (approved):

ADR/0013-integration-service-api-contracts.md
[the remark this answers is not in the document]
hongweiliu17: I see now. Thank you for the link!

ADR/0014-let-pipelines-proceed.md
```diff
@@ -0,0 +1,53 @@
...
+* Without special treatment in [STONE-459], users may be misled or confused if their tasks appear to
+  succeed but really are reporting errors under the hood.
+
+## Footnotes
+
+* We originally made this decision verbally in May of 2022, and have been operating with it as an
+  unwritten principle. Documenting it here for posterity, visibility.
+
+[STONE-459]: https://issues.redhat.com/browse/STONE-459
+[Environments]: ../ref/application-environment-api.md#environment
```
hongweiliu17, line 51:
```suggestion
[Environments]: ../ref/application-environment-api.html#environment
```

PR #53, opened 2023-02-09
ADR 0015 Two phase architecture

dirgim:
* Add Integration service Two phase arch ADR


ADR/0015-integration-service-two-phase-architecture.md
```diff
@@ -0,0 +1,87 @@
...
+5. Composite phase testing runs in full and the Snapshot is promoted according to the
+user preferences
+
+## Footnotes
+For future consideration, if two components rely on each other to the point that
+breaking changes from one component create issues in another component during testing,
+a new feature to support batching of components together is being investigated.
+This would allow for the Integration service to hold off on testing a Component build
+until the dependent build is also completed.
+
```
ralphbean, line 87:
```suggestion

We originally made this decision verbally and with diagrams back in May of 2022, and have been operating with it as the operating design since then. However, we realized (through conversations like [this slack conversation](https://redhat-internal.slack.com/archives/C02CTEB3MMF/p1669961732100869)) that it is not obvious without documentation. We are documenting it here as an ADR for posterity, visibility. 
```

dirgim: Ack, added the proposed text below this line.

ADR/0015-integration-service-two-phase-architecture.md
```diff
@@ -0,0 +1,94 @@
...
+The Integration Service is in charge of running integration test pipelines by executing
+the Tekton pipeline for each user-defined IntegrationTestScenario.
+The main payload that’s being tested are Snapshots which contain references to all
+Components that belong to the Application, along with their images.
+
+A problem we face: Since testing an application with Tekton pipelines takes time,
+and multiple component builds can happen in quick succession, there is potential
+for race conditions between different Snapshots that are created for each of those builds.
+This would primarily manifest itself by two Snapshots getting promoted in quick
+succession, but neither of them having the latest images contained within them.
```
arewm, line 20:
```suggestion
One problem faced by the integration service is caused by the fact that testing an
application with Tekton pipelines takes time. During the course of testing an application, 
multiple component builds can happen in quick succession, leading to a potential race
condition between different Snapshots that are created for each of those builds.
This would primarily manifest itself by two Snapshots getting promoted in quick
succession but neither of them having the latest images contained within them.
```

ADR/0015-integration-service-two-phase-architecture.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Decision
+
+In order to tackle the race conditions, a two phase approach to integration testing
+will take place. The phases will consist of a Component phase that will always be executed
```
arewm, line 25:
```suggestion
In order to protect against the race conditions, the integration service will leverage a two phase
approach to testing. The phases will consist of a Component phase that will always be executed
```

ADR/0015-integration-service-two-phase-architecture.md
```diff
@@ -0,0 +1,94 @@
...
+by creating a Snapshot. All components from the Global Candidate List are included
+and the Component that was built is updated/overwritten with the newly built image.
+
+![](../diagrams/ADR-0015/component-phase.jpg)
+
+### Composite phase
+
+When multiple components are built at the same time, the Integration Service tests
+the application by creating a composite Snapshot using multiple components updated
+to use the newly built images.
```
arewm, line 41: Can you provide details for how this composite Snapshot is created/updated as more components are built?  Here are some questions/self-answers which I think would be good to address/correct.

It doesn't have to be too detailed, but maybe some form of text description about how the Global Candidate List is created/maintained/checked.

Based on later wording, I assume that the global composite list is updated after the first phase by updating a single component. If the Global Candidate List is identical to the just-passed snapshot, then we are all good. If not, the second loop is run. Is this Global Candidate List what is used for sending snapshot references to other services or would a new Snapshot (the composite snapshot?) be created based off of the (then) Global Candidate List (Is it any different from the other Snapshot other than it is built off of the Global Candidate List)? 

It looks like Snapshots will also only be completely tested/promoted if it matches the Global Candidate List, correct?

ADR/0015-integration-service-two-phase-architecture.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Status
+
+In review
+
+## Context
+
+The Integration Service is in charge of running integration test pipelines by executing
+the Tekton pipeline for each user-defined IntegrationTestScenario.
+The main payload that’s being tested are Snapshots which contain references to all
```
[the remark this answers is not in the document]
dirgim: Ack, updated with the suggestion.

ADR/0015-integration-service-two-phase-architecture.md
dirgim: Ack, updated with the suggestion.

ADR/0015-integration-service-two-phase-architecture.md
dirgim: Ack, updated with the suggestion.

ADR/0015-integration-service-two-phase-architecture.md
dirgim: I updated the first phase to clarify that the Global Candidate List is updated only after the testing successfully finishes and the Global Candidate List hasn't changed in the meantime. I also added
```
The Composite phase is used when the Global Candidate List changes while testing a
Snapshot in the Component phase. The Integration service checks this before promoting
a Snapshot at the end of the Component phase.
```

to the description of the Composite phase. Let me know if that clears it up.

arewm: I think that does clear many things up. So in these situations, the Snapshot which was out of sync with the Global Candidate List will just be "discarded" in favor of the new composite snapshot created from the Global Candidate List? 

I know that this isn't the realm of the integration service, but how would this be detected/displayed in the UI, or is there no promotion representation in the UI yet?

dirgim: Yes, that's correct, we mark the discarded Snapshot's status as invalid with the reason that it was superseded by a composite snapshot - this could be picked up by the UI in the future.

ADR/0015-integration-service-two-phase-architecture.md
```diff
@@ -0,0 +1,105 @@
...
+
+When a single component image is built, the Integration Service tests the application
+by creating a Snapshot. All Components with their images from the Global Candidate List
+are included within the Snapshot and then the Component that was newly built is
+updated/overwritten to complete the Snapshot creation.
+
+After all test pipelines for the Snapshot finish successfully, the Integration service
+updates the Global Candidate List with the newly built Component image and promotes
+the Snapshot. This is done only if the Global Candidate List hasn't changed in
+the meantime.
```
arewm, line 40: Maybe we can say that it is only promoted if the rest of the Global Candidate List matches the rest of the tested Snapshot components?

ADR/0015-integration-service-two-phase-architecture.md
arewm: Cool. I am good to resolve this thread after we indicate that a new composite snapshot is created based off of the global candidate list and that the previously tested snapshot is discarded by being marked with an invalid status.

ADR/0015-integration-service-two-phase-architecture.md
```diff
@@ -0,0 +1,109 @@
...
+a Snapshot at the end of the Component phase.
+
+The Composite phase exists to resolve a race condition when teams merge multiple PRs
+to multiple components of the same application at nearly the same time.
+When multiple components are built at the same time, the Integration Service tests
+the application by creating a composite Snapshot using multiple Components updated
+to use the newly built images.
+
+After the composite Snapshot passes testing, the entire Global Candidate List is updated
+according to its contents.
```
arewm, line 59: I think that we have to contend with another race condition here. I opened https://issues.redhat.com/browse/HACBS-1779 to discuss it.

arewm (changes requested):

ADR/0015-integration-service-two-phase-architecture.md
```diff
@@ -0,0 +1,109 @@
...
+### Component phase
+
+When a single component image is built, the Integration Service tests the application
+by creating a Snapshot. All Components with their images from the Global Candidate List
+are included within the Snapshot and then the Component that was newly built is
+updated/overwritten to complete the Snapshot creation.
+
+After all test pipelines for the Snapshot finish successfully, the Integration service
+updates the Global Candidate List with the newly built Component image and promotes
+the Snapshot. This is done only if the rest of the Global Candidate List matches
```
ralphbean, line 39: > This is done only if the rest of the Global Candidate List matches the rest of the tested Snapshot components.

I think this differs from how I thought it was supposed to work. I thought that at the end of the Component phase (assuming that it passes), the newly built image is _always_ promoted to the Global Candidate List (unless we're operating on a PR). That way, the Composite phase always starts operating on an ApplicationSnapshot that is always just "built from" the current state of the Global Candidate List - whatever that is.

One undesirable side-effect of the design as I understood it, was that "the build could get broken". Meaning, two components A and B could both pass the first Component testing phase, and both get added to the Global Candidates List. Component A could then pass the Composite testing phase, but Component B could _fail_ the Composite testing phase if it somehow interacts badly with Component A.

At this point, the Snapshot that includes the newly built Component A but _not_ the newly built Component B would be promoted to the lowest static Environment, but the Snapshot that includes both the newly built Component A and the newly built Component B would _not_ be promoted. After this point, all new builds would fail until the user removes the broken Component B from the global candidate list (we have no documented way how to do this) or until the user merges a fix for Component B.

Dannyb48: > One undesirable side-effect of the design as I understood it, was that "the build could get broken". 

In our early architecting days I thought it was OK for "dev" or lowest possible static environment to be broken. I actually forgot the analogue that was shared with us but it suggested hat dev AWLAYS got deployed to whether things were broken or not? If it was broken, developers submitted new commits that would fix the issue and continue rolling things forward.

dirgim: Ok, so the key takeaway here is that the Global Candidate List should always be updated with the newly built component when the Composite phase passes successfully?

ADR/0015-integration-service-two-phase-architecture.md
```diff
@@ -0,0 +1,109 @@
...
+1. Two instances of the Component phase start up in parallel, testing two different
+Snapshots - each one including one of the two merged PRs.
+2. The first Snapshot finishes testing and detects that the Global Candidate List
+has not changed in the meantime and finalizes the Snapshot in question,
+updating the Global Candidate List with the newly built component image.
+3. The second Snapshot finishes testing. It is presented with an updated
+Global Candidate List that includes both changes, which have not been tested together yet.
+4. A Composite Snapshot is created that includes both changes and the Integration service
+starts testing it in the Composite phase.
+5. Composite phase testing runs in full and the Snapshot is promoted according to the
```
scoheb, line 93:
```suggestion
5. Composite phase testing runs in full and the **Composite** Snapshot is promoted according to the
```

ADR/0015-integration-service-two-phase-architecture.md
dirgim: Apologies, corrected my statement, what I meant to say is: 
```
The Global Candidate List should always be updated with the newly built component when the _Component_ phase passes successfully
```

arewm: ACK. I will retract my previous comment in light of that.

If the Global Candidate List is always updated on a completed (success/fail) Component snapshot, then this will not necessarily result in broken dev environments as the snapshot itself might indicate failure (and therefore block promotion). In this case, we would just be blocking promotion to the lowest environment until we have a passing Snapshot. I don't have an issue with that interface as long as we are suitably notifying users of the failure (I assume it would be a PaC notification back to GH on the commit).

dirgim: Ok, in that case we can remove this sentence so the design matches the expectations:

> This is done only if the rest of the Global Candidate List matches the rest of the tested Snapshot components.

We will make the update to our controller logic as part of [HACBS-1779](https://issues.redhat.com/browse/HACBS-1779) so the successful Component pipeline always updates the Global Candidate list.
@ralphbean @arewm @Dannyb48

ADR/0015-integration-service-two-phase-architecture.md
dirgim: Ack, updated

ADR/0015-integration-service-two-phase-architecture.md
dirgim: @ralphbean based on your comments, should we drop this sentence entirely and modify the Integration service so it doesn't update the Global Candidate List at the end of the Composite phase?

ralphbean: Yes, I think so.

ralphbean: It is replaced with the sentence above during the _component_ phase:

> After all test pipelines for the Snapshot finish successfully, the Integration service updates the Global Candidate List with the newly built Component image

ADR/0015-integration-service-two-phase-architecture.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Happy path with a single component build
+
+In the happy path without race conditions, one PR merges to one component.
+1. The Component phase runs, and records results on the Snapshot.
+2. The Integration Service detects that the Snapshot is up-to-date by checking the
+Global Candidate List and skips the second phase of testing, marking the Snapshot as
+complete.
+3. The Snapshot is promoted according to the user preferences and the
+Global Candidate List is updated.
```
ralphbean, line 77: Does this need rephrasing as a result of our discussions earlier in the ADR here?

ralphbean: No, it's probably fine as is. The next "Path with a race condition" section reads correctly.

ADR/0015-integration-service-two-phase-architecture.md
dirgim: Ack, removed the sentence and only mentioned promoting the Composite snapshot according to user preferences

ADR/0015-integration-service-two-phase-architecture.md
dirgim: I updated the wording and made the matching changes to the diagrams, please take a look

ADR/0015-integration-service-two-phase-architecture.md
```diff
@@ -0,0 +1,109 @@
...
+invalid and the testing goes into the Composite phase.
+Otherwise, the Component Snapshot is promoted according to user preferences.
+
+![](../diagrams/ADR-0015/component-phase.jpg)
+
+### Composite phase
+
+The Composite phase is used when the Global Candidate List changes while testing a
+Snapshot in the Component phase. The Integration service checks this before promoting
+a Snapshot at the end of the Component phase.
```
arewm, line 50: I don't think this last sentence is needed. Otherwise, I think it reads well!

ADR/0015-integration-service-two-phase-architecture.md
arewm: I think that this can be updated because the Global Candidate List is updated and then the Snapshot is promoted.

ADR/0015-integration-service-two-phase-architecture.md
dirgim: Yep, I can see it now too, removed it.

ADR/0015-integration-service-two-phase-architecture.md
dirgim: Ack, updated this to split the points

ADR/0015-integration-service-two-phase-architecture.md
```diff
@@ -0,0 +1,108 @@
...
+in quick succession.
+
+### Happy path with a single component build
+
+In the happy path without race conditions, one PR merges to one component.
+1. The Component phase runs, and records results on the **Component** Snapshot.
+2. The Integration Service detects that the Snapshot is up-to-date by checking the
+Global Candidate List and skips the second phase of testing, marking the Snapshot as
+complete.
+3. The Global Candidate List is updated.
```
arewm, line 76: According to above, the list is updated before marking as complete.

dirgim: Good catch, switched the order

ralphbean (approved):

arewm (changes requested):
The component-phase.jpg needs to be updated with the proper flow of always updating the global candidate list.

arewm (approved):
The text for stage 4 indicates that it promotes the component to the Global Candidate List after passing all tests instead (instead of always promoting). I must have mixed some conversations as this is consistent with the text.

PR #57, opened 2023-02-10
ADR: Use our pipelines

ralphbean:
From slack today: https://redhat-internal.slack.com/archives/C02FANRBZQD/p1676023583493379

### DIFF
--- a/ADR/0017-use-our-pipelines.md
+++ b/ADR/0017-use-our-pipelines.md
@@ -22,7 +22,7 @@ a ci/cd platform that is meant to support a secure software development lifecycl
 Use our own pipelines to build and scan Stonesoup components. Almost all of our components already
 do this today. Look for evidence in the `.tekton/` directory of their git repo.
 
-However, we have stopped short of configuring and [Application] and [Components] for Stonesoup.
+However, we have stopped short of configuring an [Application] and [Components] for Stonesoup.
 We're using the pipelines directly, but not via the Stonesoup UI. This is something we intend to
 start doing, but haven't made time to do so yet.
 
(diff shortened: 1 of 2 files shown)


arewm (approved):

ralphbean:
Related thread: https://redhat-internal.slack.com/archives/C02FTKEMBU1/p1677592606558619?thread_ts=1677590672.334459&cid=C02FTKEMBU1

ralphbean:
TODO - incorporate a link to https://issues.redhat.com/browse/STONE-434

rhopp (approved):

dirgim (approved):
LGTM

ADR/0017-use-our-pipelines.md
```diff
@@ -0,0 +1,44 @@
...
+## Decision
+
+Use our own pipelines to build and scan Stonesoup components. Almost all of our components already
+do this today. Look for evidence in the `.tekton/` directory of their git repo.
+
+However, we have stopped short of configuring and [Application] and [Components] for Stonesoup.
+We're using the pipelines directly, but not via the Stonesoup UI. This is something we intend to
+start doing, but haven't made time to do so yet.
+
+## Consequences
```
adambkaplan, line 29: What of teams that are packaging/deploying Red Hat products built on CPaaS? Would this be considered out of scope for now?

adambkaplan: Follow up question - is there an means that we can get similar data for our products?

arewm: I think that the only parts that would be built on CPaaS are those which are currently being productized. As an example, I assume that we are using the official gitops and pipelines products.

Is that what you were asking? If so, _those_ would be out of scope, but any integration/configuration in Stonesoup would be in scope.

adambkaplan: Correct - _however_ I had drafted an ADR a while back related to Pipelines where we want to deploy a ".next" version. I need to update that since the Pipelines team now thinks we have a way to make this work.

PR #60, opened 2023-02-15
Remove workspace_id from namespace metadata

hmariset:
There is no workspace ID in stonesoup and for now
the workspace name is unique identifier.

Jira-Url: https://issues.redhat.com/browse/STONEO11Y-8

### DIFF
--- a/ADR/adr-0010-namespace-metadata
+++ b/ADR/adr-0010-namespace-metadata
@@ -32,8 +32,7 @@ We will apply the following labels to Stonesoup namespaces, to make them easier 
 
 The following labels are used for billing and telemetry. Values can be left blank if they are not defined yet:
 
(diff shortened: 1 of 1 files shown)


hmariset:
@alexeykazakov @ralphbean @amfred Please review. @amisstea and I agreed upon removing the workspace_id since there is no ID  in stonesoup and workspace name is unique.

amisstea (approved):

ralphbean (approved):
ADRs, in general, are point-in-time decisions. So, we try not to revise them and instead try to make and record new decisions that overrule the old ones.

However, this is a small change though to a _very new_ ADR. No harm in updating it in my opinion.

alexeykazakov (approved):

PR #61, opened 2023-02-18
ADR-017 Posture Keeper: Service for SARIF data storage and retrieval

gorkem:
ADR for the service for storing SARIF based SAST analysis.

Related to STONE-458

### DIFF
--- a/ADR/0017-posture-keeper.md
+++ b/ADR/0017-posture-keeper.md
@@ -1,0 +1,59 @@
+# 17. Posture Keeper
+
+Date: 2023-02-18
+
+## Status
+
+Accepted
+
+## Context
+Posture keeper is a REST-based web service that allows clients to store and retrieve SAST analysis data for a specific git repository and commit SHA. The service is designed to be scalable, reliable, and secure, and uses MongoDB as its underlying data store. The service is implemented using Quarkus. By utilizing this service StoneSoup is going to be able to provides experiences that reflects the current state of security and/or quality posture of the applications.
+
+## Design Goals
+
+The following are the key design goals of the service:
+
+* Support storing and retrieving SAST analysis data for a specific git repository and commit SHA
+* Accept and produce JSON-formatted data
+* Use MongoDB as the underlying data store for efficient data retrieval and storage
+* Support secure authentication and authorization to ensure only authorized clients can upload and retrieve data
+* Implement rate limiting for the POST endpoint to prevent abuse and ensure optimal performance
+* Implement pagination for the GET endpoint to efficiently retrieve large volumes of data
+
+## Architecture
+
+The Posture Keeper is implemented as a microservice using Quarkus. The architecture consists of the following components:
+
+### Client
+
+The client is responsible for making requests to the service endpoints using HTTP. Clients can be web applications, mobile applications, or any other type of application that can communicate over HTTP.
+
+### REST API
+
+The REST API is implemented using Quarkus. The API provides endpoints for storing and retrieving SAST analysis data for a specific git repository and commit SHA. It also provides an endpoint for comparing two analysis results and returning the new defects.
+
+The REST API is designed to accept and produce JSON-formatted data. It uses the Quarkus framework to handle HTTP requests and responses, and uses the MongoDB driver to interact with the MongoDB database.
+
+### MongoDB
+
+MongoDB is used as the underlying data store for the service. It stores the uploaded SARIF files as plain JSON data, along with metadata such as the repository URL, commit SHA, and tool name.
+
+The MongoDB database is designed to be scalable, reliable, and efficient. It uses sharding and replication to ensure data availability and reliability. It also uses indexes to ensure efficient data retrieval.
+
+### Authentication and Authorization
+
+The service uses a secure authentication and authorization mechanism to ensure that only authorized clients can upload and retrieve SAST analysis data.
+
+### Rate Limiting
+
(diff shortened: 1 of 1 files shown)


ADR/0017-posture-keeper.md
```diff
@@ -0,0 +1,59 @@
+# 17. Posture Keeper
+
+Date: 2023-02-18
+
+## Status
+
+Accepted
+
+## Context
+Posture keeper is a REST-based web service that allows clients to store and retrieve SAST analysis data for a specific git repository and commit SHA. The service is designed to be scalable, reliable, and secure, and uses MongoDB as its underlying data store. The service is implemented using Quarkus. By utilizing this service StoneSoup is going to be able to provides experiences that reflects the current state of security and/or quality posture of the applications.
```
ralphbean, line 10: I recommend writing more about the background here. What are the multiple problems that we face that drive us to writing the ADR in the first place. Years from now, when we look back and wonder why we created Posture Keeper -- what will we wish we knew about how we were thinking at the time. What are the constraints facing us? (pipeline service workspaces and pvcs get pruned, for example)

Keep mention of the decision (create posture keeper) until the decision section.

ADR/0017-posture-keeper.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Design Goals
+
+The following are the key design goals of the service:
+
+* Support storing and retrieving SAST analysis data for a specific git repository and commit SHA
+* Accept and produce JSON-formatted data
+* Use MongoDB as the underlying data store for efficient data retrieval and storage
+* Support secure authentication and authorization to ensure only authorized clients can upload and retrieve data
```
ralphbean, line 19: You can likely make a much more specific statement here. Should the service utilize subject access reviews in order to conform with the general RBAC system shared by other stonesoup services?

ADR/0017-posture-keeper.md
ralphbean, line 10: The scope is mixed here. On the one hand, it is specifically about SAST. On the other hand, the text suggests that it supports security and/or quality records more generally. Which is it?

ADR/0017-posture-keeper.md
```diff
@@ -0,0 +1,59 @@
...
+* Implement rate limiting for the POST endpoint to prevent abuse and ensure optimal performance
+* Implement pagination for the GET endpoint to efficiently retrieve large volumes of data
+
+## Architecture
+
+The Posture Keeper is implemented as a microservice using Quarkus. The architecture consists of the following components:
+
+### Client
+
+The client is responsible for making requests to the service endpoints using HTTP. Clients can be web applications, mobile applications, or any other type of application that can communicate over HTTP.
```
ralphbean, line 29: In a deployment of stonesoup, should client requests from web applications (like HAC) navigate through a proxy on the host cluster?

ADR/0017-posture-keeper.md
```diff
@@ -0,0 +1,59 @@
...
+
+The service implements rate limiting for the POST endpoint to prevent abuse and ensure optimal performance. The rate limiting can be implemented using Quarkus' built-in support for rate limiting.
+
+### Pagination
+
+The service implements pagination for the GET endpoint to efficiently retrieve large volumes of data. The pagination can be implemented using the MongoDB driver.
+
+## Consequences
+
+* With the introduction of the service the pipelines that do SAST analysis should upload thier results to the service
```
ralphbean, line 57: I recommend pulling this consequence into the architecture above. There is a decision to be made over whether a taskrun or finalizer should be implemented in pipelineruns that uploads to posture keeper or if a controller should be built which watches for pipelineruns matching some criteria which uploads results on their behalf.

ADR/0017-posture-keeper.md
arewm: Security and quality can still be derived from static scans. Are you trying to ask if we are having an `and` or an `or` and asking to not have both?

I imagine that the language is in there because it depends on the scans that were run against a snapshot at whatever point in time -- whether they were primarily security or quality related.

arewm, line 10: Can we define SAST the first time it is used?

ADR/0017-posture-keeper.md
arewm: Is the intention to have this endpoint available from Outside Stonesoup as well? Or is Posture Keeper only intended to serve internal requests/workloads?

ADR/0017-posture-keeper.md
```diff
@@ -0,0 +1,59 @@
...
+
+The Posture Keeper is implemented as a microservice using Quarkus. The architecture consists of the following components:
+
+### Client
+
+The client is responsible for making requests to the service endpoints using HTTP. Clients can be web applications, mobile applications, or any other type of application that can communicate over HTTP.
+
+### REST API
+
+The REST API is implemented using Quarkus. The API provides endpoints for storing and retrieving SAST analysis data for a specific git repository and commit SHA. It also provides an endpoint for comparing two analysis results and returning the new defects.
```
arewm, line 33: What would be the basis of an analysis result? Would we be on a snapshot level or a component level?

ralphbean: FWIW, on reading this I had assumed it would be on a component level - due to the retrieval parameters being "git repository and commit SHA".

amfred:
What's the current status of this proposal? Is it still in the queue to be implemented, or no?

PR #62, opened 2023-02-27
Update status and approved date on promtion logic ADR

sonam1412:
Just to update the correct status and add the accepted date on the approved promotion logic ADR


sonam1412:
cc: @arewm @gbenhaim PTAL

gbenhaim (approved):

arewm (approved):

PR #63, opened 2023-03-06
Enable partners to contribute/maintain Tasks

sbose78:
Associated feature: STONE-549

Simple/inexpensive way to let partners contribute/maintain Tekton Tasks provided for stonesoup's consumption.

### DIFF
--- a/ADR/partner-tasks.md
+++ b/ADR/partner-tasks.md
@@ -31,7 +31,6 @@ https://issues.redhat.com/browse/STONE-549
 2. CI tests do the due diligence on the changes proposed in the PR. Success/Failures are reported in a way that the PR author can take reasonable 
 action to resolve the issue.
 3. Upon approval from the Build/Test team, the changes are merged.
-4. Manually/Automatically, a PR is opened to the https://github.com/redhat-appstudio/build-definitions repository.
 
 #### Revoking a Task
 1. Open a PR to delete the Task.
@@ -64,7 +63,7 @@ Component into StoneSoup, customize their Pipeline definition in the .tekton dir
 
 ## Alternatives
 
-* Use the github.com/redhat-appstudio/build-definitions for Task submissions by partners : This is being considered in the short-term, either way, the day-to-day operations will not quite change.
+* ~Use the github.com/redhat-appstudio/build-definitions for Task submissions by partners : This is being considered in the short-term, either way, the day-to-day operations will not quite change~ - this has been promoted to be the primary design.
 * Use Tekton Hub for host Tasks : Tekton Hub is being deprecated.
 
 
--- a/ADR/partner-tasks.md
+++ b/ADR/partner-tasks.md
@@ -1,0 +1,34 @@
+# Partner Tasks in Build/Test Pipelines
+
+Date: 2023-03-06
+
+## Status
+
+Provisional
+
+## Context
+
+* As a Red Hat Partner, I would like to offer our service's capability as a Tekton Task that would be executed in a user's build/test Pipeline on StoneSoup.
+* As a StoneSoup maintainer, I would like to provide a way to ensure the above Tekton Task is validated and approved for use in the build/test Pipeline on StoneSoup.
+
+Associated business ask
+https://issues.redhat.com/browse/STONE-549
+
+## Decision
+
+### Plumbing
+
+1. Setup a new Git repository.
+2. Define a directory structure for Task submissions as manifests in the yaml format.
+3. Configure a CI job that validates the Tasks upon opening of a pull request.
+4. Configure a CI job that generates an OCI artifact consumable in a Tekton Pipeline.
+
+### Day-to-day operations
+1. Partner opens a PR with a new/updated Task.
+2. Upon approval from the Build/Test team, the changes are merged.
+3. Manually/Automatically, a PR is opened to the https://github.com/redhat-appstudio/build-definitions repository
+
+## Consequences
+
+* We have a mechanism to take in Tekton Tasks from Partners with due diligence.
+* We do build up a little tech-debt because this process needs to be merged with the official Red Hat certification process in the future.
--- a/ADR/partner-tasks.md
+++ b/ADR/partner-tasks.md
@@ -18,17 +18,61 @@ https://issues.redhat.com/browse/STONE-549
 
 ### Plumbing
 
-1. Setup a new Git repository.
+1. Setup a new Git repository for hosting all partner Tasks.
 2. Define a directory structure for Task submissions as manifests in the yaml format.
 3. Configure a CI job that validates the Tasks upon opening of a pull request.
 4. Configure a CI job that generates an OCI artifact consumable in a Tekton Pipeline.
 
 ### Day-to-day operations
+
+#### Adding/Updating a Task
+
 1. Partner opens a PR with a new/updated Task.
-2. Upon approval from the Build/Test team, the changes are merged.
-3. Manually/Automatically, a PR is opened to the https://github.com/redhat-appstudio/build-definitions repository
+2. CI tests do the due diligence on the changes proposed in the PR. Success/Failures are reported in a way that the PR author can take reasonable 
+action to resolve the issue.
+3. Upon approval from the Build/Test team, the changes are merged.
+4. Manually/Automatically, a PR is opened to the https://github.com/redhat-appstudio/build-definitions repository.
+
+#### Revoking a Task
+1. Open a PR to delete the Task.
+2. The Build/Test team reviews the PR and merges it.
+3. The Build/Test team updates the https://github.com/redhat-appstudio/build-definitions to remove references to the Task's OCI image whenever it is reasonable to do so.
+
+#### Definition of a valid Task
+
+The due diligence is a transparent-to-the-Task-contributor CI job that's running on the repository that validates the Task before we merge it in. 
+
+Please see the following as prior art:
+1. See CI Results in https://github.com/tektoncd/catalog/
+2. https://github.com/k8s-operatorhub/community-operators/
+2. https://github.com/openshift-helm-charts/charts
+
+A non-exhaustive list of checks that would be run on a Task is:
+
+* Linting
+* Scanning
+* No privilege escalation / or no requirements that need unreasonable privileges.
+* Should be integratabtle into the build pipeline, ie, works with the inputs/outputs we have today.
+* Should work with reasonable defaults.
+* Should be skip-able if credentials/tokens are missing.
+* *TBD*
+
+## Out-of-scope
+
+* Supporting validation of Tasks inside Stonesoup before submission would be out-of-scope. However, partners should be able to import a 
+Component into StoneSoup, customize their Pipeline definition in the .tekton directory and have the changes validated in a PipelineRun execution in StoneSoup. To be able to be productive with this flow, they'd need to be able to do https://github.com/redhat-appstudio/book/pull/64 .
+
+## Alternatives
+
+* Use the github.com/redhat-appstudio/build-definitions for Task submissions by partners : This is being considered in the short-term, either way, the day-to-day operations will not quite change.
+* Use Tekton Hub for host Tasks : Tekton Hub is being deprecated.
+
 
 ## Consequences
 
 * We have a mechanism to take in Tekton Tasks from Partners with due diligence.
+* We should also be able to take in Tekton Tasks from Red Hat teams with the same level of validation/diligence that Red Hat Partners would have to go through.
 * We do build up a little tech-debt because this process needs to be merged with the official Red Hat certification process in the future.
+
+
+
--- a/ADR/partner-tasks.md
+++ b/ADR/partner-tasks.md
@@ -18,10 +18,10 @@ https://issues.redhat.com/browse/STONE-549
 
 ### Plumbing
 
-1. Setup a new Git repository for hosting all partner Tasks.
+1. Setup a new Git high-level directory in https://github.com/redhat-appstudio/build-definitions for partners to contribute Tasks to.
 2. Define a directory structure for Task submissions as manifests in the yaml format.
 3. Configure a CI job that validates the Tasks upon opening of a pull request.
-4. Configure a CI job that generates an OCI artifact consumable in a Tekton Pipeline.
+4. Optionally, configure a CI job that generates an OCI artifact consumable in a Tekton Pipeline.
 
 ### Day-to-day operations
 
--- a/ADR/partner-tasks.md
+++ b/ADR/partner-tasks.md
@@ -4,7 +4,7 @@ Date: 2023-03-06
 
 ## Status
 
-Provisional
+Accepted
 
 ## Context
 


ADR/partner-tasks.md
```diff
@@ -0,0 +1,34 @@
...
+4. Configure a CI job that generates an OCI artifact consumable in a Tekton Pipeline.
+
+### Day-to-day operations
+1. Partner opens a PR with a new/updated Task.
+2. Upon approval from the Build/Test team, the changes are merged.
+3. Manually/Automatically, a PR is opened to the https://github.com/redhat-appstudio/build-definitions repository
+
+## Consequences
+
+* We have a mechanism to take in Tekton Tasks from Partners with due diligence.
```
arewm, line 33: What would certification of these tasks look like? For inclusion in Red Hat's pipleine? For inclusion in others' pipelines? Can we provide any clarity around "due diligence"?

The proposed SLSA v1.0 is largely just a transference of trust to the build system. A "similar" build system basically has trusted build tasks
> The way we defined the core elements of a "secure" build service in the Secure Software Factory Ref Arch is pretty much as follows:
> * Secure orchestration of the build pipeline. i.e. only approved pipelines that follow established policy and rules should be allowed to run.
> * Secure orchestration of the build workload. i.e. only approved build workloads are allowed to be run, and if someone tampers with the orchestration of a workload it should be detected or prevented. An example here is using SPIFFE or other workload identities.
> * Secure the actions in the build workload itself. i.e. a build that tries to perform malicious actions should be prevented and the build workload should be isolated enough to limit blast radius.

ADR/partner-tasks.md
```diff
@@ -0,0 +1,34 @@
...
+* As a StoneSoup maintainer, I would like to provide a way to ensure the above Tekton Task is validated and approved for use in the build/test Pipeline on StoneSoup.
+
+Associated business ask
+https://issues.redhat.com/browse/STONE-549
+
+## Decision
+
+### Plumbing
+
+1. Setup a new Git repository.
```
ralphbean, line 21: One repo for all partner tasks from all partners?

Or, one repo per partner per task?

Or, one repo per partner?

sbose78: One repo for all partners to begin with.

sbose78: I left it for implementation - but it should be in a new Github organization.

ADR/partner-tasks.md
sbose78: The exhaustive list of "due diligence" is being built out by the @vdemeester 's team ( tekton ecosystem / certification ). 

I doubt we'd need to have that in our initial iteration, therefore our stripped down list, which we should iterate on:

* No privilege escalation / or no requirements that need unreasonable privileges.
* Should be integratabtle into the build pipeline, ie, works with the inputs/outputs we have today.
* Should work with reasonable defaults.
* Should be skip-able if credentials/tokens are missing.

arewm: I would expect some level of "due diligence" to be provided by Red hat as well -- in addition to tekton ecosystem certification. I know that HACBS always has an 80/20 rule where the out-of-the-box configuration was only intended to solve 80% of use cases. The flexibility of the platform allowed for the other 20% to get through.

Therefore, partner tasks might not just be external partners, but also potentially internal users. How can we (via `ec`) certify that specific tasks are "approved" for use in software that we use? How can external users certify that specific tasks meet their requirements as well?

Even if we just have a stripped down list of diligence, I think we should include _something_ on it in the ADR, also mentioning how `ec` might be able to filter. @simonbaird, fyi on the ec angle.

> Should be skip-able if credentials/tokens are missing.

Would the tasks be injected into the pipeline by default? If so, this will produce many extraneous "skipped" tasks in the UI, especially as the number of partner tasks increases.

ADR/partner-tasks.md
ralphbean: OK - does the TKNECO team own this new repo, or does the STONEBLD team own this new repo?

sbose78: STONEBLD team would own this repo.

ADR/partner-tasks.md
sbose78: Yes, I do expect internal tasks to go here as well. 

Yes, the due diligence is by Red Hat - in the form of a CI job that validates contributions made to the repository.

ADR/partner-tasks.md
```diff
@@ -0,0 +1,34 @@
...
+## Status
+
+Provisional
+
+## Context
+
+* As a Red Hat Partner, I would like to offer our service's capability as a Tekton Task that would be executed in a user's build/test Pipeline on StoneSoup.
+* As a StoneSoup maintainer, I would like to provide a way to ensure the above Tekton Task is validated and approved for use in the build/test Pipeline on StoneSoup.
+
+Associated business ask
```
Michkov, line 14: ask -> task

ADR/partner-tasks.md
Michkov: Cannot we just reuse build-definitions and create there folder partner-tasks? Especially when the tasks should be moved to build-definitions anyway, this way we will have just one source.

ADR/partner-tasks.md
```diff
@@ -0,0 +1,34 @@
...
+## Decision
+
+### Plumbing
+
+1. Setup a new Git repository.
+2. Define a directory structure for Task submissions as manifests in the yaml format.
+3. Configure a CI job that validates the Tasks upon opening of a pull request.
+4. Configure a CI job that generates an OCI artifact consumable in a Tekton Pipeline.
+
+### Day-to-day operations
```
gbenhaim, line 26: How do users discover those pipeline tasks and enable them?

ADR/partner-tasks.md
gbenhaim, line 21: Is it possible to write why we can't use tekton-hub?
https://github.com/tektoncd/hub https://hub.tekton.dev/

burrsutter: Is this where our current Tasks come from today?  If so then I do think we should be consistent here.

ADR/partner-tasks.md
burrsutter: Let's type out exactly what we mean versus the ambiguous "due diligence".  For instance:
Partner is able to test their Task in their own workspace by using a custom pipeline and modifying the contents of their .tekton/ directory.
The Partner can maintain a blog that identifies how their Task should be integrated into a user's pipelinerun definition. Instructions should include how to add a secret/token to activate the task.

Questions:
Is it possible for the Partner to use  Stonesoup as their CI solution for their Task?

ADR/partner-tasks.md
arewm: It has always been my understanding that we never wanted to enable "arbitrary" leveraging of tasks from `tekton-hub` because it isn't _necessarily_ trusted task definition. There were some initial noodling of us being able to host our own form of `tekton-hub` containing _trusted_ tasks which have gone through certain vetting. This also enables us to have a buffer from the task HEADs so that we might have time to catch potential breaking changes before they affect the pipeline.

ADR/partner-tasks.md
arewm: Even if Red Hat is "okay" with these partner-supplied tasks, all customers/users might not be okay with them for their own policies without additional diligence done. I think that we need to make sure we are transparent about what diligence happens (initially and long-term as updates are pushed) and what policies might be able to be applied to filter out these tasks.

ADR/partner-tasks.md
```diff
@@ -0,0 +1,34 @@
...
+
+Date: 2023-03-06
+
+## Status
+
+Provisional
+
+## Context
+
+* As a Red Hat Partner, I would like to offer our service's capability as a Tekton Task that would be executed in a user's build/test Pipeline on StoneSoup.
```
arewm, line 11: After reviewing the ADR, I realize that I might be conflating "Partner" with "user" or "customer." Can we define what it means to be a partner? Is there any criteria that is met? Is there a process to become one? Is this just for certain tools that want to be leveraged in the pipeline (i.e. `random-sast-scanner`) or might this also be relevant if a user wants to contribute a task that they developed back to the platform?

ADR/partner-tasks.md
arewm: Moving a comment in a later thread here as it is more relevant here.

> Would the tasks be injected into the pipeline by default? If so, this will produce many extraneous "skipped" tasks in the UI, especially as the number of partner tasks increases.

ADR/partner-tasks.md
vdemeester: @gbenhaim mainly because the hub is deprecated. And no this is not where stonesoup tasks are going today.
Upstream, the "platform" of choice to get tasks indexed and searchable is artifacthub now, and artifacthub can index any number of catalog, including ours (severals if we want), if they follow the correct organizations (versionning, …)

vdemeester: @ralphbean in TKNECO, the current idea is that we'll have "on central repo" (with almost all task, from RH teams, from partners, …) that would be "generated"/populated from any number of smaller repositories — this would allow any teams to work with the tools they want, and "release" their task when they think it's ready, so that it gets "automatically pulled" in the "Red Hat Tekton Resource Catalog" 👼🏼

ADR/partner-tasks.md
vdemeester: Technically, this is independent of the "day-to-day" operations described here. But as written above, using the correct organization, it could be easily indexed by artifacthub or something else, and/or pull in our TKNECO catalog, …

ADR/partner-tasks.md
vdemeester: > Yes, the due diligence is by Red Hat - in the form of a CI job that validates contributions made to the repository.

Ideally, the Task that would get accepted and "certified" in the TKNECO realm, would pass a lot of gates ; best practices, linting, scaning, … and be heavily tested.. etc.. All this is has yet to be properly defined but that's the idea.

ADR/partner-tasks.md
ralphbean: What @Michkov suggests would be dead-simple for the short term, and avoid engineering some new repo with a new CI propagation pipeline when we know we want to move the partner content somewhere else in the medium/long-term.

sbose78: I'm generally OK with that, except that, we'd have to have non-red hatters be part of the same repository/org. More of a stylistic concern for me. Not a strong opinion.

ADR/partner-tasks.md
sbose78: @gbenhaim , after adding the Task to Red Hat's certification repo ( the short-term/long-term one ),  this steps get the Task into the stonesoup system:

> Manually/Automatically, a PR is opened to the https://github.com/redhat-appstudio/build-definitions repository.

Additionally, the user may have to configure tokens to be able to _not skip_ the Task.

ADR/partner-tasks.md
sbose78: Added a section called "Alternatives".

ADR/partner-tasks.md
sbose78: I actually mean't "ask" :D

ADR/partner-tasks.md
sbose78: Added to "Alternatives".

ADR/partner-tasks.md
sbose78: * See the section "Definition of a valid Task"
* @burrsutter , see the "out-of-scope" section. 

let me know!

ADR/partner-tasks.md
sbose78: "Upon approval from the Build/Test team, the changes are merged.
4. Manually/Automatically, a PR is opened to the https://github.com/redhat-appstudio/build-definitions repository."

ADR/partner-tasks.md
arewm: If users are made part of the organization, we would have to reassess our assumptions about security (i.e. teams being private is enough to hide them). Would it not be enough to have partners just submit forks? Do individuals need to be members of the organization in order to fully approve and merge PRs?

ADR/partner-tasks.md
```diff
@@ -0,0 +1,34 @@
...
+* As a StoneSoup maintainer, I would like to provide a way to ensure the above Tekton Task is validated and approved for use in the build/test Pipeline on StoneSoup.
```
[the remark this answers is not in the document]
burrsitis: Not at this time.   Any end-user may craft their own custom Task and share it with the world by making it a public image in an image registry and providing the yaml snippet needed to use it.   The real question is how does this Task need to produce its meta-data and attestations such that enterprise contract customizers can provide rules associated with it.

ADR/partner-tasks.md
sbose78: > an we define what it means to be a partner? Is there any criteria that is met? Is there a process to become one? 

This process doesn't quite care about it at the moment. BU should let engineering know that 'someone' wants to contribute their Task into our build/test Pipeline.

Internal or external partner - the stonesoup engineering team is notified that we should be expecting a contribution from a specific github handle. Once we are sure its a legitimate contribution ( and CI has OK'd it ). Also, we shall take a call on whether the image used in the Tekton Task needs to be in registry.redhat.io .

ADR/partner-tasks.md
sbose78: https://redhat-internal.slack.com/archives/C02CTEB3MMF/p1678804269686459 See thread on the concern.

sbose78:
Updated ADR to specify that we are going to use https://github.com/redhat-appstudio/build-definitions as the repository where partners would contribute their Tasks.

ADR/partner-tasks.md
```diff
@@ -0,0 +1,78 @@
...
+
+* Use the github.com/redhat-appstudio/build-definitions for Task submissions by partners : This is being considered in the short-term, either way, the day-to-day operations will not quite change.
+* Use Tekton Hub for host Tasks : Tekton Hub is being deprecated.
+
+
+## Consequences
+
+* We have a mechanism to take in Tekton Tasks from Partners with due diligence.
+* We should also be able to take in Tekton Tasks from Red Hat teams with the same level of validation/diligence that Red Hat Partners would have to go through.
+* We do build up a little tech-debt because this process needs to be merged with the official Red Hat certification process in the future.
```
ralphbean, line 75:
```suggestion
* We do build up a little tech-debt because this process needs to be merged with the official Red Hat certification process in the future, but we try to keep that tech debt to a minimum by building on top of our existing process around the build-definitions repo rather than trying to create a new second repo and process.
```

ralphbean (approved):

dirgim (approved):
LGTM

sonam1412 (approved):
LGTM

ADR/partner-tasks.md
```diff
@@ -0,0 +1,77 @@
+# Partner Tasks in Build/Test Pipelines
+
+Date: 2023-03-06
+
+## Status
+
+Provisional
```
ralphbean, line 7: +1 to merge, @sbose78 

```suggestion
Accepted
```

PR #64, opened 2023-03-06
 Specify a custom token for use with the build/test pipeline

sbose78:
As a user, I'd like to specify a custom token for use with the build/test pipeline, especially with a parter Task.

https://issues.redhat.com/browse/STONE-343


ADR/user-specified-tokens-build.md
```diff
@@ -0,0 +1,28 @@
...
+
+Provisional
+
+
+## Context
+
+As a user, I would like to configure sensitive information/tokens/credentials for the build/test pipeline to consume as part of a partner task.
+
+
+## Decision
```
gbenhaim, line 15: Can you please describe how secrets rotation/revocation will be handled?

ADR/user-specified-tokens-build.md
```diff
@@ -0,0 +1,28 @@
...
+
+1. UI will take in sensitive information as form input and upload it to SPI using the user's authentication token. See https://github.com/redhat-appstudio/service-provider-integration-operator/blob/main/docs/USER.md#uploading-access-token-to-spi-using-kubernetes-secret .
+2. In the above form, the user should be able to choose the name of the `Secret` from a dropdown. 
+The dropdown would be a pre-populated list of `Secret` names associated with known partner integrations. The user should also be allowed to provide a name barring the ones that already exist in the namespace.
```
gbenhaim, line 19: we need a naming convention to avoid the collision of secret names between partner tasks.

ADR/user-specified-tokens-build.md
```diff
@@ -0,0 +1,28 @@
...
+3. The UI will create an `SPIAccessTokenBinding` CR that would create and link the secret with the _pipeline_ `serviceaccount`. See https://github.com/redhat-appstudio/service-provider-integration-operator/blob/main/docs/USER.md#linking-a-secret-to-a-pre-existing-service-account .
```
gbenhaim, line 20: I guess that you are referring to the pipeline service account in the user's namespace?
If so, when a user has multiple environments (which in the first implementation map to namespaces), in which namespace the integration pipeline runs?

ADR/user-specified-tokens-build.md
Michkov, line 19: How this list of Secrets will be provided?

ADR/user-specified-tokens-build.md
```diff
@@ -0,0 +1,28 @@
...
+ 
+
+
+## Consequences
+
+* Users would be able to create custom secrets that partner Tasks could consume.
+* Partner Tasks would need to support reading the token from linked service account, instead of it being passed as a pipeline parameter.
```
gbenhaim, line 27: IIUC the proposal, secrets are global to the pipeline, so any task can consume them. I think that we need to mention it to the users when he/she upload secrets or modify the pipeline to include additional tasks.

ADR/user-specified-tokens-build.md
amfred: Proposal:
The user revokes the secret outside of Stonesoup.
The user generates a new credential outside of Stonesoup.
The user overwrites the secret in Stonesoup with a new one using our UI.

I propose that we don't have any more sophisticated secret rotation on our side at this point.

ADR/user-specified-tokens-build.md
chmouel: or perhaps better a completely random name and some sensible labels?

skabashnyuk (commented):
Do we know if we need to spread this secret to multiple k8s clusters?

ADR/user-specified-tokens-build.md
arewm: Should we also support simply _deleting_ a secret in the UI?

ADR/user-specified-tokens-build.md
arewm: I think completely random variable names would be worse from a UI perspective than scoping variables to a specific task. Scoping variables to a task would also make it easier to configure the task in a Pipeline definition as the secret configuration would remain constant and wouldn't have to be fiddled with based on the secret set. Even if the "sensible labels" were used for displaying the dropdown instead of the name of the secret, debugging processes would likely be complicated if randome secret names were used instead.

ADR/user-specified-tokens-build.md
```diff
@@ -0,0 +1,28 @@
...
+
+
+## Context
+
+As a user, I would like to configure sensitive information/tokens/credentials for the build/test pipeline to consume as part of a partner task.
+
+
+## Decision
+
+1. UI will take in sensitive information as form input and upload it to SPI using the user's authentication token. See https://github.com/redhat-appstudio/service-provider-integration-operator/blob/main/docs/USER.md#uploading-access-token-to-spi-using-kubernetes-secret .
```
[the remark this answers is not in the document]
arewm: I would _assume_ that a user would enter a single plaintext secret string in the expected format for the partner task and that the system would do any encoding needed to put it into a k8s secret. If a task needs more than one secret to run properly, the task owner can decide to require multiple secrets or to generate a combined secret (i.e. json map) and unpack it internally.

ADR/user-specified-tokens-build.md
```diff
@@ -0,0 +1,28 @@
...
+Date: 2023-03-03
+
+## Status
+
+Provisional
+
+
+## Context
+
+As a user, I would like to configure sensitive information/tokens/credentials for the build/test pipeline to consume as part of a partner task.
```
arewm, line 12: Is this explicitly for the use within partner tasks, or is it generically to be used by tests in the workspace (i.e. if credentials are needed for user-provided integration tests)?

ADR/user-specified-tokens-build.md
arewm, line 19: Building on an earlier question about whether these are just for partner tasks, would a user be able to enter a freeform secret name if they were able to leverage them in their integration tests?

ADR/user-specified-tokens-build.md
arewm: The scope of the secrets should definitely be mentioned in the document -- i.e. whether they are only in the target workspace or more broadly set.

ADR/user-specified-tokens-build.md
arewm: >  I think that we need to mention it to the users when he/she upload secrets or modify the pipeline to include additional tasks.

Mention it to users when they are configuring tasks, mention it to the users when configured tasks are skipped, or both?

ADR/user-specified-tokens-build.md
arewm: Is there any protection that Stonesoup could implement to invalidate/delete/etc. if secrets are referenced in tasks other than the intended target? This would have the potential to protect users from accidentally printed secrets as well as potentially "malicious" tasks that try to just get all secrets in the workspace.

I imagine this would be out of scope for an initial implementation, but it sounds like a good protective measure longer term.

sbose78:
TODO: We don't really need secret linking. Adding more details.

ADR/user-specified-tokens-build.md
sbose78: Updated the ADR:

* UI is responsible for the name of the `Secret`. UI has the freedom to choose legal names based on what's available on the namespace.
* UI sets a discoverable label which the build-service or integration service would depend on.

ADR/user-specified-tokens-build.md
sbose78: It would be both.

ADR/user-specified-tokens-build.md
sbose78: Since the `Secret` name is more of a runtime implementation detail, I reworked the ADR to say that users shouldn't really have any say in the name of the `Secret` - they should only have a say on what the associated parameter name is ( and the content of the `Secret` of course ). The UI should definitely notify the user what the name of the `Secret` is though, so that the user is able to reference it in the customized Pipeline. What do you think ?

ADR/user-specified-tokens-build.md
sbose78: Thanks for the question, and I like @amfred 's proposal. That's a good question @arewm !

ADR/user-specified-tokens-build.md
sbose78: Yes, in the user's namespace at the moment. I checked with @dirgim , he mentioned that, at the moment, we are going to be running integration tests in the user's namespace itself.

I thought it would be implicit but SPI today supports SPIAccessTokenBinding in the same namespace only. Let me update the ADR.

ADR/user-specified-tokens-build.md
sbose78: This would be a hardcoded list of parameter names.

sbose78: * UI shows the user the list of parameter names that she could optionally set.

sbose78:
Updates to the ADR:

- UI shows the user the list of parameter names that she could optionally set.
- UI is responsible for the name of the Secret. UI has the freedom to choose legal names based on what's available on the namespace.
- UI sets a label on the Secret, that makes the Secret discoverable to the build-service or integration service.
- Secret will not be linked to the SA.

ADR/user-specified-tokens-build.md
arewm: That is awesome -- but I think that also means that we have to loosen the language from "partner task" in the ADR as that seems to be too narrowly focused. Can it just be generalized to "a part of a task"?

If we want to call out the fact that this enables the configuring and running of partner tasks, then can we relegate that wording/comment to the consequences?

ADR/user-specified-tokens-build.md
arewm: As long as there is a clear user direction between the secret need (i.e. partner task), setting, and using (i.e. in integration tests) then I think we should be good. 

As an example, after configuring a secret (or if we can see the secrets that are set for the workspace), then we can get a display bubble showing something to the effect of "In order to leverage this secret in a task, do X or Y."

Looking at the proposed changes, I think this relationship (as well as the irrelevance of the secret name) makes sense.

ADR/user-specified-tokens-build.md
```diff
@@ -0,0 +1,64 @@
...
+
+| Pipeline Parameter    | Contents of the Secret | Name of the Secret        |
+| --------------------  | ---------------------- | ------------------------  |
+| codecov-github-token  | *input from the user*  | *autogenerated*           | 
+| codecov-token-secret  | *input from the user*  | *autogenerated*           |
+| redhat-ocm-token      | *input from the user*  | *autogenerated*           | 
+| *input from the user* | *input from the user*  | *autogenerated*           | 
+
+
+The build-service or the integration-service, reponsible for generating the PipelineRun needs to be able to map the name of the PipelineRun `parameter` with that of the `Secret` created by the UI, with contents provided by the user. To help the build-service do so, the `secrets` would be labelled with the name of the parameter for the build-service to look up, by the UI.
```
arewm, line 38: How can we prevent malicious tasks from getting access to secrets in a workspace that they shouldn't? 
* Create a pattern of isolation across tasks
  * If we can enforce that tasks need to have a unique identifier/prefix, then we can ensure that there is adequate enough isolation between task maintainers
  * We would still want to enable task maintainers to reuse secrets across tasks, so we don't want to force too fine of a granularity
  * Tasks should not try to request secrets outside their namespace
* Enforce pattern of isolation
  * This could just be enforced on the task submission side with "due diligence"
  * An additional guarantee would be to have additional checks on the build- or integration-service side preventing secret mounting

ADR/user-specified-tokens-build.md
arewm: Sure. Let me know if you want to chat about it.

ADR/user-specified-tokens-build.md
```diff
@@ -0,0 +1,64 @@
...
+ secret:
+   type: kubernetes.io/basic-auth
+   labels:
+     "appstudio.redhat.com/codecov-github-token":"true"
+```
+
+
+## Consequences
+
+* The build-service and the integration service needs to be able to read the contents of all secrets in the namespace.
```
arewm, line 57: The services will also have to iterate through all secrets to look for potential matches.

ADR/user-specified-tokens-build.md
```diff
@@ -0,0 +1,64 @@
...
+spec:
+ repoUrl: https://snyk.io/
+ secret:
+   type: kubernetes.io/basic-auth
+   labels:
+     "appstudio.redhat.com/codecov-github-token":"true"
+```
+
+
+## Consequences
```
arewm, line 55: An additional consequence is that if a user wants to configure a secret on multiple workspaces (i.e. for integration tests), they will need to set the secret multiple times, even if the content is the same.

ADR/user-specified-tokens-build.md
sbose78: Ideally, they should search by label, since the UI should be setting that via the SPIAccessTokenBinding CR.

ADR/user-specified-tokens-build.md
```diff
@@ -0,0 +1,64 @@
...
+As a user, I would like to configure sensitive information/tokens/credentials for the build/test pipeline to consume as part of a partner task.
+
+
+## Decision
+
+### How would the sensitive information be persisted and made available to the PipelineRun ?
+
+1. The UI would take in sensitive information as form input and upload it to SPI using the user's authentication token. See https://github.com/redhat-appstudio/service-provider-integration-operator/blob/main/docs/USER.md#uploading-access-token-to-spi-using-kubernetes-secret .
+
+2. Subsequently, the UI would create an `SPIAccessTokenBinding` CR with the appropriate label in the `.spec.secret.labels` to make the `Secrets` available to the build service or the integration service.
```
ralphbean, line 21: Does the `SPIAccessTokenBinding` expire?

What happens when future build pipelineruns are started by PaC after it has expired? Should something be responsible for re-generating it just-in-time?

ADR/user-specified-tokens-build.md
Michkov: With customization it's not possible to prevent - user can manually modify the pipelinerun definition.

arewm: Oof. this sounds like a large problem. Thinking out loud...

This would be a problem _today_ -- not just if we start accepting partner tasks. Can the build or integration service (whomever is responsible for untangling the mappings) not perform validation during the mapping process? If tasks are not referencing secrets _directly_ then the services can just resolve and provide all secrets "scoped" to a task? That wouldn't necessarily help because the tasks can still be customized. Can we prevent the pipeline definitions from modifying certain tasks or detect if modifications have been made to tasks, preventing secrets from being set if the task is not in "approved form"?

ADR/user-specified-tokens-build.md
sbose78: Does this help https://github.com/redhat-appstudio/book/pull/64/commits/1191aa8bca2e893fe1765be91f4c257bd7c3f673 to begin with ?

arewm: It is helpful in that we mention integration tests somewhere now. Can we just strike `partner` from this line or add `or user`?

ADR/user-specified-tokens-build.md
arewm: @sbose78, maybe you meant to respond on this comment. This is better with the latest push. :)

ADR/user-specified-tokens-build.md
scoheb: The CR has an expiration setting. There is a default but you can set to `-1` for no expiration.

ADR/user-specified-tokens-build.md
sbose78: We do plan on having enforcement around the level of privilege a PipelineRun workload could have 
1) by limiting how powerful the `pipeline` sa is.
2) by ensuring the user is not able to use a different 'sa' available in the namespace.

> How can we prevent malicious tasks from getting access to secrets in a workspace that they shouldn't?

By educating users about not putting random Tasks off the internet in their Git .tekton/ directory. Therefore, it is in the hands of the users to not introduce a malicious Task in their workspace.

If they did manage to do so out of a lapse in good decision-making, only the Components in the specific namespace would be impacted.

sbose78: >  If tasks are not referencing secrets directly then the services can just resolve and provide all secrets "scoped" to a task? That wouldn't necessarily help because the tasks can still be customized.

+1

> Can we prevent the pipeline definitions from modifying certain tasks or detect if modifications have been made to tasks, preventing secrets from being set if the task is not in "approved form"?

We can mostly do things around : 
a) validate the image used to run Task container
b) validate the bundle being used to reference the Task
c) validating the yaml isn't full-proof but definitely possible using stackrox-like solutions.

However, ultimately, if the binary in the Task's image tries to do something malicious, it needs to be stopped by the boundaries of the pipeline `sa` privileges.

ADR/user-specified-tokens-build.md
gbenhaim: I would say when uploading the secret.

ADR/user-specified-tokens-build.md
gbenhaim: @sbose78 if the @amfred proposal is the accepted way, can you please add it to this ADR?

ADR/user-specified-tokens-build.md
ralphbean: Today I learned!

ADR/user-specified-tokens-build.md
sbose78: https://github.com/redhat-appstudio/book/pull/64/commits/d1a2fd060af3e9317e98a03e76938c06d30f8b8b done!

ralphbean:
@skabashnyuk to your question:

> Do we know if we need to spread this secret to multiple k8s clusters?

Today, I think the answer is "no", we do not need to spread it to multiple member clusters. If the _workspace_ is allocated to member cluster "foo01", then the `SPIAccessTokenBinding` and the `PipelineRun` should both only appear in the `username-tenant` namespace on the "foo01" member cluster. The secret should not be made available or bind-able in a hypothetical `username-tenant` namespace on the "foo02" member cluster, if such a namespace were to be created.

ADR/user-specified-tokens-build.md
```diff
@@ -0,0 +1,75 @@
...
+
+| Pipeline Parameter    | Contents of the Secret | Name of the Secret        |
+| --------------------  | ---------------------- | ------------------------  |
+| codecov-github-token  | *input from the user*  | *autogenerated*           | 
+| codecov-token-secret  | *input from the user*  | *autogenerated*           |
+| redhat-ocm-token      | *input from the user*  | *autogenerated*           | 
+| *input from the user* | *input from the user*  | *autogenerated*           | 
+
+
+The build-service or the integration-service, reponsible for generating the PipelineRun needs to be able to map the name of the PipelineRun `parameter` with that of the `Secret` created by the UI, with contents provided by the user. To help the build-service do so, the `secrets` would be labelled with the name of the parameter for the build-service to look up, by the UI.
```
Michkov, line 43: When user add secret (seems to be global scope) then we should check all his components if pipelineruns in .tekton folder does contain the parameter and if so then create pull-request -> filling the secret?

ADR/user-specified-tokens-build.md
```diff
@@ -0,0 +1,75 @@
...
+The build-service or the integration-service, reponsible for generating the PipelineRun needs to be able to map the name of the PipelineRun `parameter` with that of the `Secret` created by the UI, with contents provided by the user. To help the build-service do so, the `secrets` would be labelled with the name of the parameter for the build-service to look up, by the UI. 
+
+```
+apiVersion: appstudio.redhat.com/v1beta1
+kind: SPIAccessTokenBinding
+metadata:
+ name: doesnt-matter
+ namespace: user-tenant
+spec:
+ repoUrl: https://snyk.io/
```
Michkov, line 52: should have also `lifetime: '-1'`  so that secret is persistent.

ADR/user-specified-tokens-build.md
```diff
@@ -0,0 +1,75 @@
...
+ secret:
+   type: kubernetes.io/basic-auth
+   labels:
+     "appstudio.redhat.com/codecov-github-token":"true"
+```
+
```
Michkov, line 58: What is process of secret deletion?

ADR/user-specified-tokens-build.md
```diff
@@ -0,0 +1,75 @@
...
+While uploading a token, the user should be able to choose the name of the Pipeline `parameter` from a UI dropdown. The dropdown would be a pre-populated and hard-coded list of names associated with known partner integrations. 
+
+The user should also be allowed to provide a parameter name of choice. The same could be used in a customized PipelineRun definition in the user's source code's .tekton/ directory.
+
+| Pipeline Parameter    | Contents of the Secret | Name of the Secret        |
+| --------------------  | ---------------------- | ------------------------  |
+| codecov-github-token  | *input from the user*  | *autogenerated*           | 
+| codecov-token-secret  | *input from the user*  | *autogenerated*           |
+| redhat-ocm-token      | *input from the user*  | *autogenerated*           |
```
Michkov, line 39: These partner integration should be included in our default build pipelines? Otherwise users would need to define them in .tekton dir which is not user-friendly.

ADR/user-specified-tokens-build.md
```diff
@@ -0,0 +1,75 @@
...
+
+
+The build-service or the integration-service, reponsible for generating the PipelineRun needs to be able to map the name of the PipelineRun `parameter` with that of the `Secret` created by the UI, with contents provided by the user. To help the build-service do so, the `secrets` would be labelled with the name of the parameter for the build-service to look up, by the UI. 
+
+```
+apiVersion: appstudio.redhat.com/v1beta1
+kind: SPIAccessTokenBinding
+metadata:
+ name: doesnt-matter
+ namespace: user-tenant
```
sbose78, line 50: TODO : add label with the name of the Component.

ADR/user-specified-tokens-build.md
rohitkrai03, line 52: What should be the repoUrl? A fixed value for all secrets?

ADR/user-specified-tokens-build.md
```diff
@@ -0,0 +1,75 @@
...
+
+* The build-service and the integration service needs to be able to read the contents of all secrets in the namespace.
+* The UI would need to update the list of parameter names expected in the dropdown whenever a new partner is onboarded. 
+* The Build team would need to provide a hardcoded list of parameter names to the UI team that we expect users to be able to configure. This has to be done everytime a new partner is onboarded.
+* If a user wants to configure a secret on multiple workspaces (i.e. for integration tests), they will need to set the secret multiple times, even if the content is the same.
+
+## Out-of-scope
+
+* Making the senstive tokens available in other namespaces or clusters.
+* Handling a token rotation strategy. At the moment, as suggested, we will support updating/deletion of the token.
```
rohitkrai03, line 75: Do we need a UI for deleting secrets or tokens?

ADR/user-specified-tokens-build.md
```diff
@@ -0,0 +1,75 @@
...
+contributed to the StoneSoup Build/Test Pipeline by different organizations inside Red Hat.
+
+
+## Decision
+
+### How would the sensitive information be persisted and made available to the PipelineRun ?
+
+1. The UI would take in sensitive information as form input and upload it to SPI using the user's authentication token. See https://github.com/redhat-appstudio/service-provider-integration-operator/blob/main/docs/USER.md#uploading-access-token-to-spi-using-kubernetes-secret .
+
+2. Subsequently, the UI would create an `SPIAccessTokenBinding` CR with the appropriate label in the `.spec.secret.labels` to make the `Secrets` available to the build service or the integration service.
```
ralphbean, line 26: Can you revise the text to give an example of the appropriate label?

Here, do you mean the `tekton.dev/...` annotations (not labels) that when they appear on a `Secret`, cause tekton to perform credential initialization? https://github.com/tektoncd/pipeline/blob/main/docs/auth.md#understanding-credential-selection

ralphbean: Oh, no, I see now. I think you intend for this to be a custom labeling (like `appstudio.redhat.com/<parameter-name>` +/- the component's name), to be interpreted by our controllers (build-service and integration-service).

ADR/user-specified-tokens-build.md
ralphbean, line 43: What should integration-service do here? The pipelinerun that it will initiate is provided by the user in a form resolvable by the git resolver: https://issues.redhat.com/browse/STONE-53

ralphbean: Suggesting an alternative in https://github.com/redhat-appstudio/book/pull/75

ralphbean:
> We don't really need secret linking. 

@sbose78 can you explain why not?

sbose78:
@ralphbean , I took in your suggested change. ⚠️ @rohitkrai03 , we have a change in the ADR.

I was avoiding users' exposure to secret names but I realized there's no way out - secret names are going to be out in the open in users' .tekton/ repositories. Also, having the UI generate the secret name ( with randomized strings ) and then have the same make its way into user's .tekton/pipelinerun.yaml would make https://issues.redhat.com/browse/STONE-680 harder to achieve.

ADR/user-specified-tokens-build.md
sbose78: Accepted 🟢

ADR/user-specified-tokens-build.md
sbose78: Good question, repoURL doesn't really matter. @skabashnyuk ?

ADR/user-specified-tokens-build.md
```diff
@@ -0,0 +1,89 @@
...
+
+1. The user revokes the secret outside of Stonesoup.
+2. The user generates a new credential outside of Stonesoup.
+3. The user overwrites the secret in Stonesoup with a new one using our UI.
+
+## Consequences
+
+* The user has to _know_ what secret name is expected by their pipelinerun, and supply that in their
+  upload secret through our UI. We can provide suggestions for known partner integrations in our UI,
+  but this will need to permit open-ended input from users to permit completely custom user-supplied
```
sbose78, line 77: >  open-ended input from users to permit completely custom user-supplied Tasks

Unlike secrets for most partner Tasks which would not be Component-specific, completely user-supplied Tasks will warrant Component-specific secrets. 

The UI needs to expose a form input field where the user would be able to specify whether a `Secret` is for 'global' use or for a Component.

ADR/user-specified-tokens-build.md
```diff
@@ -0,0 +1,139 @@
...
+| Name of the Secret   | Token data |
+| ----------- | ----------- |
+| synk-secret      | *user-provided*   |
+| tidelift-token   | *user-provided*   |
+| *user-provided*  | *user-provided*   |
+
+
+2. Subsequently, the UI would create an `SPIAccessTokenBinding` CR with the appropriate name in the `.spec.secret.name` such that the secret appears under a well-known name. The well-known name would be specified in the Task definition by the Task author.
+
+3. The PipelineRun must reference the secret by name.
```
sbose78, line 31: @Michkov 
This won't be needed if the `default` secret name is used. This is probably relevant for custom user-provided Tasks?

ADR/user-specified-tokens-build.md
```diff
@@ -0,0 +1,89 @@
...
+2. The user generates a new credential outside of Stonesoup.
+3. The user overwrites the secret in Stonesoup with a new one using our UI.
+
+## Consequences
+
+* The user has to _know_ what secret name is expected by their pipelinerun, and supply that in their
+  upload secret through our UI. We can provide suggestions for known partner integrations in our UI,
+  but this will need to permit open-ended input from users to permit completely custom user-supplied
+  tasks.
+* Partner tasks will need to advertise in their documentation what secret _name_ they expect to be
```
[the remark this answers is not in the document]
Michkov: Do we want to include partner tasks in our generic pipelineruns? If so then I would require the secret as parameter of the task and then predefine in the pipelinerun (.spec.tasks[xxx].params) with empty string so that user can easily add it. And if not defined then skip.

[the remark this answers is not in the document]
jduimovich: I would say for some partner tasks, the default pipelines should include them this is called our as an option to _promote_ a partner task. The PM needs to decide which partners get this _special treatment_ and our team pre-integrate their tasks into our default pipelines. In other cases, for now, they will need to customize the pipelines and manually insert the task/secret

ADR/user-specified-tokens-build.md
```diff
@@ -0,0 +1,139 @@
...
+
+
+| Name of the Secret   | Token data |
+| ----------- | ----------- |
+| synk-secret      | *user-provided*   |
+| tidelift-token   | *user-provided*   |
+| *user-provided*  | *user-provided*   |
+
+
+2. Subsequently, the UI would create an `SPIAccessTokenBinding` CR with the appropriate name in the `.spec.secret.name` such that the secret appears under a well-known name. The well-known name would be specified in the Task definition by the Task author.
```
rohitkrai03, line 29: Where would the UI get these appropriate name? Is it going to be a fixed set that UI knows?

ADR/user-specified-tokens-build.md
```diff
@@ -0,0 +1,139 @@
...
+  name: my-upload-secret
+  labels:
+    spi.appstudio.redhat.com/upload-secret: token
+    spi.appstudio.redhat.com/token-name: my-spi-access-token-binding-for-snyk
+type: Opaque
+stringData:
+  spiTokenName: my-spi-access-token-binding-for-snyk
+  providerUrl: https://snyk.io/
+  # The username is irrelevant, but has to be here for SPI to process this
+  userName: my-username-goes-here
```
rohitkrai03, line 55: So UI should add a random username here?

ADR/user-specified-tokens-build.md
rohitkrai03: So UI should use appropriate providers here based on the parameter selected?

rohitkrai03 (changes requested):

ADR/user-specified-tokens-build.md
arewm: @Michkov, are you trying to ask if it would be supported for there to be a secret which does not have a repoURL (i.e. to be associated with a custom task were created which didn't specify that)?

ADR/user-specified-tokens-build.md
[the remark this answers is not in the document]
arewm: It would likely be more part of the parter task flow, but could we conditionally require (or require generally as a best practice) to define default parameters so that a task would "gracefully fail" by default until an appropriate secret is provided?

ADR/user-specified-tokens-build.md
sbose78: > Yeah - it just represents the provider. It has to match between the upload secret and the spiaccesstokenbinding in order to have the right one injected.

Right, but apart from it being useful to surface 'text', does it really matter ? Since the linking is anyway happening because of the name of the secret.

ADR/user-specified-tokens-build.md
sbose78: Yes.

sbose78:
I think this is good to be merged ?

rohitkrai03 (commented):
@sbose78 I don't see the section about creating components with certain labels for attaching specific secrets to specific components. I thought there was one explaining that process.

elsony (commented):
Based on my understanding, the secret is currently tied to the component level. However, the user may want to use a different secret for different environments, e.g. prod vs staging, since they can be deployed to a different cluster. We should also provide the setting on the environment binding level so that it can be customized in each environment.

ADR/user-specified-tokens-build.md
```diff
@@ -0,0 +1,139 @@
...
+kind: Secret
+metadata:
+  name: my-upload-secret
+  labels:
+    spi.appstudio.redhat.com/upload-secret: token
+    spi.appstudio.redhat.com/token-name: my-spi-access-token-binding-for-snyk
+type: Opaque
+stringData:
+  spiTokenName: my-spi-access-token-binding-for-snyk
+  providerUrl: https://snyk.io/
```
christianvogt, line 53: Are `spiTokenName` and `providerUrl` required properties?

cmoulliard (commented):
Where is the `SPIAccessTokenBinding` CR defined (= which github repo) ? Is it documented part of the https://github.com/redhat-appstudio/book ? Which component will consume it (= name of the component packaging the controller processing it) ?

amfred:
@sbose78 I'm inclined to merge this in because the questions seem to be mostly resolved, and you can update it as needed.  What do you think?

PR #65, opened 2023-03-08
Log conventions ADR

ralphbean:
In support of [STONE-201](https://issues.redhat.com/browse/STONE-201).

This is a new PR including content from https://github.com/redhat-appstudio/book/pull/21 plus some re-work in https://github.com/amfred/book-of-appstudio/pull/2 based on feedback in the original PR. It is being re-created here so that the changes exist on a shared branch in the main book repo, so we can all edit.


ADR/0006-log-conventions.md
```diff
@@ -0,0 +1,112 @@
...
+produced the log. See section on automatically added logs below.
+
+### 4. Who was involved?
+
+Use the key `namespace` when logging the namespace of the *targeted resource* that is being modified
+or interacted with. (Note that the `namespace_name` key is automatically added by the fluent
+collectors and reflects the namespace in which the controller is running. See section on
+automatically added logs below.)
+
+Include a `user_id` if one exists and is applicable for the event being logged.
```
kim-tsao, line 57: Is this ADR specific to audit logs or all logs in general?  We should avoid collecting PII such as `user_id` in regular logs but they are acceptable in audit logs.  I believe audit logs are the only ones we are forwarding to CloudWatch/Splunk so we should make this distinction clear.

ADR/0006-log-conventions.md
```diff
@@ -0,0 +1,112 @@
...
+2023-03-07T11:32:29.020Z    INFO    controllers.Component   API Resource changed: UPDATE {"namespace": "samburrai-tenant", "audit": "true", "resource": "go-net-http-hello-prqx", "kind": "Component", "action": "UPDATE"}
+```
+
+The meaning of **timestamp** and log level should be self evident. The **logger** name helps your team understand which part of your controller is emitting the log (usually `Log.WithName` from `zap`). The **message** is a human readable string describing the event being logged. The **json data** contains additional fields useful for searching.
+
+### 1. When did it happen?
+
+Encode timestamps in UTC/ISO-8601 format.
+
+### 2. What happened?
```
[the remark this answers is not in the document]
arewm: In the second example above, `UPDATE` is recorded twice. Do we want duplication of information in the string and JSON or should they be mutually exclusive and _supportive_ of each other?

ADR/0006-log-conventions.md
arewm: Related to this, should data like `namespace_name` that is auto-added by fluentd be considered PII as well and handled specially?

ralphbean: > Is this ADR specific to audit logs or all logs in general?

I understand it to be for controller logs in general, including audit logs, in order to support the use cases in [STONE-200](https://issues.redhat.com/browse/STONE-200), [STONE-201](https://issues.redhat.com/browse/STONE-201), and [STONE-241](https://issues.redhat.com/browse/STONE-241).

> We should avoid collecting PII such as `user_id` in regular logs but they are acceptable in audit logs.

What you write makes sense.

> I believe audit logs are the only ones we are forwarding to CloudWatch/Splunk so we should make this distinction clear.

My understanding is that we desire _all_ logs to be forwarded to cloudwatch (for the purposes of our own team and SRE to debug and understand our system in production), but we desire that only audit logs be forwarded to splunk (for privileged query by security).

I am not aware of a method to easily log twice, once with a user id and once without, and ensure that one log is sent to cloudwatch (collected from pod stdout/err by fluent) while the other log is sent to splunk (forwarded how?).

Due to lack of clarity around the above, I think we should revise this to limit the scope of this ADR to being strictly about general logs, and drop this line about including a `user_id` if applicable. A consequence should be added at the end of the doc indicating that we'll need to revisit this if/when we need to include user ids specifically for audit logs in the future.

@kim-tsao, wdyt?

ADR/0006-log-conventions.md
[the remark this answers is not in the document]
ralphbean: Good call. Will revise.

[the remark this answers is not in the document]
ralphbean: WDYT of 043c00b?

ADR/0006-log-conventions.md
kim-tsao: @ralphbean that sounds reasonable, thanks!  Can you also clarify the question @arewm raised about `namespace_name` - can that be PII?

ADR/0006-log-conventions.md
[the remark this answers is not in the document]
arewm: It is better that it was but still hard to parse what should be included at a quick glance. Maybe something like this would simplify the representation.

```
## Decision: Log conventions

## Decision: Questions to answer in logs

### 1. When did it happen?

**Included in:** `TIMESTAMP`

[...]

### 2. What happened?

**Included in:** `MESSAGE`, `ADDITIONAL_JSON_DATA`

[...]

### 3. Where did it happen?

**Included in:** _none_

[...]
```

To summarize the required keys in `ADDITIONAL_JSON_DATA`, we can have a simplified representation in the `Consequences` section if it makes sense.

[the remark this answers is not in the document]
ralphbean: Updated in 842ec54.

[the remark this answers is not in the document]
arewm: Cool, thanks.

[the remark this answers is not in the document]
ralphbean: For the summarized ADDITIONAL_JSON_DATA, here's a jsonschema for it. Helpful? Or, overkill?

```json
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "type": "object",
  "properties": {
    "action": {
      "description": "Indicates what action the log event is responding to",
      "type": "string",
      "enum": ["VIEW", "ADD", "UPDATE", "DELETE"]
    },
    "namespace": {
      "description": "Indicates the target namespace of the resource being reconciled",
      "type": "string"
    },
    "kind": {
      "description": "The kind of the resource being reconciled",
      "type": "string"
    },
    "resource": {
      "description": "The name of the resource being reconciled",
      "type": "string"
    },
    "source": {
      "description": "A url to the source code that is emitting this log",
      "type": "string"
    },
    "audit": {
      "description": "Set to true for specific types of audit logs required by SSML.PW.5.1.4",
      "type": "boolean"
    }
  },
  "required": [],
  "not" {
    "required": [
      "namespace_name",
      "container_name",
      "pod_name",
      "container_image",
      "pod_ip",
      "host",
      "hostname",
      "namespace_labels",
      "message",
      "level",
      "time"
    ]
  }
}
```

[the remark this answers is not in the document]
arewm: That doesn't feel like overkill to me. It looks like a concise representation of the more detailed descriptions provided throughout!

ADR/0006-log-conventions.md
ralphbean: I bet it will be simpler to just align in both prod and stage - make them identical. But, let's leave that decision to the Core team (unless they ask for our input).

ralphbean: Updated in 0d130033b1b1bdfc4770c75ee2ebecbf23163138.

arewm (approved):

ADR/0006-log-conventions.md
amfred: The good news is that in our PIA, we said we would have PII in the logs. But limiting the usage of PII to "just enough" to serve our legitimate needs is the way to go.

amfred (approved):

ADR/0006-log-conventions.md
```diff
@@ -0,0 +1,196 @@
...
+
+Accepted
+
+## Context
+
+We need log conventions to make our offering easier to operate and maintain. These will also make it possible for us to create queries that cross components in our logs, which supports [STONE-201](https://issues.redhat.com/browse/STONE-201).
+
+## Decision: Log conventions
+
+In our controller logs, we will use structured log messages, partially formatted in JSON with key-value pairs as described below.
```
jgwest, line 16: It may be worth mentioning that, since all of our controllers (operators) are based on the controller-runtime framework (e.g. via operator-sdk), and it has an opinionated logging mechanism (utility function in `sigs.k8s.io/controller-runtime/pkg/log`, log frontend is logr, log backend is zap), that this logging mechanism does conform to the prescribed format, and can be used.

For example, using the standard controller-runtime logging mechanism will produce a text string in the prescribed format:
```go
logger "sigs.k8s.io/controller-runtime/pkg/log"

func f() {
    log = logger.FromContext(ctx)

    log.Info("log message", "jsonfield1", "jsonfield1value", etc... )
}
```

ADR/0006-log-conventions.md
```diff
@@ -0,0 +1,196 @@
...
+* Fluent Bit optimized configuration sends logs to `<kubernetes-nodeName>-application.var.log.containers.<kubernetes-podName>_<kubernetes-namespace>_<kubernetes-container-name>-<kubernetes-containerID>`
+* Fluentd sends logs to `<kubernetes-podName>_<kubernetes-namespace>_<kubernetes-containerName>_<kubernetes-containerID>`
+
+For more details on Fluentd vs. Fluent Bit logs, see [Set up Fluent Bit as a DaemonSet to send logs to CloudWatch Logs](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Container-Insights-setup-logs-FluentBit.html).
+
+It is best to avoid using the same key as the log collectors in our controller logs, to avoid confusion.
+
+### TaskRun logs
+
+TaskRun logs are specifically out of scope of this ADR. Those logs are typically meant to be viewed
```
harrietgrace, line 120: Should any logs related to the running application be called out here as well? Application logs generated after it's deployed would contain info that is meant to be viewed and should be human-readable. @jgwest @jannfis does that sound right to you?

ralphbean: Agreed, let's call it out of scope. In practice, we can't force the user's app's logs to conform to our scheme. I don't think it is desired, either.

ADR/0006-log-conventions.md
ralphbean: Incorporated in 5b55e6a39048c7afae1b3e22d6c6137d1481093d

ADR/0006-log-conventions.md
ralphbean: Added in 15ac904cc11502c2e4bf59eec60f334b9b08749c.

elsony (approved):

PR #66, opened 2023-03-10
ADR: Enable perf testing in StoneSoup

jhutar:
This is related to https://issues.redhat.com/browse/STONE-679 and was initially discussed with Ralph and Brian.


ADR/0018-apps-continuous-perf-testing.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Status
+
+In consideration
+
+## Context
+
+In general, performance testing is just another form of testing that helps application teams to ensure there are no regressions in their code and that their application behaves as expected.
+
+The IntegrationTestScenario pipelines in StoneSoup are not suitable for full-blown performance and scale testing (that usually takes lots of time and involves human analysis), but are a good place for a quick and small-scale regression test.
```
arewm, line 13: Why are they not sufficient? Can we expound upon this a little bit to inform what classes of problems might need to investigate a different testing path?

ADR/0018-apps-continuous-perf-testing.md
```diff
@@ -0,0 +1,72 @@
...
+>
+> If any of these metrics does not align to historical results, we mark this test result as a failure.
+
+And even if some test fails, the application team should be able to review the data and change the test definition, so next time a new test result is being evaluated based on the historical results, this new result will be one of them.
+
+The algorithm that says if the current test passes or fails when compared to historical results can vary. It can be:
+
+- Checking minimum and maximum of the metric in historical data and making sure the new result is in this range.
+- Using a floating average of historical data and setting a safe range around it and making sure the new result is in that range.
+- Some machine learning thingy.
```
arewm, line 34: Can this phrasing be more precise? You don't have to go into implementation details, but an idea/proposal/future direction that doesn't include `thingy` :)

ADR/0018-apps-continuous-perf-testing.md
```diff
@@ -0,0 +1,72 @@
...
+
+Goal of this ADR is to propose a way for this kind of testing to be implemented in Stonesoup (feature [STONE-679](https://issues.redhat.com/browse/STONE-679)). Even if it would not deprecate full-blown performance and scale testing, having a release gate with some short and small scale performance test is desirable for many application teams.
```
arewm, line 36: What are the differences between the proposed approach and "full-blown performance and scale testing"? Is there a comparison between them to highlight the benefits of one over the other? How is the current proposal not sufficient for a "full blown" test?

ADR/0018-apps-continuous-perf-testing.md
```diff
@@ -0,0 +1,72 @@
...
+
+We are considering two different options here. They differ in what portion of logic mentioned above is executed where.
```
arewm, line 38: You mention two options but only describe Horreum. What is the other option? Can we have a brief summary of both and then make a decision on Horreum if that is determined to be the best solution?

ADR/0018-apps-continuous-perf-testing.md
```diff
@@ -0,0 +1,72 @@
...
+- Horreum: a results storage service: <https://github.com/Hyperfoil/Horreum>
+- CPT: Continuous Performance Testing
+
+## Decision
+
+Let's use this architecture with a single Horreum instance that would be managed by Red Hat and used by all customers.
+
+![Architecture diagram with Horreum](assets/0018-apps-continuous-perf-testing.svg "Architecture diagram with Horreum")
+
+1. Performance test runs in the Tekton pipeline and generates JSON with test parameters and results.
```
arewm, line 52: Are there standards for this JSON format? Do we need to be prescriptive of how to record data? Have we considered what would be required for tenant isolation?

ADR/0018-apps-continuous-perf-testing.md
```diff
@@ -0,0 +1,72 @@
...
+
+### Glossary
+
+- SUT: Subject Under Test
+- Horreum: a results storage service: <https://github.com/Hyperfoil/Horreum>
+- CPT: Continuous Performance Testing
+
+## Decision
+
+Let's use this architecture with a single Horreum instance that would be managed by Red Hat and used by all customers.
```
arewm, line 48: Is this something that would be available to all internal and external tenant clusters?

ADR/0018-apps-continuous-perf-testing.md
```diff
@@ -0,0 +1,72 @@
...
+
+![Architecture diagram with Horreum](assets/0018-apps-continuous-perf-testing.svg "Architecture diagram with Horreum")
+
+1. Performance test runs in the Tekton pipeline and generates JSON with test parameters and results.
+2. We gather configured monitoring metrics and add them to the JSON.
+3. We upload the JSON with all the data to Horreum.
+4. Horreum performs result analysis, looking for changes in configured metrics.
+4. We get PASS/FAIL decision from Horreum back to the pipeline, so pipeline can rerurn proper result.
```
arewm, line 56: How long does it take for the analysis to be run? How is the pipeline run? Is it run as an `IntegrationTestScenario`?

ADR/0018-apps-continuous-perf-testing.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Consequences
+
+Pros:
+
+- We can provide some best practices to customers wanting to do some basic performance regression testing.
+- Horreum is end-to-end solution that already exists, is already used by multiple teamsn and has documentation for the whole process.
```
arewm, line 63:
```suggestion
- Horreum is end-to-end solution that already exists, is already used by multiple teams and has documentation for the whole process.
```

ADR/0018-apps-continuous-perf-testing.md
```diff
@@ -0,0 +1,72 @@
...
+- StoneSoup UI can display all the trends because historical data will be available in Horreum.
+- Red Hat team that develops Horreum is willing to help with this effort.
+- Integration scripts that would be running in the pipeline (to gather monitoring data, to upload results to Horreum, to get PASS/FAIL decision from Horreum...) mostly exists.
```
arewm, line 66: What additional work would be needed as part of this? How much of the additional work here is already "available"? How much needs to be developed? How much is customized to Stonesoup?

ADR/0018-apps-continuous-perf-testing.md
```diff
@@ -0,0 +1,72 @@
...
+
+Cons:
+
+- Somebody has to manage the Horreum instance. Required development time on it.
+- Although Horreum uses Keycloak as well, some changes in the Horreum authentication mechanism might be required.
+- Maybe also some scaling concerns of Horreum.
```
arewm, line 72: What does this mean? How concerned should I be with this point? Is there a threshold where this starts to become _more_ of a concern?

ADR/0018-apps-continuous-perf-testing.md
arewm, line 48: How resource intensive is the Horreum instance? Is it possible to separate out its use per workspace? @amisstea, FYI.

ADR/0018-apps-continuous-perf-testing.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Decision
+
+Let's use this architecture with a single Horreum instance that would be managed by Red Hat and used by all customers.
+
+![Architecture diagram with Horreum](assets/0018-apps-continuous-perf-testing.svg "Architecture diagram with Horreum")
+
+1. Performance test runs in the Tekton pipeline and generates JSON with test parameters and results.
+2. We gather configured monitoring metrics and add them to the JSON.
+3. We upload the JSON with all the data to Horreum.
```
arewm, line 54: Who is the `We` here? Is it some Pipeline definition that can be used commonly across the board? Will there be some service running that will need to be integrated with for performance testing?

ADR/0018-apps-continuous-perf-testing.md
```diff
@@ -0,0 +1,72 @@
...
+> Example: Imagine you develop a web e-shop application that uses PostgreSQL data backend. Your perf test is browsing through a goods catalog and measures latency of pages. When you want to decide on pass or fail result of this test, you need to check that metrics like below are aligned to previous results of a same test with same configuration:
+>
+> - Average page load latency.
+> - Backend service CPU and memory usage...
+> - PostgreSQL CPU and memory usage…
+> - Internal PostgreSQL metrics like number of commits during the test or average number of database sessions during the test…
+>
+> If any of these metrics does not align to historical results, we mark this test result as a failure.
+
+And even if some test fails, the application team should be able to review the data and change the test definition, so next time a new test result is being evaluated based on the historical results, this new result will be one of them.
```
arewm, line 28: Not really related to this point, but is there a concept of a "better fail" as well where you would be testing to see if a failure happened less catastrophically?

ADR/0018-apps-continuous-perf-testing.md
jhutar: Hello! I think this depends on how are you doing performance and scale testing. In my expirience it is usually far from being fully automated as it involves lots of iterations and human analysis of resutls. Also some testing can be very expensive and is not happening often. The testing we are trying to propose here is just an fully automated performance regression check. I have added some more explanation to `The IntegrationTestScenario pipelines in StoneSoup are not suitable...` paragraph about as well.

ADR/0018-apps-continuous-perf-testing.md
jhutar: > Are there standards for this JSON format?

It can be any valid JSON - Horreum can handle it. I would consider it implementation detail as of now.

> Do we need to be prescriptive of how to record data?

No, because it will be different for every test - if you are testing an web app, you will be interested in latency and number number of requests per second and failure rate... and if you are testing application processing kafka messages, you will measure per message processing duration and amount of data processed...

> Have we considered what would be required for tenant isolation?

If you are asking about isolating test runs itself: Yes, noisy neighbours will cause performance test results will vary and it is up to test design, deployment requirements (e.g. make sure your pods have requests and limits configured) and results evaluation to overcome this (we have some experience with it).

If you are asking about if Horreum can handle multiple tenants: Yes, Horreum uses Keycloak for fine grained user permissions.

ADR/0018-apps-continuous-perf-testing.md
jhutar: > How long does it take for the analysis to be run?

Analysis on Horreum would be (almost) instant.

> How is the pipeline run?

Actual test run would take time as per the test case. Every application have to have it's own same as every application have to have it's own functional tests.

> Is it run as an IntegrationTestScenario?

I'm not too knowledgeable about StoneSoup internals, but I assume this all will be just another Tekton PipelineRun.

ADR/0018-apps-continuous-perf-testing.md
jhutar: I will ask John O'Harra to answer this.

ADR/0018-apps-continuous-perf-testing.md
jhutar: As of now, I know multiple teams are using Horreum and it is far from it's full capacity. But to be fair, there was no propper perfomance&scale testing of Horreum itself as far as I know, so we really do not know what is the actual scale Horreum can operate at. It is same for StoneSoup itself (I'm part of the team that tries to perf&scale test StoneSoup itself - see STONE-224). So the answer is "I do not know" as of now.

ADR/0018-apps-continuous-perf-testing.md
jhutar: Again, will ask John to answer here.

ADR/0018-apps-continuous-perf-testing.md
arewm: Fine to leave it as an implementation detail for now, I just wanted to make sure that it was one and that we didn't need to identify some standard.

Re: multiple tenants, I was mostly interested in ensuring that tests in one workspace wouldn't be able to populate/retrieve data from another workspace. This sounds like Keycloak would be sufficient. Can you just call this bit out somewhere as a reminder if its importance?

ADR/0018-apps-continuous-perf-testing.md
arewm: So you are proposing that a single Horreum instance would be used to store, analyze, and output data from both the internal Red Hat tenant cluster and the external public tenant cluster? I feel like this is contrary to the standard. @ralphbean, is there any record of isolation between tenant clusters to be used as a precedence?

ADR/0018-apps-continuous-perf-testing.md
arewm: We can be more descriptive here then -- mentioning the specific scale concerns (i.e. untested). This might give credence to having multiple Horreum instances, one per tenant.

ADR/0018-apps-continuous-perf-testing.md
jhutar: At the end it is up to the customer what it will run in the pipeline (same as for functional tests), but AFAICT Tekton Pipeline in StoneSoup needs to end with either PASS or FAIL.

On the other hand, it would be possible for StoneSoup UI to distinguish failures. E.g. if you are measuring latency and historically your latency was 1.00s +- 0.10s, but your new test run measured latency 1.11s, it is still fail (depends on your threshold settings), but it might be other kind of fail than is the new latency is 3.00s.

ADR/0018-apps-continuous-perf-testing.md
jhutar: Added some more info. Hope it is better now.

ADR/0018-apps-continuous-perf-testing.md
jhutar: OK, made it clearer I hope :)

ADR/0018-apps-continuous-perf-testing.md
jhutar: Sorry, this is left-over from some previous edits.

ADR/0018-apps-continuous-perf-testing.md
jhutar: Thank you! Reported https://github.com/Hyperfoil/Horreum/issues/420 and added a line to the ADR about it to "Cons" section.

ADR/0018-apps-continuous-perf-testing.md
jhutar: I assume it is similar to Tekton here: there is one Tekton that is going to run pipelines from both internal teams and external customers? But maybe I'm missing your point.

ADR/0018-apps-continuous-perf-testing.md
jhutar: Fixed (together with few more typos :)).

ADR/0018-apps-continuous-perf-testing.md
```diff
@@ -0,0 +1,69 @@
...
+
+- We can provide some best practices to customers wanting to do some basic performance regression testing.
+- Horreum is end-to-end solution that already exists, is already used by multiple teams and has documentation for the whole process.
+- StoneSoup UI can display all the trends because historical data will be available in Horreum.
+- Red Hat team that develops Horreum is willing to help with this effort.
+- Integration scripts that would be running in the pipeline (to gather monitoring data, to upload results to Horreum, to get PASS/FAIL decision from Horreum...) mostly exists.
+
+Cons:
+
+- Somebody has to manage the Horreum instance. Required development time on it.
```
ralphbean, line 67:
```suggestion
- Somebody has to manage the Horreum instance. Required development time on it. Service instance needs ownership.
```

ADR/0018-apps-continuous-perf-testing.md
jhutar: I see! Great. Changed the wording to mention that.

ADR/0018-apps-continuous-perf-testing.md
```diff
@@ -0,0 +1,69 @@
...
+- Although Horreum uses Keycloak as well, some changes in the Horreum authentication mechanism might be required.
```
ralphbean, line 68:
```suggestion
- Although Horreum uses Keycloak as well, some changes in the Horreum authentication mechanism might be required to cooperate with Stonesoup member cluster RBAC.
```

ADR/0018-apps-continuous-perf-testing.md
```diff
@@ -0,0 +1,69 @@
...
+Let's use this architecture with a single Horreum instance that would be managed by Red Hat and used by all customers.
+
+![Architecture diagram with Horreum](assets/0018-apps-continuous-perf-testing.svg "Architecture diagram with Horreum")
+
+1. Performance test runs in the Tekton pipeline and generates JSON with test parameters and results.
+2. Pipeline gathers configured monitoring metrics and add them to the JSON.
+3. Pipeline uploads the JSON with all the data to Horreum.
+4. Horreum performs result analysis, looking for changes in configured metrics.
+4. Pipeline gets PASS/FAIL decision from Horreum back to the pipeline, so pipeline can return proper result.
+
```
ralphbean, line 54: Horreum provides a UI, right?

Can you add a section here describing whether or not users are expected to interact with that UI - or if it will be hidden.

(For example, ArgoCD exists behind our Gitops Service APIs, but we conceal it entirely from the user. They cannot access the Argo UI.)

ADR/0018-apps-continuous-perf-testing.md
ralphbean: Yeah - on reading here, it makes sense to me to shard this the same way we shard all other services - one instance per member cluster.

ADR/0018-apps-continuous-perf-testing.md
jhutar: Thanks, added.

ADR/0018-apps-continuous-perf-testing.md
jhutar: I see! Added.

ADR/0018-apps-continuous-perf-testing.md
```diff
@@ -0,0 +1,70 @@
...
+- We can provide some best practices to customers wanting to do some basic performance regression testing.
+- Horreum is end-to-end solution that already exists, is already used by multiple teams and has documentation for the whole process.
+- StoneSoup UI can display all the trends because historical data will be available in Horreum.
+- Red Hat team that develops Horreum is willing to help with this effort.
+- Integration scripts that would be running in the pipeline (to gather monitoring data, to upload results to Horreum, to get PASS/FAIL decision from Horreum...) mostly exists.
+
+Cons:
+
+- Somebody has to manage the Horreum instance. Required development time on it. Service instance needs ownership.
```
ralphbean, line 69: We need to resolve this (somehow) before moving forwards much more. We can't introduce a new subsystem without clear ownership of updating it, accepting updates to it, uptime, etc.

ADR/0018-apps-continuous-perf-testing.md
jhutar: > Who is the We here?

Oh, sorry. That means the pipeline that runs the test and communicates with Horreum.

> Is it some Pipeline definition that can be used commonly across the board?

Some parts will be definitely shared, just parametrized.

> Will there be some service running that will need to be integrated with for performance testing?

AFAICT just that Horreum.

ADR/0018-apps-continuous-perf-testing.md
psuriset: @ralphbean   
Are you asking about production ownership or Development efforts? 
I thought AppSRE will take ownership of maintaining these services.

ADR/0018-apps-continuous-perf-testing.md
johnaohara: There are existing tools/scripts to monitor performance workloads and process the data in Horreum.  The additional work required would be to integrate into Stonesoup;
 - providing the data required for the Stonesoup UI 
 - determine how users configure the change detection parameters for their tests
 - document / provide a step for integrating the change detection into their pipeline (e.g. for Jenkins we have a Horreum plugin that allows users to easily upload their data)

ADR/0018-apps-continuous-perf-testing.md
johnaohara: There are other considerations wrt multiple instances vs a single instance. Horreum has been designed as a multi-tenanted system from the outset, with a comprehensive security model and Row Level Security implemented to prevent leaking of data between tenants. We could provide a "configuration" per workspace that would provide isolation between workspaces.  While it is possible to stand up and instance per workspace, that would require further development work (i.e. operators for provisioning per worksapce, data backup/restore etc)  and management would become more onerous on the SRE team(s)

ADR/0018-apps-continuous-perf-testing.md
arewm: Are there pushed changes or am I looking in the wrong place? I do not see any change in the wording.

ADR/0018-apps-continuous-perf-testing.md
arewm: Instead of indicating that scripts mostly exist, can we just highlight what _does_ exist here and then in the `cons`, highlight what _does not_ exist?

ADR/0018-apps-continuous-perf-testing.md
arewm: Nevermind, I was looking at an old commit. THanks.

ADR/0018-apps-continuous-perf-testing.md
amfred: We are in fact separating Red Hat from other tenants at the cluster level.  Different clusters, different databases.

ADR/0018-apps-continuous-perf-testing.md
amfred: In Stonesoup, if it's an additional PipelineRun, we'd have to surface the ability to run a performance evaluation as a new option in the UI.

If it's just an additional Task within the standard PipelineRun that users are already using, then it wouldn't necessarily require additional UI/UX work. 

However, we do have "time to completion" targets for our standard PipelineRuns, so if the performance test takes 10 minutes or more, it would probably not be feasible to include it in the standard PipelineRun.

I also think these tests will be resource hogs and therefore cost us additional money (for now while the service is free) and later, cost the users additional money (in subscription fees).  We are starting out by charging users for PipelineRun minutes.  I wouldn't be surprised if a performance test costs as much as the rest of the build process in terms of PipelineRun minutes, and therefore if it was run on every build, the cost of Stonesoup could double.

That extra cost argues for making this an optional feature that our customers don't get by default.

Because of those considerations, this feature should come in through the Stonesoup feature request process rather than a technical-only ADR. Is there a prioritized feature on the STONE board for this yet?

amfred (changes requested):

ADR/0018-apps-continuous-perf-testing.md
arewm: In addition to the cost of the PipelineRun, there would be some additional cost of activities within Horreum (I assume) as I was trying to get at in another comment: https://github.com/redhat-appstudio/book/pull/66/files/b6b5d6a4af4b7d7eb583c11bf576e34de1ae1a4f#r1133956096

ADR/0018-apps-continuous-perf-testing.md
jhutar: I see. Right. Added what John wrote above in d244174.

jhutar:
Hello @amfred .

> We are in fact separating Red Hat from other tenants at the cluster level. Different clusters, different databases.

I see, thank you! I think I added note about it to the doc in 104bfa0.

jhutar:
Hello @amfred again (for some reason I can not reply directly to your comment in GH UI, so adding here)

> In Stonesoup, if it's an additional PipelineRun, we'd have to surface the ability to run a performance evaluation as a new option in the UI.
> 
> If it's just an additional Task within the standard PipelineRun that users are already using, then it wouldn't necessarily require additional UI/UX work.
> 
> However, we do have "time to completion" targets for our standard PipelineRuns, so if the performance test takes 10 minutes or more, it would probably not be feasible to include it in the standard enterprise/advanced PipelineRun, and we would definitely not include it in the standard basic PipelineRun (5 minutes or less to completion).

I assume this will be very similar workflow as we plan for functional testing pipelines. It would not be part pipelines bounded by "time to completion" targets.

UI would still be required as we would need a way to allow user to configure Horreum (other way might be to configure the Horreum by code from the pipeline itself) and present some graphs based on Horreum data to explain PASS/FAIL decision to the user.

> I also think these tests will be resource hogs and therefore cost us additional money (for now while the service is free) and later, cost the users additional money (in subscription fees). We are starting out by charging users for PipelineRun minutes. I wouldn't be surprised if a performance test costs as much as the rest of the build process in terms of PipelineRun minutes, and therefore if it was run on every build, the cost of Stonesoup could double.

Yes, I agree perf test cost is substantiation and proper charging of users would be required (possibly based on used CPU and memory in the user namespace).

> That extra cost argues for making this an optional feature that our customers don't get by default.

Yes, it has to be optional because I assume most of the customers do not have any performance test. But having the option for that and documentation on how to do that would provide benefit to customers.

> Because of those considerations, this feature should come in through the Stonesoup feature request process rather than a technical-only ADR. Is there a prioritized feature on the STONE board for this yet?

There is https://issues.redhat.com/browse/STONE-679 - not sure if that is what you are asking for.

ADR/0018-apps-continuous-perf-testing.md
arewm: Thanks for separating into one per tenant cluster!

ADR/0018-apps-continuous-perf-testing.md
arewm: AppSRE won't take ownership of maintaining services. All services will be maintained with a hybrid SRE model with shared responsibility between SRE and the workload teams.

ADR/0018-apps-continuous-perf-testing.md
jhutar: I hope I answered concerns about perf testing pipelines cost in https://github.com/redhat-appstudio/book/pull/66#issuecomment-1472303775 - TL;DR yes, there will be some - similar as for functional tests.

ADR/0018-apps-continuous-perf-testing.md
jhutar: Hello @rvansa. Would you have some estimate on how resource intensive Horreum is please?

ADR/0018-apps-continuous-perf-testing.md
jhutar: Also pinged another Horreum engineer in the other thread to get Horreum resource consumption estimate.

ADR/0018-apps-continuous-perf-testing.md
johnaohara: @jhutar unfortunately @rvansa is no longer supporting Horreum. 

@arewm it is possible to spread the cost of running Horreum across multiple workspaces. wrt an actual figure of the cost to run the service, it depends on the SLA requirements and projected utilization levels. This something we are not aware of atm.

For our internal service, we host multiple teams atm using a single instance, we use 1 machine for the Horreum instance, 1 for postgres database and one for backup. 

We can provide details for resource usage for our current internal instance. We are also currently designing a scaling test to document how the resource requirements scale with load

With this information it should be possible to estimate the cost of running a Horreum service in a cloud provider

arewm (approved):

ADR/0018-apps-continuous-perf-testing.md
```diff
@@ -78,5 +78,6 @@ Cons:
 - Horreum is used by multiple teams without any capacity issues, but Horreum itself was not perf&scale tested formally, so there might be some scaling issues.
 - We would need to develop StoneSoup UI to get graphs/results from Horreum and to allow users to configure the change detection parameters for their tests.
 - If we need one Horreum instance per workspace, that would require further development work (i.e. operator for provisioning per worksapce, data backup/restore etc).
+- As the performance testing pipelines can be some heavy on resources, attribution of costs to users/workspaces might be tricky. Maybe we could add Horreum to scope of <https://issues.redhat.com/browse/STONE-215>.
```
johnaohara, line 81: There are 2 components to attributable cost here I think.  The resources/cost to actually run the tests which could be substantial for a period of time,  and the resource cost of Horreum processing a result.  
- 1: resource cost of running tests: I would have thought this could be accounted for using the same methodology for the build pipelines as it would be part of the pipeline
- 2: resource cost for Horreum: will be dependent on size and frequency of data ingest and processing. This might be tricky to attribute to individual users/piplines, however storage resource usage of historic data would be easier to attribute to users/workspaces

arewm: I agree that 1 would likely fall under "for free"

amfred:
@jhutar I'm fine with running a POC of this on the development versions of Stonesoup, but if the POC is successful we'll have to play catch-up on the compliance, security and SRE work required before moving it into production.  We'll have to allow time for that.  This would be true for any new component, and this particular feature includes a new service, new database and new people supporting it, so it's a new component.   So I'll approve with that caveat.

amfred (approved):

ADR/0018-apps-continuous-perf-testing.md
jhutar: Hope it is OK to resolve this now. Please reopen if there is something to discuss.

ADR/0018-apps-continuous-perf-testing.md
jhutar: Right. Exploded these to two distinct lines: cfc1704

jhutar:
>

jhutar:
Oh my. Obviously I can not use right buttons :-/ (please ignore this close and reopen event if you can and do not laugh at me publicly)

jhutar:
> @jhutar I'm fine with running a POC of this on the development versions of Stonesoup, but if the POC is successful we'll have to play catch-up on the compliance, security and SRE work required before moving it into production. We'll have to allow time for that. This would be true for any new component, and this particular feature includes a new service, new database and new people supporting it, so it's a new component. So I'll approve with that caveat.

Yes, I agree this is a whole new subsystem and will require all the steps you mentioned (and probably more) :-/

johnaohara:
> @jhutar I'm fine with running a POC of this on the development versions of Stonesoup, but if the POC is successful we'll have to play catch-up on the compliance, security and SRE work required before moving it into production. We'll have to allow time for that. This would be true for any new component, and this particular feature includes a new service, new database and new people supporting it, so it's a new component. So I'll approve with that caveat.

It would be great to get more insight into those requirements so we can anticipate how much work is involved

ralphbean (approved):
Thanks for all the work on this.

PR #68, opened 2023-03-15
Propagate quay.io credentials to target clusters

sbose78:
.. for supporting private quay.io images.

resolves https://issues.redhat.com/browse/STONE-706 as part of https://issues.redhat.com/browse/STONE-289


ADR/propagate-pull-secrets.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Overview
+
+This proposal builds on the capabilties of the https://github.com/redhat-appstudio/image-controller which sets up a Quay.io image repository per `Component` and 'downloads' a robot account token into the user's namespace for builds to be able to push to it.
+
+#### 1. Scenario: When a new Environment is defined
+
+1. The image-controller copies the pull(actually, push)`Secret` from the user's namespace into the target cluster's namespace for every Component in the workspace/namespace.
+2. The image-controller links the copied `secret` to the `deployer` `serviceAccount`. If the `deployer` `serviceAccount` is absent, the `Secret` 
+    would be linked to the `default` `serviceAccount` to account for distributions of Kubernetes which do not have a `deployer` `serviceAccount`.
```
arewm, line 25: Is it something that is needed to support BYO clusters?

ADR/propagate-pull-secrets.md
```diff
@@ -0,0 +1,50 @@
...
+Date: 2023-03-16
+
+## Status
+
+Provisional
+
+## Context
+
+* As a user, I would expect StoneSoup to generate a private Quay.io image repository if my source code repository is  private.
+* As a user, I would expect StoneSoup to gracefully deploy images that were pushed to private Quay.io repositories in every relevant environment.
```
arewm, line 12: The expectation is that all Environments configured in a workspace should be able to pull images for components even if those images are private repositories?

ADR/propagate-pull-secrets.md
```diff
@@ -0,0 +1,50 @@
...
+#### 1. Scenario: When a new Environment is defined
+
+1. The image-controller copies the pull(actually, push)`Secret` from the user's namespace into the target cluster's namespace for every Component in the workspace/namespace.
+2. The image-controller links the copied `secret` to the `deployer` `serviceAccount`. If the `deployer` `serviceAccount` is absent, the `Secret` 
+    would be linked to the `default` `serviceAccount` to account for distributions of Kubernetes which do not have a `deployer` `serviceAccount`.
+3. At a heartbeat of 24 hours, the `Secret` in the target cluster's namespace is reconciled.
+
+#### 2. Scenario: When a new Component is created
+
+1. As part of the existing 'daily chores' of the image-controller, a new Quay.io image repository and robot account token is made available in the
```
arewm, line 30: Can you define `daily chores`? Do they run on a cron or does this just describe the normal functionality of the image-controller?

ADR/propagate-pull-secrets.md
```diff
@@ -0,0 +1,50 @@
...
+
+1. Request re-generation of the Quay.io robot account token by setting the annotation `image.redhat.com/generate: 'true'` on the relevant Components.
+  The resulting Quay.io API call automatically revokes the old token.
+2. Continue to *Scenario 2: When a new Component is created*
+
+
+## Consequences
+
+1. The reference to the `Secret` is not visible in the GitOps repo.
+2. The 'disaster recovery' scenario would bank on requesting a re-creation of the image repository credentials, which would then be propagated to the relevant Environments.
```
arewm, line 47: What is the structure of the secret? After disaster recovery, is the secret mapped appropriately so that an update will occur instead of just an addition?

ADR/propagate-pull-secrets.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Decision
+
+### Overview
+
+This proposal builds on the capabilties of the https://github.com/redhat-appstudio/image-controller which sets up a Quay.io image repository per `Component` and 'downloads' a robot account token into the user's namespace for builds to be able to push to it.
+
+#### 1. Scenario: When a new Environment is defined
+
+1. The image-controller copies the pull(actually, push)`Secret` from the user's namespace into the target cluster's namespace for every Component in the workspace/namespace.
```
arewm, line 23: How will Environments use the multitudes of pull secrets for each component?

ADR/propagate-pull-secrets.md
```diff
@@ -0,0 +1,50 @@
...
+2. Continue to *Scenario 2: When a new Component is created*
+
+
+## Consequences
+
+1. The reference to the `Secret` is not visible in the GitOps repo.
+2. The 'disaster recovery' scenario would bank on requesting a re-creation of the image repository credentials, which would then be propagated to the relevant Environments.
+3. This process doesn't yet use a secret storage mechanism like SPI / Vault and the tech debt needs to be factored into future plans for improvement.
+4. The image-controller needs to have permissions to read the credentials associated with an `Environment`.
+5. Since no new pull secrets are created ( we re-use the one created for the builds ), we are potentially planting a `Secret` which also has 'push' permissions even though the deployment of a `pod` only needs to be able to pull an image.
```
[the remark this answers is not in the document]
arewm: +1

ADR/propagate-pull-secrets.md
sbose78: Yes, the content of the `Secret` should change, the shape and name should be the same.

ADR/propagate-pull-secrets.md
```diff
@@ -0,0 +1,52 @@
...
+### Overview
+
+1. This proposal builds on the capabilties of the https://github.com/redhat-appstudio/image-controller which sets up a Quay.io image repository per `Component` and 'downloads' a robot account token into the user's namespace for builds to be able to push to it.
+
+2. This proposal builds on the capabilities of SPI's `RemoteSecret` API. See https://github.com/redhat-appstudio/book/pull/70/files
+
+
+#### When a new `Component` is created
+
+1. A new Quay.io image repository and robot account token is made available in the namespace that would only be able to "pull" the `Component`'s image.
```
sbose78, line 26: > and robot account token is made available in the namespace that would only be able to "pull" the `Component`'s image

The image-controller would need to produce this - will be tracked.

ADR/propagate-pull-secrets.md
sbose78: > all Environments configured 

This should only be for relevant environments, ex, those with SnapShotEnvironmentBindings.

ADR/propagate-pull-secrets.md
ralphbean, line 26: Can you clarify which namespace here?

* The namespace that represents the user's appstudio control plane workspace?
* The namespace represented by the Environment's associated DeploymentTarget?

I think it is the first one - but want to double check. (Maybe update the language to be more specific?)

ralphbean, line 26:
```suggestion
1. image-controller creates a new Quay.io image repository
2. image-controller provisions a new robot account token that would only be able to "pull" the `Component`'s image.
3. image-controller creates a `Secret` in the user's workspace namespace containing the robot account token
```

ADR/propagate-pull-secrets.md
```diff
@@ -0,0 +1,52 @@
...
+#### When an expression of intent to deploy an `Application` to an `Environment` is made.
+
+
+1. Watch for creation of new `SnapshotEnvironmentBindings`.
+2. Query the 'pull' `Secret` generated by the image-controller. 
+3. Label the `Secret` with the name of the `Environment` ( ex, `spi.appstudio.redhat.com/target-environment: prod`) the `Secret` needs to be copied to, based on the `SnapshotEnvironmentBinding` CR.
+3. SPI generates a `RemoteSecret` CR referencing the 'pull' `Secret` above. ( See https://github.com/redhat-appstudio/book/pull/70/files )
+3. SPI reconciles the `RemoteSecret` CR, copies over the 'pull' `Secret` to the relevant cluster/namespace and links it to the `default` `serviceaccount`.
+
+TODO: Which Stonesoup controller should be doing (1), (2) and (3)
```
ralphbean, line 39: My two cents: this should be application-service, since it is already watching for and reacting to changes in the SnapshotEnvironmentBindings in order to write out updates to the gitops repo.

cc @elsony

ADR/propagate-pull-secrets.md
arewm: Is disaster recovery coming back from the deletion of everything or the case where credentials need to be cycled (i.e. if leaked)? If the latter, I don't see any cleanup happening so the number of secrets would just be increasing, some just less valid than others.

ADR/propagate-pull-secrets.md
sbose78: The appstudio control plane namespace, yes (sure, I'll update)

ADR/propagate-pull-secrets.md
```diff
@@ -0,0 +1,54 @@
...
+
+* As a user, I would expect StoneSoup to generate a private Quay.io image repository if my source code repository is  private.
+* As a user, I would expect StoneSoup to gracefully deploy images that were pushed to private Quay.io repositories in every relevant environment.
+
+
+## Decision
+
+### Overview
+
+1. This proposal builds on the capabilties of the https://github.com/redhat-appstudio/image-controller which sets up a Quay.io image repository per `Component` and 'downloads' a robot account token into the user's namespace for builds to be able to push to it.
```
ralphbean, line 19:
```suggestion
1. This proposal builds on the capabilties of the [image-controller](https://redhat-appstudio.github.io/book/book/image-controller.html) which sets up a Quay.io image repository per `Component` and 'downloads' a robot account token into the user's namespace for builds to be able to push to it.
```

ADR/propagate-pull-secrets.md
elsony: How often does the `Secret` changes? If we are only doing that immediately after the creation of `SnapshotEnvironmentBindings`, is there a reason why we don't include that information during the creation of the `SnapshotEnvironmentBindings` so we don't have extra logic to handle this?

ADR/propagate-pull-secrets.md
sbose78: When all the secrets are lost, we'd need to be able to re-pull them from Quay.io. What we'd do is regenerate the robot account token on quay.io and pull it into a secret.

arewm: Ack. This is considering all secrets being lost. So I wouldn't expect any names to change (and therefore any need to perform clean up first). Thanks.

ADR/propagate-pull-secrets.md
ralphbean: @sbose78, wdyt? I think the Secret will "normally" only be provisioned once when the Component is created - although it is possible we'll want to delete it and have it be reprovisioned manually - or, someday introduce automated rotation.

ralphbean (changes requested):
I'm +1 to the content in general, except we need a decision about which component will perform the labelling of the Secret.

Some options to think about:

* HAS could do it, since the labeling happens in response to new SnapshotEnvironmentBindings.
* Image-controller could do it, since it created the secret in the first place and could be considered responsible for its lifecycle.
* Integration-service could do it, since it creates and updates (some, but not all) SnapshotEnvironmentBindings (see [ADR 16](https://redhat-appstudio.github.io/book/ADR/0016-integration-service-promotion-logic.html)). Here, we could say that integration-service must label the secret when it promotes content for the first time. If a user manually promotes a Snapshot to an Environment for the first time then the user (HAC) has to also label the Secret for that Environment.

ADR/propagate-pull-secrets.md
sbose78: @elsony Yeah, that's do-able as well. My proposal mostly intended to address a separation of concerns but doing it along with creation of `SnapshotEnvironmentBindings` works as well.

@ralphbean , Yeah, the Secret shouldn't change a lot. I'm hoping if/when the contents of the `Secret` does change, SPI should take care of getting it sync'd.

sbose78:
I'd prefer an event-driven model where 'something' listens to creation of `SnapshotEnvironmentBindings` and does the needful. We need a reconciler to do this, and it doesn't quite matter which controller it resides it in. 

Let's go for a new reconciler in the image-controller ? @Michkov , what do you think ?

Michkov:
Hot potato :-), image controller is owned by Build team, environment/snapshot/environmentBinding is currently out-of-scope of build team domain. So team is missing knowledge when there will be any issue then it will be harder for us to debug. On the other hand integration team is focused on  environment/snapshot/environmentBinding ;-).

ADR/propagate-pull-secrets.md
```diff
@@ -0,0 +1,54 @@
...
+3. image-controller creates a `Secret` in the user's workspace namespace containing the robot account token
+2. The image-controller labels the `Secret` with the name `Application`/`Component` the `Secret` is associated with. 
+
+
+#### When an expression of intent to deploy an `Application` to an `Environment` is made.
+
+
+1. Watch for creation of new `SnapshotEnvironmentBindings`.
+2. Query the 'pull' `Secret` generated by the image-controller. 
+3. Label the `Secret` with the name of the `Environment` ( ex, `spi.appstudio.redhat.com/target-environment: prod`) the `Secret` needs to be copied to, based on the `SnapshotEnvironmentBinding` CR.
```
dirgim, line 37: If the secret would need to be copied to multiple environments, then I'm not sure how it would work with a single label on the Secret.

ADR/propagate-pull-secrets.md
```diff
@@ -0,0 +1,54 @@
...
+
+1. This proposal builds on the capabilties of the [image-controller](https://redhat-appstudio.github.io/book/book/image-controller.html) which sets up a Quay.io image repository per `Component` and 'downloads' a robot account token into the user's namespace for builds to be able to push to it.
+
+2. This proposal builds on the capabilities of SPI's `RemoteSecret` API. See https://github.com/redhat-appstudio/book/pull/70/files
+
+
+#### When a new `Component` is created
+
+1. image-controller creates a new Quay.io image repository
+2. image-controller provisions a new robot account token that would only be able to "pull" the `Component`'s image.
```
gazarenkov, line 27: Not sure what does robot accout token provisioning mean. I thought it can be only generated on quay.io web page and copy-pasted...Or there is some more automated way? 
Anyway, we can leverage storing this secret data on secret storage using RemoteSecret on this stage by for example:
- create upload secret with the robot accout token and `Application`/`Component` annotations. 
- It's controller will upload token data to the storage and create RemoteSecret with corresponding `Application`/`Component` labels (upload secret will be immediately deleted).

sbose78: 👍

ADR/propagate-pull-secrets.md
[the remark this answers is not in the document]
sbose78: My proposal had started with looking out for `SnapshotEnvironmentBinding` because that shows the intent to deploy - having a Component/Application doesn't quite show that intent :-)  Thoughts?

amfred:
@sbose78 Would you be OK with making one more round of updates based on the comments, and let's see if we can get this merged?

mmorhun:
When https://github.com/redhat-appstudio/image-controller/pull/67 is merged, then both push and pull tokens will be stored in respective RemoteSecrets. Right now, pull token is propagated into all environments, but push token is requested only for current (build) namespace (where build pipeline is started). So, sending push token somewhere is matter of filling right data on the corresponding RemoteSecret.

Another point is that HAC UI currently use `image.redhat.com/generate` annotation to request a new image repository, but that's deprecated method as it's tightly bound to the Component. Instead, it should use new API. Image repository should be requested by creating corresponding [`ImageRepositopry` CR](https://github.com/redhat-appstudio/image-controller#appstudio-component-image-repository). This way also allows to rotate credentials and gives more info.

ralphbean:
I'm taking a look at our [oldest open pull request](https://github.com/konflux-ci/architecture/pulls?q=is%3Apr+is%3Aopen+sort%3Acreated-asc). This one looks like it is now out-moded by https://github.com/konflux-ci/architecture/pull/147. I'm thinking we should close it without merging.

PR #69, opened 2023-03-16
add roles and permissions for DeploymentTarget and DeployementClaim

hongweiliu17:
Add roles and permission for
  * deploymenttargets
  * deploymenttargetclaims

### DIFF
--- a/ADR/0011-roles-and-permissions.md
+++ b/ADR/0011-roles-and-permissions.md
@@ -25,6 +25,7 @@ We will use the built-in Kubernetes RBAC system for Stonesoup's role and permiss
 | Contributor   | Workspace               | Access to namespaces that backs workspace                           |
 |               | Application & Component | appstudio.redhat.com      | get, list, watch                        | applications, components, componentdetectionqueries
 |               | Environment             | appstudio.redhat.com      | get, list, watch                        | promotionruns, snapshotenvironmentbindings, snapshots, environments
+|               | DeploymentTarget & DeploymentTargetClaim | appstudio.redhat.com      | get, list, watch       | deploymentttargets, deploymenttargetclaims
 |               | *GitOps*                | managed-gitops.redhat.com | get, list, watch                        | gitopsdeployments
 |               | PipelineRun             | tekton.dev                | get, list, watch                        | pipelineruns
 |               | Pipeline Results        | results.tekton.dev        | get, list                               | results, records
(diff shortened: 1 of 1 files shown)


ADR/0011-roles-and-permissions.md
```diff
@@ -25,6 +25,7 @@ We will use the built-in Kubernetes RBAC system for Stonesoup's role and permiss
 | Contributor   | Workspace               | Access to namespaces that backs workspace                           |
 |               | Application & Component | appstudio.redhat.com      | get, list, watch                        | applications, components, componentdetectionqueries
 |               | Environment             | appstudio.redhat.com      | get, list, watch                        | promotionruns, snapshotenvironmentbindings, snapshots, environments
+|               | DeploymentTarget       | appstudio.redhat.com      | get, list, watch                        | deploymentttargets, deploymenttargetclaims, deploymenttargetclasses
```
hongweiliu17, line 28: I use `DeploymentTarget` as the value of `Permissions`, should we list the all three CRDs in the `Permissions`?

hongweiliu17:
@gorkem @alexeykazakov @MatousJobanek Could you take a look at this update? Thanks!

ADR/0011-roles-and-permissions.md
```diff
@@ -57,6 +59,7 @@ We will use the built-in Kubernetes RBAC system for Stonesoup's role and permiss
 | Admin         | Workspace               | Access to namespaces that backs workspace                           |
 |               | Application & Component | appstudio.redhat.com      | *                                       | applications, components, componentdetectionqueries
 |               | Environment             | appstudio.redhat.com      | *                                       | promotionruns, snapshotenvironmentbindings, snapshots, environments
+|               | DeploymentTarget       | appstudio.redhat.com      | get, list, watch                        | deploymentttargets, deploymenttargetclaims, deploymenttargetclasses
```
ralphbean, line 62:
```suggestion
|               | DeploymentTarget        | appstudio.redhat.com      | *                        | deploymentttargets, deploymenttargetclaims
|               | DeploymentTargetClass   | appstudio.redhat.com      | get, list, watch                        | deploymenttargetclasses
```

For BYOC (bring your own compute/bring your own cluster), the Admin needs to be able to create an Environment, create a DT, and create a DTC. The admin user already has `*` on Environment. We should set DeploymentTarget and DeploymentTargetClaim to match.

However, the Admin should _not_ be able to create or delete deploymenttargetclasses, which (I think) are cluster scoped.

MatousJobanek: two comments:
1. what you have written above doesn't match what is in the "admin" template now: https://github.com/codeready-toolchain/host-operator/blob/master/deploy/templates/nstemplatetiers/appstudio/spacerole_admin.yaml#L45-L54 But that's probably just a detail and we will align the permissions with this ADR as part of this story: [ASC-300](https://issues.redhat.com/browse/ASC-300) cc @ranakan19
2. if `deploymenttargetclasses` is a cluster-scoped resource, then we cannot use the template to grant the permissions - the template is applied in a namespace-scope only. In that case, the permissions for `deploymenttargetclasses` has to be granted to "all authenticated users" via ClusterRole and ClusterRoleBinding. @hongweiliu17 could you please remove `deploymenttargetclasses` from the template and create the ClusterRole(Binding) either on your side or via infra-deployments?

hongweiliu17: Thanks for the details, it is very clear.

hongweiliu17: I add DeployementTargetClass with addition info (`Cluster-scoped`)as below:
```
| | DeploymentTargetClass (Cluster-scoped) | appstudio.redhat.com | get, list, watch | deploymenttargetclasses
```

alexeykazakov (approved):

ralphbean (approved):

arewm:
Should we continue to update the content of this previous ADR if there are changes that need to be made to the roles and permissions or should we instead create a new representation elsewhere in the book?

This will continue to be a vital part of architectural decisions, i.e. https://github.com/redhat-appstudio/book/pull/27#discussion_r1150590778.

PR #71, opened 2023-03-22
Add a shot blurb on optional label flag in integration tests

jinqi7:


jinqi7:
@dirgim @sonam1412 @jsztuka  @hongweiliu17  Can you please help to review the PR? Thx.

book/integration-service.md
```diff
@@ -137,6 +137,11 @@ Following the [annotation guidelines](https://docs.google.com/document/d/1gyXM3p
 ```
 The Integration service will copy the annotations and labels from the Build PipelineRun and append those to the Test PipelineRuns for traceability across the system per [Labels and Annotations for StoneSoup pipelines](https://docs.google.com/document/d/1fJq4LDakLfcAPvOOoxxZNWJ_cuQ1ew9jfBjWa-fEGLE/edit#) and [StoneSoup builds and tests PRs](https://docs.google.com/document/d/113XTplEWRM63aIzk7WwgLruUBu2O7xVy-Zd_U6yjYr0/edit#)
 
+Below Label provide users an option whether an IntegrationTestScenario will be taken into account when determining if the Snapshot has passed testing. It will be copied to the subsequence Test PipelineRuns.
```
dirgim, line 140: typo: subsequence -> subsequent

dirgim: Also my suggestion is to change `Below Label provide users an option` into `The label below provides users with an option`.

ralphbean: and, I suggest modifying `whether an IntegrationTestScenario will be taken into account when determining if the Snapshot has passed testing.` to `whether the results of an IntegrationTestScenario will be taken into account when determining if the Snapshot has passed all required testing.`

jinqi7: @ralphbean , I changed it  to " whether the result of a pipelineRun created according the IntegrationTestScenario will be taken into account when determining if the Snapshot has passed all required testing" by referring to your comment. Is it OK for you?

dirgim (approved):
LGTM

jsztuka (approved):
lgtm

book/integration-service.md
```diff
@@ -137,6 +137,11 @@ Following the [annotation guidelines](https://docs.google.com/document/d/1gyXM3p
 ```
 The Integration service will copy the annotations and labels from the Build PipelineRun and append those to the Test PipelineRuns for traceability across the system per [Labels and Annotations for StoneSoup pipelines](https://docs.google.com/document/d/1fJq4LDakLfcAPvOOoxxZNWJ_cuQ1ew9jfBjWa-fEGLE/edit#) and [StoneSoup builds and tests PRs](https://docs.google.com/document/d/113XTplEWRM63aIzk7WwgLruUBu2O7xVy-Zd_U6yjYr0/edit#)
 
+The Label below provides users an option whether the result of a pipelineRun created according to the IntegrationTestScenario will be taken into account when determining if the Snapshot has passed all required testing. It will be copied to the subsequent Test PipelineRuns.
```
arewm, line 140: The label is indicated as providing an option to users. What is the default value of the label if it is not configured?

arewm, line 140: I think that I am still confused by this sentence. Maybe wording it something like this will be clearer (and removing the use of option + optional)?

> The `test.appstudio.openshift.io/optional` label can leveraged by the user to affect the determination of whether a Snapshot has passed all required testing. If it is set to `true`, then the result of the PipelineRun created according to the IntegrationTestScenario will need to be passing for the overall Snapshot to pass all tests.

Don't take the words verbatim as I think they are still not accurate, just as a suggestion for how I think we can get clarity.

book/integration-service.md
ralphbean: Yes, thank you!

ralphbean (approved):

book/integration-service.md
jinqi7: If the label is not defined in an IntegrationTestScenario, integration service will consider it as mandatary.

jinqi7: If the label is set to "true", the result of the PipelineRun will not be taken into account when determining if the Snapshot has passed all required testing.

jinqi7: How about the updated description?

arewm: That works, thanks. Can you just comment on the default behavior if the label is absent? Or will this label never be absent in the future because some service is always setting it to the default of `false`?

jinqi7: It's descibed - If the label is not defined in an IntegrationTestScenario, integration service will consider it as "false".

arewm: Many thanks. Sorry for the oversight. Maybe I wasn't fully awake yet!

PR #77, opened 2023-03-29
Revise log conventions ADR

ralphbean:
Based on feedback in slack: https://redhat-internal.slack.com/archives/C02C3SE8QS2/p1680071345179399

See also previous discussion in https://github.com/redhat-appstudio/book/pull/65.

### DIFF
--- a/ADR/0006-log-conventions.md
+++ b/ADR/0006-log-conventions.md
@@ -218,7 +218,7 @@ func f() {
     }
   },
   "additionalProperties": true,
-  "required": [],
+  "required": ["ts", "level", "msg"],
   "not" {
     "required": [
       "namespace_name",
--- a/ADR/0006-log-conventions.md
+++ b/ADR/0006-log-conventions.md
@@ -168,7 +168,7 @@ func f() {
 
 ## Appendix
 
-### JSONSchema for ADDITIONAL_JSON_DATA
+### JSONSchema
 
 ```json
 {
--- a/ADR/0006-log-conventions.md
+++ b/ADR/0006-log-conventions.md
@@ -25,7 +25,7 @@ For example:
 {"ts": "2023-03-07T11:32:29.020Z", "level": "info", "logger": "controllers.Component", "msg": "API Resource updated", "namespace": "samburrai-tenant", "audit": "true", "name": "go-net-http-hello-prqx", "controllerKind": "Component", "action": "UPDATE"}
 ```
 
-The **ts** field stands for "timestamp" (encoded as UTC/ISO-8601) and the meaning of log **level** should be self evident. The **logger** name helps your team understand which part of your service is emitting the log (usually `Log.WithName` from `zap`). The **msg** is a human readable string describing the event being logged. Further **json** key value pairs contain additional fields useful for searching.
+The **ts** field stands for "timestamp" (encoded as UTC/ISO-8601) and the meaning of log **level** should be self evident. The **logger** name helps your team understand which part of your service is emitting the log (usually `Log.WithName` from `zap`). The **msg** is a human readable string describing the event being logged. Further **json** key value pairs contain additional fields useful for searching. See full list of fields in the [Appendix](#Appendix).
 
 The opinionated logging mechanism from the controller-runtime framework already conforms to this scheme. Care should be taken to make sure specific key/value pairs are logged.
 
--- a/ADR/0006-log-conventions.md
+++ b/ADR/0006-log-conventions.md
@@ -156,7 +156,7 @@ func f() {
 * Individual teams should still be able to include key value pairs in their controller logs that are
   not mentioned in this doc, enabling debugging methods unique to that controller.
 * We decided to omit including a `user_id` for situations where one is relevant. This would be
-  helpful for audit purposes (see "Who was involved? above). But, the user id in our system is the
+  helpful for audit purposes (see "Who was involved?" above). But, the user id in our system is the
   same as the username, which constitutes PII - which we do not want to log and forward to other
   systems. If we decide that we do need a `user_id` for audit purposes in the future, then we will
   need to revisit this decision.
--- a/ADR/0006-log-conventions.md
+++ b/ADR/0006-log-conventions.md
@@ -21,8 +21,8 @@ In our controller logs, we will use structured log messages, formatted in JSON w
 For example:
 
 ```
-{"ts": "2023-03-07T11:32:29.167Z", "level": "info", "logger": "controllers.Component", "msg": "updating devfile component name kubernetes-deploy ...", "namespace": "samburrai-tenant", "resource": "go-net-http-hello-prqx", "kind": "Component"}
-{"ts": "2023-03-07T11:32:29.020Z", "level": "info", "logger": "controllers.Component", "msg": "API Resource updated", "namespace": "samburrai-tenant", "audit": "true", "resource": "go-net-http-hello-prqx", "kind": "Component", "action": "UPDATE"}
+{"ts": "2023-03-07T11:32:29.167Z", "level": "info", "logger": "controllers.Component", "msg": "updating devfile component name kubernetes-deploy ...", "namespace": "samburrai-tenant", "name": "go-net-http-hello-prqx", "controllerKind": "Component"}
+{"ts": "2023-03-07T11:32:29.020Z", "level": "info", "logger": "controllers.Component", "msg": "API Resource updated", "namespace": "samburrai-tenant", "audit": "true", "name": "go-net-http-hello-prqx", "controllerKind": "Component", "action": "UPDATE"}
 ```
 
 The **ts** field stands for "timestamp" (encoded as UTC/ISO-8601) and the meaning of log **level** should be self evident. The **logger** name helps your team understand which part of your service is emitting the log (usually `Log.WithName` from `zap`). The **msg** is a human readable string describing the event being logged. Further **json** key value pairs contain additional fields useful for searching.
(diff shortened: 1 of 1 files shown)


ralphbean:
Here, I'm trying to accommodate what I perceive to be an _inconvenience_ imposed by the way the ADR was originally stated.

Is this the right response to the feedback? I'm interested in honest review.

Michkov (approved):

ADR/0006-log-conventions.md
```diff
@@ -60,32 +58,30 @@ produced the log. See section on automatically added logs below.
...
 automatically added logs below.)
 
-This should appear in the `ADDITIONAL_JSON_DATA` at the end of the log line and can optionally
-appear in the human-readable `MESSAGE`.
+Optionally, also refer to the namespace of the *targeted resource* in the human-readable `msg`.
 
 ### 5. Where it came from?
 
-**Included in:** `MESSAGE`, `ADDITIONAL_JSON_DATA`
+**Included in:** `msg`, `kind`, `resource`, and `source
```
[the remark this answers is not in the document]
ralphbean: @skabashnyuk can you review the update in c8bb76c1111368f7fb9eca411e62ce7a59475c7c.

ADR/0006-log-conventions.md
```diff
@@ -11,39 +12,36 @@ Accepted
...
-## Decision: Log conventions
+We initially decided to require log messages formatted by the `console` encoder, thinking that it would be advantageous to be able to consume logs in a human readable format directly from our controller pods (2023-03-08), but on [subsequent review](https://redhat-internal.slack.com/archives/C02C3SE8QS2/p1680074942662149?thread_ts=1680071345.179399&cid=C02C3SE8QS2) we realized that this bears an undesirable cost in configuration of our logging: the default encoder for the "production" profile in zap is the `json` encoder (2023-03-29).
 
-In our controller logs, we will use structured log messages, partially formatted in JSON with key-value pairs as described below.
+## Decision: Log conventions
 
-```
-TIMESTAMP LOG_LEVEL LOGGER MESSAGE ADDITIONAL_JSON_DATA
-```
+In our controller logs, we will use structured log messages, formatted in JSON with key-value pairs as described below.
```
mmorhun, line 19: Log messages in JSON format might be good for production use-case, but for development it's not that great, because the information a developer want to see first gets scattered in the JSON.
In the default logging format (generated by operator-sdk), log messages looks like:
```
2023-02-17T10:41:10Z	ERROR	controllers.ComponentOnboarding	failed to update initial build annotation	{"ComponentOnboarding": "integ-e2e-qplx/integration-suite-test-component-git-source-hwtj", "error": "Operation cannot be fulfilled on components.appstudio.redhat.com \"integration-suite-test-component-git-source-hwtj\": the object has been modified; please apply your changes to the latest version and try again"}
```
so it's easy to find any errors occurred and the short error description is also at the beginning. If a developer needs more details on the error, it's easy to read the line further.

In the proposed JSON format, can we guarantee the fields order, so it's easy to read the log?

ADR/0006-log-conventions.md
```diff
@@ -11,39 +12,36 @@ Accepted
...
 
 For example:
 
 ```
-2023-03-07T11:32:29.167Z    INFO    controllers.Component   updating devfile component name kubernetes-deploy ...   {"namespace": "samburrai-tenant", "resource": "go-net-http-hello-prqx", "kind": "Component"}
-2023-03-07T11:32:29.020Z    INFO    controllers.Component   API Resource updated {"namespace": "samburrai-tenant", "audit": "true", "resource": "go-net-http-hello-prqx", "kind": "Component", "action": "UPDATE"}
+{"ts": "2023-03-07T11:32:29.167Z", "level": "info", "logger": "controllers.Component", "msg": "updating devfile component name kubernetes-deploy ...", "namespace": "samburrai-tenant", "name": "go-net-http-hello-prqx", "controllerKind": "Component"}
+{"ts": "2023-03-07T11:32:29.020Z", "level": "info", "logger": "controllers.Component", "msg": "API Resource updated", "namespace": "samburrai-tenant", "audit": "true", "name": "go-net-http-hello-prqx", "controllerKind": "Component", "action": "UPDATE"}
```
mmorhun, line 25: It would be nice to have an example with all possible fields included

ADR/0006-log-conventions.md
```diff
@@ -11,39 +12,36 @@ Accepted
...
 ```
 
-The meaning of **timestamp** and log level should be self evident. The **logger** name helps your team understand which part of your controller is emitting the log (usually `Log.WithName` from `zap`). The **message** is a human readable string describing the event being logged. The **json data** contains additional fields useful for searching.
+The **ts** field stands for "timestamp" (encoded as UTC/ISO-8601) and the meaning of log **level** should be self evident. The **logger** name helps your team understand which part of your service is emitting the log (usually `Log.WithName` from `zap`). The **msg** is a human readable string describing the event being logged. Further **json** key value pairs contain additional fields useful for searching.
```
mmorhun, line 28: Only a part of allowed fields are described here, which is confusing to me. Should we have all possible fields explained in one pace, so it's easy to find what is what?

ADR/0006-log-conventions.md
```diff
@@ -11,39 +12,36 @@ Accepted
...
 
-Encode timestamps in UTC/ISO-8601 format at the start of the log line.
+Encode timestamps in UTC/ISO-8601 format.
 
 ### 2. What happened?
 
-**Included in:** `MESSAGE`, `ADDITIONAL_JSON_DATA`
+**Included in:** `msg` and `action`
 
 Use the key `action` with possible values `VIEW`, `ADD`, `UPDATE`, and `DELETE`, if applicable.
```
mmorhun, line 42: Why does `CREATE` not included?

ADR/0006-log-conventions.md
```diff
@@ -60,32 +58,31 @@ produced the log. See section on automatically added logs below.
...
 Optionally, use the key `source` to direct developers to the source code where the action occurred.
 
-Include these in the `ADDITIONAL_JSON_DATA` at the end of the log line. They can optionally appear
-in the human-readable `MESSAGE`.
+Optionally, refer to these in the human-readable `msg`.
 
 ### Controller audit logs
 
-**Included in:** `ADDITIONAL_JSON_DATA`
+**Included in:** `audit`
```
mmorhun, line 85: Would be nice to have a sentence that explains what _audit_ log entry mean.

ADR/0006-log-conventions.md
```diff
@@ -137,7 +134,7 @@ want to for the goals of this ADR.
   logging mechanism already conforms to the prescribed format, and can be used, it should be
   _easier_ for teams to conform to this ADR than would be otherwise.
 
-  For example, using the standard controller-runtime logging mechanism will produce a text string in
+  For example, using the standard controller-runtime logging mechanism will produce a json string in
```
mmorhun, line 137: To avoid (slightly) different logs configuration between operators, it'd be great to have a common configuration of the logger that could be directly copied into each operator.

ADR/0006-log-conventions.md
Michkov: Maybe we can use standard CRUD (Create, Read, Update, Delete)? Or kubernetes verbs (delete, create, get, update/patch). Is it `VIEW`, `ADD`, `UPDATE`, and `DELETE` some standard, it slightly invokes SQL (probably due to upper case) but there is INSERT  and SELECT.

ADR/0006-log-conventions.md
mmorhun: Agree, I just was asking to document default approach for all controllers

ADR/0006-log-conventions.md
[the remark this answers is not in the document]
ralphbean: Done in 6d518ca.

ADR/0006-log-conventions.md
ralphbean: No, I think the `VIEW`, `ADD`, `UPDATE`, and `DELETE` are not based on some standard. We have the freedom to change it here, so long as we agree and settle on something.

Any strong opinions on this? If we're going to change it, let's change it now and let this settle.

ADR/0006-log-conventions.md
```diff
@@ -150,15 +149,14 @@ func f() {
 }
 ```
 
-* By using structured logs that are only partially formatted as JSON, we should strike a balance
-  between easy readability and support for centralized queries. It should be easy for a human to
-  read pod logs of the controller directly since the first portion of each line contains a human
-  readable string early on, while the JSON formatted suffix supports the creation of queries that
-  span subsystems in centralized aggregated logs.
+* By using structured logs that are formatted entirely as JSON, we lose "easy readability" of
```
kpiwko, line 152: Is the loss of readability a big deal? There's many tools that people can pipe JSON formatted logs to if needed.

ralphbean: No, it's not a big deal.

ADR/0006-log-conventions.md
ralphbean: That can be found down in the `# Appendix`.

mmorhun: Can we add something like "See full list of fields in Appendix"?

ADR/0006-log-conventions.md
mmorhun: Well, I had use-case where creation of an object failed. So, I had to log the error with action CREATE which is not listed here.

ADR/0006-log-conventions.md
ralphbean: Yes, good idea. Will do.

ADR/0006-log-conventions.md
ralphbean: Would `ADD` not be acceptable in that situation?

ADR/0006-log-conventions.md
mmorhun: I see, thanks

ADR/0006-log-conventions.md
mmorhun: Thanks, we'll take into account

mmorhun (approved):

PR #79, opened 2023-03-29
Rename Stonesoup back to AppStudio

arewm:
Also include the renaming of diagrams


arewm:
I renamed all instances that I could find except for this one as I wasn't sure if it should be renamed as well:

https://github.com/redhat-appstudio/book/blob/52d1eed5941f55c5232e691dcc7d4412a1042912/ADR/0010-namespace-metadata.md?plain=1#L37

ADR/assets/0018-apps-continuous-perf-testing.svg
ralphbean, line 1: It looks like the svg didn't actually get updated here?

ralphbean (approved):

ADR/assets/0018-apps-continuous-perf-testing.svg
arewm: Yeah. I tried to do it locally but Inkscape seems to crash on my computer whenever I attempt to modify it.

Michkov:
On 'RHTAP - Architecture Deep Dive' document there is AppStudio used 151x and App Studio used 134x

arewm:
In https://github.com/redhat-appstudio/book/pull/34, it was AppStudio so that is the pattern that I followed to change it back to.

alexeykazakov (approved):

PR #84, opened 2023-04-05
SPI CR permission refinement

skabashnyuk:
That is a refinement of SPI CR permission. 
That is an internal object that is used to communicate between OAuth service and the operator. 
It is created on behalf of the user. But they never directly interact with it.
In the future, we are planning to get rid of it and use a different technique. 
More about this object https://issues.redhat.com/browse/SVPI-303.


sbose78 (approved):
/lgtm

alexeykazakov (approved):

PR #88, opened 2023-04-11
fix file name for workload secret mgmt

gorkem:


arewm (changes requested):
I am trying to consume 21 here: https://github.com/redhat-appstudio/book/pull/86/commits/4c34db0f0bb1af30dbcc775f0707d3f23dba7620

... but am not getting approvals for it. :)

arewm (approved):

PR #89, opened 2023-04-11
chore(HACBS-1490): Add book for internal services

scoheb:
- Add new entry in the book for internal services


book/internal-services.md
```diff
@@ -0,0 +1,83 @@
+# Internal Services
+
+## Overview
```
arewm, line 3: Nit: Is capitalization in this section because they are "keywords"? It feels a little odd since they are not necessarily proper nouns.

book/internal-services.md
```diff
@@ -0,0 +1,83 @@
...
+### WATCH
+
+| Custom Resource            | When?                                                      | Why?                                                                               |
+|----------------------------|------------------------------------------------------------|------------------------------------------------------------------------------------|
+| InternalRequest            | Always                                                     | To provide an API to process an internal request                                   |
+| PipelineRun                | Once the PipelineRun is created                            | To relay the Internal PipelineRun status to the remote InternalRequest for viewing |
+
+## Security Risk Mitigations
+
+Enabling remote clusters the ability to run internal jobs carrying certain security risks. The following list describes measures in places to mitigate those risks.
```
arewm, line 57: Is the list below just a set of mitigations for this one singular security risk? If so, can we be a little more descriptive about the risk? I don't have a lot to add off hand other than enabling malicious actors to pivot into private networks and being able to achieve arbitrary code execution.

book/internal-services.md
```diff
@@ -0,0 +1,83 @@
...
+
+* The creation of an `InternalRequest` custom resource requires permission on the Remote Cluster.
+* The Internal Services controller instance is configured to watch a specific cluster.
+  * This cluster is provided to the controller as an argument to a secret that was added by the admin team.
+  * The secret contains a KUBECONFIG file.
+* Only pipelines that are defined within the controller's namespace can be executed on the Internal, Private cluster.
+  * The deployment of these pipelines is controlled by App Interface and is orchestrated via GitHub pull requests.
+  * In addition, pipelines deployed to production are pinned to Git commit SHAs described in App Interface.
```
arewm, line 65: These are important for "our" use of the controller but they feel like implementation details that are more relevant for a threat model than a general overview. 

I feel like a more appropriate further refinement of this might be along the lines of tightly limiting the actions of the pipeline to reduce any potential harm due to arbitrary code execution.

book/internal-services.md
```diff
@@ -0,0 +1,83 @@
...
+3. The Internal Services controller verifies whether the remote namespace is allowed in the `Config` CR.
+4. If it is not allowed, the Internal Services controller updates the `InternalRequest` CR `status` field accordingly and stops reconciling.
+5. If it is allowed, a `PipelineRun` is created based on the `spec.request` and `spec.params` from the `InternalRequest` CR.
+6. The `PipelineRun` is watched by the Internal Services controller.
+7. Upon completion, the Internal Services controller updates the `status` section of the `InternalRequest` CR
+8. The remote cluster calling pipeline sees the update to the `status` of the `InternalRequest` CR and continues its execution.
+
+### References
+
+* App Interface - https://gitlab.cee.redhat.com/service/app-interface
```
arewm, line 83: I think this is an implementation detail and isn't needed.

diagrams/internal-services/internal-services-controller-overview.jpg
arewm, line 1: This uses the public/private barrier whereas elsewhere remote is used. It would be okay to use public here as long as you introduce it as being "where" the remote cluster is within the document.

book/internal-services.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Detailed Workflow
+
+1. A `InternalRequest` CR is created by a pipeline run by a service account on the remote cluster as part of a pipeline.
+    * The `spec.request` should reference the pipeline to execute.
+    * The `spec.params` should reference the parameters to pass to the pipeline.
+2. The `InternalRequest` CR is noticed by the Internal Services controller and attempts to reconcile it.
+3. The Internal Services controller verifies whether the remote namespace is allowed in the `Config` CR.
+4. If it is not allowed, the Internal Services controller updates the `InternalRequest` CR `status` field accordingly and stops reconciling.
+5. If it is allowed, a `PipelineRun` is created based on the `spec.request` and `spec.params` from the `InternalRequest` CR.
```
arewm, line 76: Does the pipeline definition come from the `InternalRequest` CR or is the CR only used to pass inputs to a pre-defined pipeline definition?

arewm (commented):
Overall, I think it looks really good. I think we can just bolster up the motivation for this, specifically focusing on the security concerns which initiated this approach.

book/internal-services.md
```diff
@@ -0,0 +1,83 @@
...
+The diagram below shows the interaction of the internal services controller and other services and shows the flow of custom resources
+
+![](../diagrams/internal-services/internal-services-controller-overview.jpg)
+
+## Terminology
+
+**InternalRequest** - The custom resource that describes the internal service to trigger the internal job on.
+**Remote Cluster** - A Red Hat Trusted Application member cluster that resides outside of Red Hat's network.
+**Internal, Private Cluster** - A cluster that is not externally addressable but which has access to Red Hat's network.
+**App Interface** - A repository that serves as a central coordination point for hosted services operated by the Application SRE team.
```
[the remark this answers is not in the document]
scoheb: Fixed

book/internal-services.md
```diff
@@ -0,0 +1,83 @@
...
+* The Internal Services controller only acts on remote namespaces that are specifically allowed in the `Config` custom resource.
+
+## Detailed Workflow
+
+1. A `InternalRequest` CR is created by a pipeline run by a service account on the remote cluster as part of a pipeline.
+    * The `spec.request` should reference the pipeline to execute.
+    * The `spec.params` should reference the parameters to pass to the pipeline.
+2. The `InternalRequest` CR is noticed by the Internal Services controller and attempts to reconcile it.
+3. The Internal Services controller verifies whether the remote namespace is allowed in the `Config` CR.
+4. If it is not allowed, the Internal Services controller updates the `InternalRequest` CR `status` field accordingly and stops reconciling.
```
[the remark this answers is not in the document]
scoheb: updated

book/internal-services.md
scoheb: updated

book/internal-services.md
scoheb: removed

book/internal-services.md
scoheb: Fixed

book/internal-services.md
```diff
@@ -0,0 +1,83 @@
...
+## System Context
+
+The diagram below shows the interaction of the internal services controller and other services and shows the flow of custom resources
+
+![](../diagrams/internal-services/internal-services-controller-overview.jpg)
+
+## Terminology
+
+**InternalRequest** - The custom resource that describes the internal service to trigger the internal job on.
+**Remote Cluster** - A Red Hat Trusted Application member cluster that resides outside of Red Hat's network.
```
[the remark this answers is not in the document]
scoheb: Fixed

book/internal-services.md
scoheb: saved those bullets for the Threat Model.

scoheb:
@arewm I addressed your comments. thanks!

book/internal-services.md
```diff
@@ -0,0 +1,81 @@
...
+The following list describes measures in places to mitigate those risks.
+
+* The creation of an `InternalRequest` custom resource requires permission on the Remote Cluster.
+* The Internal Services controller instance is configured to watch a specific cluster.
+  * This cluster is provided to the controller as an argument to a secret that was added by the admin team.
+  * The secret contains a KUBECONFIG file.
+* Only `Pipelines` that are defined and exist within the controller's namespace can be executed on the internal, private cluster.
+* The Internal Services controller only watches and acts on remote namespaces that are specifically allowed in the `Config` custom resource.
+
+## Detailed Workflow
```
arewm, line 70: Is there a way that we can easily indicate where the actions are happening -- whether on the internal or remote clusters? I almost feel like swimlanes or a table would be better here.

book/internal-services.md
```diff
@@ -0,0 +1,81 @@
...
+
+![](../diagrams/internal-services/internal-services-controller-overview.jpg)
+
+## Terminology
+
+**InternalRequest** - The custom resource that describes the internal service to trigger the internal job on.
+**Remote Cluster** - An AppStudio cluster residing outside a private network.
+**Internal, Private Cluster** - A cluster that is not externally addressable but which has access to a private network.
+
+## Resources
```
arewm, line 26: Are the CRs short enough that we can include sample snippets?

book/internal-services.md
```diff
@@ -0,0 +1,81 @@
...
+The results and outcome of the pipeline are added as an update to the custom resources. The remote cluster watches these resources to determine the outcome.
+
+## Goals
+
+* The Internal Services system attempts to enable execution of internal jobs via a polling mechanism.
+* The Internal Services system provides execution results back to the requesting cluster.
+
+## System Context
+
+The diagram below shows the interaction of the internal services controller and other services and shows the flow of custom resources
```
arewm, line 16: Do we want to call out any of the other services by name and link to them (i.e. release service)? Or they are not _really_ a dependency so the link doesn't need to be made (in either direction)?

scoheb: mentioned Release Service

book/internal-services.md
scoheb: added

book/internal-services.md
scoheb: the system diagram shows this in detail. Is that diagram missing anything?

book/internal-services.md
```diff
@@ -0,0 +1,81 @@
+# Internal Services
+
+## Overview
+
+The Internal Services system consists of a kubernetes controller (running on an internal, private cluster) that is capable of watching and reconciling custom resources on a remote cluster.
+These custom resources describe which pipelines and parameters to use to execute internal jobs on an internal, private cluster.
+The results and outcome of the pipeline are added as an update to the custom resources. The remote cluster watches these resources to determine the outcome.
+
```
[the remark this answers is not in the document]
scoheb: do you mean we should provide reasons why we chose this design? Other system entries in the book do not attempt to address this.

Would you have any suggestions? 

I can only think of the pattern of the `Bastion Cluster` that was initially discussed but trashed due to SRE and ProdSec concerns.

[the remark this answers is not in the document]
arewm: We have this covered already in the ADR: https://github.com/redhat-appstudio/book/blob/main/ADR/0003-interacting-with-internal-services.md :)

book/internal-services.md
arewm: The diagram is helpful when looking at it. The issue is likely the view I have where the diagram is separated more. Do the numbers here correspond to the numbers in the diagram? If so, indicating that would help.

scoheb: updated

diagrams/internal-services/internal-services-controller-overview.jpg
scoheb: I amended the text to specify `remote, public clusters`

ralphbean (approved):

PR #91, opened 2023-04-13
doc(STONEINTG-371): Document behavior for components with no image

14rcole:
There is a race condition in the integration service when two component builds are triggered in close succession with one another.  Components whose builds have not yet finished are sometimes added to the ApplicationSnapshot tested by the integration service.  This can lead to undefined failures in the service.  This was resolved by checking each component being added to the snapshot and omitting components with no image.

This documentation change clarifies the behavior described above.


14rcole:
I'm unable to add a label, but I believe this is not a significant change and requires the `skip-adr-check` label.

dirgim (approved):
LGTM

14rcole:
@sonam1412 PTAL when you have the chance

arewm (approved):

PR #93, opened 2023-04-17
chore(STONE-109): ADR for Release Attribution

scoheb:


ralphbean (approved):

ADR/0023-release-and-releaseplan-authorization.md
```diff
@@ -0,0 +1,55 @@
...
+- `Release`: A CRD that describes what should be released and using which plan.
+- `ReleasePlan`: A CRD that describes where and how an application should be released.
+- `ReleaseStrategy`: A CRD that describes which pipeline to execute along with specific pipeline parameters.
+
+#### Architecture
+
+![](../diagrams/ADR-0023/ReleaseAndReleasePlanAuthorization.jpg)
+
+##### Capturing authorization data
+* The idea is that when a release pipeline executes, the `Pipeline`'s `Tasks` have access to the name of the associate that authorized the change in production.
```
arewm, line 28: What is this idea? Are you trying to defer how the release pipeline has access to the authorized associate as an implementation detail? If so, I think we can make that clearer, maybe as a precondition before the bullets start.

ADR/0023-release-and-releaseplan-authorization.md
```diff
@@ -0,0 +1,55 @@
...
+* `Authorization` for a particular `Release` should be obtained by capturing the user that has created the object.
```
arewm, line 29: We can change the hypothetical language like "should" to assertions of what will happen.

ADR/0023-release-and-releaseplan-authorization.md
```diff
@@ -0,0 +1,55 @@
...
+
+Date: 2023-04-17
+
+## Status
+
+Accepted
+
+## Context
+
+* It is imperative to know which associate has authorized a change to production in order to have accountability for changes.
```
arewm, line 11: The term associate seems a little corporate. We can change this to an individual.

Are we limiting to a single individual, or is it possible to have multiple individuals?

ADR/0023-release-and-releaseplan-authorization.md
```diff
@@ -0,0 +1,55 @@
...
+* The idea is that when a release pipeline executes, the `Pipeline`'s `Tasks` have access to the name of the associate that authorized the change in production.
+* `Authorization` for a particular `Release` should be obtained by capturing the user that has created the object.
+  * The `status` section of the `Release` CR should be updated with to specify the `author`.
+* A `Standing Authorization` for an application should be obtained by capturing the user that creates or updates the `ReleasePlan` CR **when** `annotations.standing-authorization` is set to `true`
+  * Since we expect ReleasePlans to be updated by users who may not intend to provide a `Standing Authorization` (i.e. Change `ReleaseStrategy`), it is important to intercept only when the above annotation is set.
+  * The `status` section of the `ReleasePlan` CR should be updated with the `author`.
+
+##### Processing a Release
+* When a `Release` is processed, the Release service must follow this process:
+  * Verify that the `status.author` is populated.
```
arewm, line 37: Is it possible to prevent this from being set by the user themselves? If not, this could be tampered with to claim sign-off by a different author.

ADR/0023-release-and-releaseplan-authorization.md
```diff
@@ -0,0 +1,55 @@
...
+#### Architecture
+
+![](../diagrams/ADR-0023/ReleaseAndReleasePlanAuthorization.jpg)
+
+##### Capturing authorization data
+* The idea is that when a release pipeline executes, the `Pipeline`'s `Tasks` have access to the name of the associate that authorized the change in production.
+* `Authorization` for a particular `Release` should be obtained by capturing the user that has created the object.
+  * The `status` section of the `Release` CR should be updated with to specify the `author`.
+* A `Standing Authorization` for an application should be obtained by capturing the user that creates or updates the `ReleasePlan` CR **when** `annotations.standing-authorization` is set to `true`
+  * Since we expect ReleasePlans to be updated by users who may not intend to provide a `Standing Authorization` (i.e. Change `ReleaseStrategy`), it is important to intercept only when the above annotation is set.
```
arewm, line 32: What would happen if a standing authorization was configured for one individual and then another individual changed it. Would the standing authorization be changed to the new individual?

Instead of using standing-authorization as a boolean, I think it might make more sense for it to be a list of users. Therefore, if any one of those users modifies the `ReleasePlan` then their authorization will be given. If any other user modifies the ReleasePlan, then authorization will need to be revoked. 

I don't know the details of the release process, but should we actually block all releases if a ReleasePlan is modified by an individual that doesn't have standing authorization until authorization has been given? This would be to prevent the ReleasePlan from being unintentionally (or intentionally) changed to an undesired state and not detecting the unintended state until too late.

ADR/0023-release-and-releaseplan-authorization.md
```diff
@@ -0,0 +1,55 @@
...
+  * The `status` section of the `ReleasePlan` CR should be updated with the `author`.
+
+##### Processing a Release
+* When a `Release` is processed, the Release service must follow this process:
+  * Verify that the `status.author` is populated.
+  * If empty, consult the `ReleasePlan` to look if `status.author` is present and populated.
+  * If `status.author` of `ReleasePlan` is empty, fail with `ValidationError`.
+  * If found, copy `ReleasePlan` `status.author` to `Release` `status.author`.
+  * Verify if author is a "real" `User` as opposed to a `ServiceAccount`.
+    * The system should be able to validate and obtain user info from the SSO service.
```
arewm, line 42: What is being validated? The fact that the user is an actual user, that the user is the same one that requested the action, some other metadata about the user to ensure that they have proper authorization, or something else?

arewm (changes requested):

ADR/0023-release-and-releaseplan-authorization.md
scoheb: The status section of the CR can only be changed by the controller

ADR/0023-release-and-releaseplan-authorization.md
```diff
@@ -0,0 +1,55 @@
...
+  * Update `status.isAuthorVerified` if verification passed.
+  * Otherwise, fail with a `ValidationError`.
+
+* At this point, the `status.author` is non-empty and has been validated to be a real `User`.
+  * The Release Service makes the `status.author` value available to the Release pipelines.
+
+##### Removing authorization data
+* A `Standing Authorization` for an application can be deleted by capturing the user that updates the `ReleasePlan` CR **when** `annotations.standing-authorization` is set to `false`
```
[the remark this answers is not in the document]
scoheb: sentence fixed.

ADR/0023-release-and-releaseplan-authorization.md
```diff
@@ -0,0 +1,55 @@
...
+* A `Standing Authorization` for an application should be obtained by capturing the user that creates or updates the `ReleasePlan` CR **when** `annotations.standing-authorization` is set to `true`
+  * Since we expect ReleasePlans to be updated by users who may not intend to provide a `Standing Authorization` (i.e. Change `ReleaseStrategy`), it is important to intercept only when the above annotation is set.
+  * The `status` section of the `ReleasePlan` CR should be updated with the `author`.
+
+##### Processing a Release
+* When a `Release` is processed, the Release service must follow this process:
+  * Verify that the `status.author` is populated.
+  * If empty, consult the `ReleasePlan` to look if `status.author` is present and populated.
+  * If `status.author` of `ReleasePlan` is empty, fail with `ValidationError`.
+  * If found, copy `ReleasePlan` `status.author` to `Release` `status.author`.
```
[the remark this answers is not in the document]
scoheb: fixed indentation.

ADR/0023-release-and-releaseplan-authorization.md
scoheb: updated. we simply want to know if this author field is a known user in SSO. This rules out **kube:admin** and any other **ServiceAccount**. It will also help weed out the cases where the user has left the company.

ADR/0023-release-and-releaseplan-authorization.md
```diff
@@ -0,0 +1,55 @@
+# Release and ReleasePlan Authorization
```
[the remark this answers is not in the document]
scoheb: Yes, we are focused on **Ensuring that we can attribute the source of a release action**

We can change the term `authorization` for something else...but this term is mentioned in the original JIRA Feature.

wdyt @ralphbean @brianwcook ?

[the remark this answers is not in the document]
scoheb: done

ADR/0023-release-and-releaseplan-authorization.md
scoheb: > What would happen if a standing authorization was configured for one individual and then another individual changed it. Would the standing authorization be changed to the new individual?

Yes, it would be changed.

> Instead of using standing-authorization as a boolean, I think it might make more sense for it to be a list of users. Therefore, if any one of those users modifies the ReleasePlan then their authorization will be given. If any other user modifies the ReleasePlan, then authorization will need to be revoked.

First, a change to the standing-authorization must be done by changing a specific annotation. By doing so, we capture that user and record it. That becomes the new  standing-authorization.

> I don't know the details of the release process, but should we actually block all releases if a ReleasePlan is modified by an individual that doesn't have standing authorization until authorization has been given? This would be to prevent the ReleasePlan from being unintentionally (or intentionally) changed to an undesired state and not detecting the unintended state until too late.

Yes, the ADR mentions that we will mark the Release as failed (`ValidationError`) if we do not have a valid author (either from the Release or the ReleasePlan CRs). Because of this, no pipeline would even be triggered.

Therefore, with no authorization (singular or standing), releases are blocked.

ADR/0023-release-and-releaseplan-authorization.md
scoheb: updated.

ADR/0023-release-and-releaseplan-authorization.md
scoheb: done

ADR/0023-release-and-releaseplan-authorization.md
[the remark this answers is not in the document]
arewm: I think there still needs to be a little indentation change. 39 needs to be indented from 38 and 40 needs to be in line with 38?

ADR/0023-release-and-releaseplan-authorization.md
```diff
@@ -0,0 +1,57 @@
...
+  * The `status` section of the `ReleasePlan` CR will be updated with the `author`.
+
+##### Processing a Release
+* When a `Release` is processed, the Release service must follow this process:
+  * Verify that the `status.author` is populated.
+    * If empty, consult the `ReleasePlan` to look if `status.author` is present and populated.
+    * If `status.author` of `ReleasePlan` is empty, fail with `ValidationError`.
+  * If found, copy `ReleasePlan` `status.author` to `Release` `status.author`.
+  * Verify if author is a "real" `User` as opposed to a `ServiceAccount`.
+    * Given an author value, the system should be able to query the SSO service to determine that the user is a known user in SSO.
```
arewm, line 42: Is this possible today or is this something that needs to be figured out still? I am just trying to figure out what the `should` means here. :)

ADR/0023-release-and-releaseplan-authorization.md
arewm: I will rephrase my concern with a potential solution. Should anyone with a role of `contributor` or higher be "trusted" enough to sign off on a release? 

If we might consider that anyone with these roles would not be trusted, then we shouldn't enable the `standing-authorization` to be updated to just anyone and we need to ensure that there are additional checks in place. The potential checks as I mentioned them would have no effect, however, because if someone can modify a release then they can modify a releaseplan. Therefore, the behavior of `standing-authorization` as mentioned should be fine. If any checks are needed, there should be additional parameters on the releaseplanadmission. This falls into the other conversation on this ADR and does not need to be re-hashed here.

I think I have clarified my concerns, so feel free to resolve this conversation if you think we are on the same page.

scoheb: we are on the same page.

ADR/0023-release-and-releaseplan-authorization.md
scoheb: changed wording

ADR/0023-release-and-releaseplan-authorization.md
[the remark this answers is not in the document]
scoheb: I see. fixed again.

arewm:
Looks good, I can approve once the remaining conversation is resolved.

ADR/0023-release-and-releaseplan-authorization.md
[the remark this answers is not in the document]
ralphbean: On second read, yeah - let's go with attribution. In the future we may have additional requests on the process that layer on the second part of the problem @arewm described. This really is only recording whodunnit which maps most closely to "attribution".

arewm (approved):
Looks great, thanks!

PR #95, opened 2023-04-19
doc(STONEINTG-202): Create a JSON schema for the standardized test ou…

Josh-Everett:
Update details on how test output is validated.

### DIFF
--- a/ADR/0013-integration-service-api-contracts.md
+++ b/ADR/0013-integration-service-api-contracts.md
@@ -104,6 +104,47 @@ Example for a run with an error:
 ```
 
 
+#### Tekton Result Schema Validation
+The test output of the Tekton result **HACBS_TEST_OUTPUT** will be validated using the [jsonschema validator package](https://github.com/santhosh-tekuri/jsonschema).
+The schema is configured as follows:
+
+```
+{
+  "$schema": "http://json-schema.org/draft/2020-12/schema#",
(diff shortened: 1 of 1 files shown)


ADR/0013-integration-service-api-contracts.md
```diff
@@ -104,6 +104,48 @@ Example for a run with an error:
...
+  "$schema": "http://json-schema.org/draft/2020-12/schema#",
+  "type": "object",
+  "properties": {
+    "result": {
+      "type": "string",
+      "enum": ["SUCCESS", "FAILURE", "WARNING", "SKIPPED", "ERROR"]
+    },
+    "namespace": {
+      "type": "string"
+    },
```
arewm, line 122: I assume that default values cannot be specified since they are not relevant for schema verifications, but if so, we can specify the default value here.

arewm (approved):

dirgim (approved):
LGTM

Josh-Everett:
Recent push removes line restricting maxLength to reflect new code.

Josh-Everett:
Also I lack write access but I think otherwise we are good to go.

ADR/0013-integration-service-api-contracts.md
dirgim: That's correct, but it's mentioned in the other part of this document

PR #98, opened 2023-04-24
ADR 22 Secret Management For User Workloads update

skabashnyuk:
This PR addresses recent [discussions](https://redhat-internal.slack.com/archives/C02CTEB3MMF/p1681914904953249) , and keeps it updated with the current development process. Also related to https://github.com/redhat-appstudio/book/pull/68

New problems/cases
* It should be possible to distinguish in time the process of uploading confidential information and deploying it in the destination environment. For example, image-controller creates an image pull secret after `Component` is created and deploys it after `Environment` is made.
* It should be possible to set multiple destination targets.
* The UI should be able to search existing RemoteSecret by target environment/component/application.

### DIFF
--- a/ADR/0022-secret-mgmt-for-user-workloads.md
+++ b/ADR/0022-secret-mgmt-for-user-workloads.md
@@ -74,9 +74,9 @@ metadata:
     name: upload-secret-data-for-remote-secret
     namespace: default
     labels:
-        spi.appstudio.redhat.com/upload-secret: remotesecret
+        appstudio.redhat.com/upload-secret: remotesecret
     annotations:
-        spi.appstudio.redhat.com/remotesecret-name: test-remote-secret
+        appstudio.redhat.com/remotesecret-name: test-remote-secret
 type: Opaque
 stringData:
     username: u


ADR/0022-secret-mgmt-for-user-workloads.md
```diff
@@ -26,65 +29,140 @@ Accepted
...
+The proposed solution is to create a new Kubernetes Custom Resource (CR) called `RemoteSecret`. It serves as a representation of the Kubernetes Secret that is stored in permanent storage, which is also referred to as `SecretStorage`. This Custom Resource includes references to targets, like Kubernetes namespaces, that may also contain the required data to connect to a remote Kubernetes. To perform an upload to permanent storage, a temporary **Upload Secret** is utilized, which is represented as a regular Kubernetes Secret with special labels and annotations that the SPI controller recognizes. Different `SecretStorage` implementations, like AWS Secret Manager or HashiCorp Vault, can be used. It is recommended to create the RemoteSecret first and then use the linked Upload Secret to upload secret data. However, in simple cases, the Upload Secret can be used to perform both uploading and the creation of RemoteSecret in a single action.
```
mmorhun, line 34: Could you elaborate more on why it's recommended to create the RemoteSecret first and then use the linked Upload Secret?

ADR/0022-secret-mgmt-for-user-workloads.md
```diff
@@ -10,6 +10,9 @@ Accepted
 
 * When user workloads are deployed to environments, the system should be able to provide a way to inject values that are specific to the environment. Currently, this is done through environment variables that are managed as overlays on the GitOps repository for the application. However, this method does not provide a good way to manage `Secret`. This ADR addresses the secret management of user workloads for different environments.
 * As a StoneSoup component, I expect to have the ability to securely upload a secret, associate it with the component, and propagate it to the target environment.
+* It should be possible to distinguish in time the process of uploading confidential information and deploying it in the destination environment. For example, image-controller creates an image pull secret after `Component` is created, and deploys it after `Environment` is made.
```
mmorhun, line 13: I'd like to have example in this item clarified.
I completely agree that credentials upload and their further provision could be separated in time and should be supported. However, the task of Image Controller is to provision image repository and upload its credentials to SPI. So, in general, selecting target environments should not be done by Image Controller.

ADR/0022-secret-mgmt-for-user-workloads.md
```diff
@@ -26,65 +29,140 @@ Accepted
...
+        appstudio.redhat.com/remotesecret-target-application: coffee-shop
+spec:
+    target:
+        - namespace: jdoe-tenant
+          secretName: test-remote-secret-secret-5bmq9
+status:
+    TBD
+```
+
+#### Example: If RemoteSecret has to be created all Environments of certain component and application
```
mmorhun, line 118: If we are keeping Image Controller in context, then, probably, the credentials to access Component image repository should be propagated into all environments (even ones created after image repository provision). Therefore, this use-case should be primary for Image Controller.

ADR/0022-secret-mgmt-for-user-workloads.md
mmorhun: I see, thank you for the explanation

ADR/0022-secret-mgmt-for-user-workloads.md
mmorhun: Thank you

mmorhun (approved):

ADR/0022-secret-mgmt-for-user-workloads.md
```diff
@@ -26,65 +29,142 @@ Accepted
...
 ```
+
+#### Example: If RemoteSecret has to be created with target namespace and Environment
+```yaml
+kind: RemoteSecret
+metadata:
+    name: test-remote-secret-secret
+    labels:
+        appstudio.redhat.com/remotesecret-target-environment: prod
```
gazarenkov, line 108: Can we remove "remotesecret-" leaving just appstudio.redhat.com/target-environment for ex.
I believe the fact it is belongs to RemoteSecret is enough?

ADR/0022-secret-mgmt-for-user-workloads.md
```diff
@@ -26,65 +29,142 @@ Accepted
...
+        appstudio.redhat.com/remotesecret-target-component: m-service
+        appstudio.redhat.com/remotesecret-target-application: coffee-shop
+spec:
```
gazarenkov, line 111: Do we need to specify a k8s Secret Type somewhere?

ADR/0022-secret-mgmt-for-user-workloads.md
```diff
@@ -26,65 +29,142 @@ Accepted
...
- ...
+    namespace: jdoe-tenant
+spec:
+    target:
+        - namespace: spi-test-target
+          secretName: test-remote-secret-secret-5bmq9
+status:
+    TBD
 ```
+> :warning: **Note for this and following:** `RemoteSecret` is not Ready, `SecretData` referenced to it has to be uploaded. See `Uploading secret data to RemoteSecret` example.
```
metlos, line 51: maybe explain somewhere what `Ready` means for the remote secret.

ADR/0022-secret-mgmt-for-user-workloads.md
```diff
@@ -26,65 +29,142 @@ Accepted
...
+        appstudio.redhat.com/remotesecret-target-component: m-service
+        appstudio.redhat.com/remotesecret-target-application: coffee-shop
+spec:
+    target:
+        - namespace: jdoe-tenant
+          secretName: test-remote-secret-secret-5bmq9
+status:
+    TBD
+```
+> **Note :** The component that will keep the connection between Environments and RemoteSecret's target should track all environments including the one that can be created after RemoteSecret creation.
```
metlos, line 135: As far as I looked this is the first mention of some kind of external component that should be keeping the environment and the remote secret target(s) in sync. You mention this component and the reason for it to be TBD at the very end of this chapter, but IMHO it should be mentioned before the examples so that the readers can anticipate it.

ADR/0022-secret-mgmt-for-user-workloads.md
metlos: it's there just to say that this info is meant for remote secrets. it could get confused with deployment target which I believe doesn't overlap with this?

ADR/0022-secret-mgmt-for-user-workloads.md
```diff
@@ -95,12 +175,11 @@ spec:
...
-|     Role      | Permissions             | API Groups                | Verbs                                   | Resources
-|---------------|-------------------------|---------------------------|-----------------------------------------|----------------------------------------------------------------------
-| Contributor   | RemoteSecret            | appstudio.redhat.com      | get, list, watch                        | remotesecret
-| Maintainer    | RemoteSecret            | appstudio.redhat.com      | get, list, watch, create, update, patch | remotesecret
-| Admin         | RemoteSecret            | appstudio.redhat.com      | get, list, watch, create, update, patch | remotesecret
-
+| Role        | Permissions  | API Groups           | Verbs                                   | Resources    |
+|-------------|--------------|----------------------|-----------------------------------------|--------------|
+| Contributor | RemoteSecret | appstudio.redhat.com | get, list, watch                        | remotesecret |
+| Maintainer  | RemoteSecret | appstudio.redhat.com | get, list, watch, create, update, patch | remotesecret |
```
gorkem, line 181: Since maintainers are not permitted to create regular secrets today, the create will have no effect. Should we delay the introduction of these permissions until an alternate way to introduce secret data is available  or they are able to create secrets?

ADR/0022-secret-mgmt-for-user-workloads.md
gazarenkov: Yes, it is definitely for remotesecret b/c it is part of its metadata (it's RemoteSecret's label)

In fact I do not think "-target" is bringing a lot of information here as well. 
IMO, the same as for remotesecret-, it just makes the name longer for nothing.
It even may act as additional point of confusion for users. Talking RHTAP terminology it is JUST environment/component/application. Prefix target- may make one think that it is something else.

ADR/0022-secret-mgmt-for-user-workloads.md
gazarenkov: I think there are 2 means of Status:
- related to StoredData. Ready would mean it is stored (ready to use by RemoteSecret)
- related to Targets. Ready would mean (per each) the Secret is created (ready to use by business logic)

ADR/0022-secret-mgmt-for-user-workloads.md
gazarenkov: @metlos Agreed, we need the status which would reflect some state of working Secret. For start we can just reflect the fact that it is created, then we can check its healthy and maybe later even refresh its data.
And we can evolve the specification with it.

ADR/0022-secret-mgmt-for-user-workloads.md
gazarenkov: It may be confusing indeed.
It seems that only Admins can normally deal with the Secrets now, right?

If so, I'd vote to decrease Maintainer permissions (to Contributor level) for the time.

ADR/0022-secret-mgmt-for-user-workloads.md
```diff
@@ -10,6 +10,9 @@ Accepted
...
 * As a StoneSoup component, I expect to have the ability to securely upload a secret, associate it with the component, and propagate it to the target environment.
+* It is important to distinguish the process of uploading confidential information and deploying it in the destination environment. This should/can be performed by different roles: first, creating a searchable `RemoteSecret` with linked `SecretData` that is stored remotely on `SecretStorage`, and second, finding this `RemoteSecret` and creating a working Secret with `SecretData` linked to it.
+* It should be possible to set multiple destination targets.
+* Any consumer  should be able to search should be able to perform a search of existing `RemoteSecret` by target environment/component/application.
```
mshaposhnik, line 15:
```suggestion
* Any consumer should be able to perform a search of existing `RemoteSecret` by target environment/component/application.
```

ADR/0022-secret-mgmt-for-user-workloads.md
gazarenkov: @skabashnyuk 
> Now the question of changing the type of an already uploaded secret, it's an interesting point to consider. If someone decides to change the type of an already uploaded secret, the controller should check if the new type is compatible with the existing one.

I am not sure how valid it is in general. 
The question of "compatibility" is not coming from the secret format, secret is just a transport to bring info to the business logic. For example: user application which uses uname/password env variables does not care about the secret bringing it. So, I really doubt compatibility validation can be performed on RS framework level.
To me the choice looks like:
- RS always trying to sync the secrets with RS applying the changes with no conditions
- RS.spec.secret is immutable i.e. RS may change it's state to "unsynced" (for ex) but not try to sync existed secrets (I like this option more for the time - it looks more fair and straightforward)

metlos: >     RS.spec.secret is immutable i.e. RS may change it's state to "unsynced" (for ex) but not try to sync existed secrets (I like this option more for the time - it looks more fair and straightforward)

This is indistinguishable from the case where the user updates the secret that is the target of the remote secret. In that case, we IMHO should update the secret to conform to the spec.

All in all, I think we should be "dumb" in this regard - we just keep the target secret in sync with the `spec.secret` of the RS. If this means changing the type of the secret, so be it. It may render the data of the secret unusable in the target (because the entries `data` may be incompatible with the new secret type) but that is "not our problem" so to speak. We just reconcile the state.

I think not having the secret spec in the RS but only mirror what we get in the upload secret is a very non-standard thing to do, because there would be no trace of what the target secret of the RS should look like anywhere in the CRD.

On contrary, I would go as far as saying that we should enhance the CRD with a "set of allowed/enforced data fields" so that the spec of the RS documents what kind of secret it is going to create in as much detail as possible. We could also add some smarts to this with marking the RS as invalid if the secret type and allowed fields were incompatible.

ADR/0022-secret-mgmt-for-user-workloads.md
```diff
@@ -26,65 +29,156 @@ Accepted
...
+        appstudio.redhat.com/component: m-service
+        appstudio.redhat.com/application: coffee-shop
+spec:
+    secret:
+        name: test-remote-secret-secret
+    target:
+        - namespace: jdoe-tenant
+status:
+    TBD
+```
```
gazarenkov, line 147: Do we really need so many examples of creating RS with targets?
It makes reader think that is it the main usecase while the documentation tells opposite.
Does it makes sense to demonstrate more how to make RS with the Secret of different types, with or w/o SA instead?

ADR/0022-secret-mgmt-for-user-workloads.md
```diff
@@ -26,65 +29,156 @@ Accepted
...
+                      generateName: sa-from-remote-
+    targets:
+        - namespace: "test-target-namespace-1"
+        - namespace: "test-target-namespace-2"
+        - namespace: "test-target-namespace-3"
+        - namespace: "test-target-namespace-rainbow"
+          apiUrl: "over-the-rainbow"
+          clusterCredentialsSecret: "team-a--prod-dtc--secret"
+status:
+    TBD
```
metlos, line 60:
```suggestion
  conditions:
  - lastTransitionTime: "..."
    message: ""
    reason: DataFound
    status: "True"
    type: DataObtained
  - lastTransitionTime: "..."
    message: "some of the targets were not deployed to"
    reason: PartiallyInjected
    status: "False"
    type: Deployed
  targets:
  - namespace: "test-target-namespace-1"
    secretName: secret-from-remote-lsdjf
    serviceAccountNames:
    - sa-from-remote-llrkt
  - namespace: "test-target-namespace-2"
    secretName: secret-from-remote-lemvs
    serviceAccountNames:
    - sa-from-remote-lkejr
  - namespace: "test-target-namespace-3"
    secretName: secret-from-remote-kjfdl
    serviceAccountNames:
    - sa-from-remote-lmval
  - namespace: "test-target-namepace-rainbow"
    apiUrl: "over-the-rainbow"
    error: "Connection refused"
```

ADR/0022-secret-mgmt-for-user-workloads.md
```diff
@@ -26,65 +29,156 @@ Accepted
...
-  name: jdbc-connection-parameter-prod
+    name: test-remote-secret-secret
 spec:
-  secret:
-    environment: prod
-    name: jdbc-connection-parameter
- ...
+    target:
+status:
+  TBD
```
metlos, line 74:
```suggestion
  conditions:
  - lastTransitionTime: "..."
    message: ""
    reason: AwaitingData
    status: "False"
    type: DataObtained
```

ADR/0022-secret-mgmt-for-user-workloads.md
```diff
@@ -26,65 +29,156 @@ Accepted
...
- ...
+    name: test-remote-secret-secret
+spec:
+    secret:
+        name: test-remote-secret-secret
+    target:
+        - namespace: jdoe-test-ns
+        - namespace: jdoe-prod-ns
+status:
+    TBD
```
metlos, line 90:
```suggestion
  conditions:
  - lastTransitionTime: "..."
    message: ""
    reason: DataFound
    status: "True"
    type: DataObtained
  - lastTransitionTime: "..."
    message: ""
    reason: Injected
    status: "True"
    type: Deployed
  targets:
  - namespace: "jdoe-test-ns"
    secretName: test-remote-secret
  - namespace: "jdoe-prod-ns"
    secretName: test-remote-secret
```

ADR/0022-secret-mgmt-for-user-workloads.md
```diff
@@ -26,65 +29,156 @@ Accepted
...
+spec:
+    secret:
+        name: test-remote-secret-secret
+    target:
+        - namespace: jdoe-test-ns
+        - namespace: jdoe-prod-ns
+          apiURL: https://somedomain.copm:443
+          clusterCredentialsSecret: team-a--prod-dtc--secret
+status:
+    TBD
```
metlos, line 109:
```suggestion
  conditions:
  - lastTransitionTime: "..."
    message: ""
    reason: DataFound
    status: "True"
    type: DataObtained
  - lastTransitionTime: "..."
    message: ""
    reason: Injected
    status: "True"
    type: Deployed
  targets:
  - namespace: "jdoe-test-ns"
    secretName: test-remote-secret
  - namespace: "jdoe-prod-ns"
    apiUrl: "https://somedomain.copm:443"
    secretName: test-remote-secret
```

ADR/0022-secret-mgmt-for-user-workloads.md
```diff
@@ -26,65 +29,156 @@ Accepted
...
+        appstudio.redhat.com/environment: prod
+        appstudio.redhat.com/component: m-service
+        appstudio.redhat.com/application: coffee-shop
+spec:
+    secret:
+        name: test-remote-secret-secret
+    target:
+        - namespace: jdoe-tenant
+status:
+    TBD
```
metlos, line 128:
```suggestion
  conditions:
  - lastTransitionTime: "..."
    message: ""
    reason: AwaitingData
    status: "False"
    type: DataObtained
```

metlos (commented):
I added suggestions to replace the TBD statuses with how they look like in our actual implementation at this point in time, along with some minor suggestions for replacing the out-of-date content.

gazarenkov (approved):

PR #103, opened 2023-05-19
Establish generic Tekton results conventions

arewm:
Generalize the Tekton result name conventions to support results from test-like and scan-like Tasks.


ADR/0027-tekton-results-naming-convention.md
```diff
@@ -0,0 +1,152 @@
...
+    "result": {
+      "type": "string",
+      "enum": ["SUCCESS", "FAILURE", "WARNING", "SKIPPED", "ERROR"]
+    },
+    "namespace": {
+      "type": "string"
+    },
+    "timestamp": {
+      "type": "string",
+      "pattern": "^[0-9]{10}$"
```
zregvart, line 68: If this is a string why not prescribe it to be in RFC3339 format?

ADR/0027-tekton-results-naming-convention.md
```diff
@@ -0,0 +1,152 @@
...
+- **namespace** - The rego namespace for the test policy. If not set, it will be assumed to have the value `default`
+- **timestamp** - An UNIX epoch timestamp of the test completion time
+- **successes** - The number of successful checks in the form of an integer
+- **note** - A short note provided by test workstream to provide additional information about the test
+- **failures** - The number of failed checks in the form of an integer
+- **warnings** - The number of warning checks in the form of an integer
+
+
+#### `TEST_OUTPUT` Schema Validation
+The output of the Tekton result **TEST_OUTPUT** will be validated using the [jsonschema validator package](https://github.com/santhosh-tekuri/jsonschema).
```
zregvart, line 51: Why mandate a specific implementation? Why not say JSON Schema validation will be performed.

ADR/0014-let-pipelines-proceed.md
```diff
@@ -32,7 +34,7 @@ their lower [Environments]?
 All scanning and linting TaskRuns should *succeed* even if they find problems in the content they
 are evaluating.
 
-Use the `HACBS_TEST_OUTPUT` result convention from [ADR-0013] to expose those results and render them
+Use the `TEST_OUTPUT` result convention from [ADR-0025] to expose those results and render them
```
hongweiliu17, line 37:
```suggestion
Use the `TEST_OUTPUT` result convention from [ADR-0027] to expose those results and render them
```

ADR/0014-let-pipelines-proceed.md
```diff
@@ -49,5 +51,5 @@ for users ([STONE-459]).
 
 [STONE-459]: https://issues.redhat.com/browse/STONE-459
 [Environments]: ../ref/application-environment-api.html#environment
-[ADR-0013]: 0013-integration-service-api-contracts.html
+[ADR-0025]: 0025-tekton-results-naming-convention.html
```
hongweiliu17, line 54:
```suggestion
[ADR-0027]: 0027-tekton-results-naming-convention.html
```

ADR/0014-let-pipelines-proceed.md
```diff
@@ -7,7 +7,9 @@
 
 Accepted
 
-Relates to [ADR 13. AppStudio Test Stream - API contracts](0013-integration-service-api-contracts.html)
+Relates to:
+* [ADR 13. AppStudio Test Stream - API contracts](0013-integration-service-api-contracts.html)
+* [ADR 25. Tekton Results Naming Convention](0025-tekton-results-naming-convention.html)
```
hongweiliu17, line 12:
```suggestion
* [ADR 27. Tekton Results Naming Convention](0027-tekton-results-naming-convention.html)
```

ADR/0027-tekton-results-naming-convention.md
arewm: I included it because that was the language in ADR-13. I see other ADRs not linking to a specific implementation. I changed the text.

ADR/0014-let-pipelines-proceed.md
arewm: Thanks for the catch!

ADR/0027-tekton-results-naming-convention.md
arewm: @sonam1412 @dirgim, would this be a change that we can propose in this ADR -- to change the timestamp validation?

ADR/0027-tekton-results-naming-convention.md
```diff
@@ -0,0 +1,150 @@
...
+        "medium": {
+          "type": "integer",
+          "minimum": 0
+        },
+        "low": {
+          "type": "integer",
+          "minimum": 0
+        }
+      },
+      "required": ["critical", "high", "medium", "low"]
```
mattreid, line 140: @arewm Do you want to consider if there's a fix available as part of this proposal (https://issues.redhat.com/browse/RHTAP-1015)? or is that too specific for this moment in time?

arewm: Clair and roxctl both have the fixable classification. We can propose something to meet that request in this ADR change or it can be left out.

If we are to try to encapsulate the fixable/unpatched concept into this return value, I would probably suggest a format like

```
{
  "vulnerabilities": {
    "critical": 1,
    "high": 0,
    "medium": 1,
    "low":0
  },
  "unpatched_vulnerabilities": {
    "critical": 0,
    "high": 1,
    "medium": 0,
    "low":1
  }
}
```

where the backwards-compatible case would default to those that are fixable and the unfixable ones have the new classification which _can_ be introduced in policies if desired.

arewm: @MartinBasti, FYI

MartinBasti: +1

also there are vulnerabilities with "unknown" severity, shall it be part of this ADR?

arewm: Do you have an example of this? Is there a bug tracking this?

MartinBasti: it's in clair output and also in quay.io 

https://quay.io/repository/martinbasti/single-container-app/manifest/sha256:edbffe44dd94babd6e4d25708cff444e23e7db8eb7c427ef1d0c50ae33d26f68?tab=vulnerabilities&fixable=true

There is a bug tracker, but I cannot get the ID right now

hongweiliu17 (commented):
I left some minor comments and don't have more point in my mind for now. Thanks for creating this ADR.

ADR/0027-tekton-results-naming-convention.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Results for Test-like Tasks
+Test-like tasks are those whose results can be immediately classified as successful or failed. If these tests are taken into account in the final suitability of an artifact for promotion or release, then all pass/fail determination would be deferred to the task run.
+
+The output of each Tekton task will be provided in a minimized [Tekton result](https://tekton.dev/docs/pipelines/tasks/#emitting-results) in JSON format listing all test failures. The name of the result will be **TEST_OUTPUT**.
```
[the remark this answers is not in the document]
mattreid: I'm hesitant to call everything that gets included in these results a test _failure_, since the test won't always fail (our build pipeline tests try to always run successfully so they don't halt a pipeline run). In the case of clair, as an example, it's reporting results to Enterprise Contract, and the contract decides what to do with the information. If no contract is enforced, then the data produced by clair will never fail anything. 

Suggested alternate wording as an option to consider.

```suggestion
The output of each Tekton task will be provided in a minimized [Tekton result](https://tekton.dev/docs/pipelines/tasks/#emitting-results) in JSON format listing the categories of reported test results and the count of results within each.  The name of the result will be **TEST_OUTPUT**.
```

zregvart:
Could we examine using in-Toto [test result](https://github.com/in-toto/attestation/blob/main/spec/predicates/test-result.md) attestation format?

arewm:
@zregvart, I don't think that the in-toto attestation format will be able to work. Instead of reporting the number of tests in each category, it reports each of the tests in each category. While this is useful information, it will be too verbose for the minimized output needed in the Tekton results. If replacing a standard with the in-toto one, the detailed output would be more reasonable: https://github.com/redhat-appstudio/book/blob/main/ADR/0013-integration-service-api-contracts.md#detailed-conftest-output-json.

ADR/0027-tekton-results-naming-convention.md
```diff
@@ -0,0 +1,150 @@
...
+
+
+#### `TEST_OUTPUT` Schema Validation
+The output of the Tekton result **TEST_OUTPUT** can be validated using the following schema:
+
+```
+{
+  "$schema": "http://json-schema.org/draft/2020-12/schema#",
+  "type": "object",
+  "properties": {
```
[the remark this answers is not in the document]
arewm: I think it is a great idea to track the digests that are processed, but I think it would be better as a separate result. This would prevent the bleed-through of implementations around multi-arch/Image Indexes into generic test frameworks. Instead, if a test/scan framework _is_ aware of these situations, it enables us to continue to refine the specification to limit the blast radius.

Some potential result names could be `REFS_PROCESSED`, `IMAGES_PROCESSED`, ~`IMAGE_ASSOCIATIONS`~, `DIGEST_OUTPUT`, `ASSOCIATED_REFS_OUTPUT`, ....

ADR/0027-tekton-results-naming-convention.md
dirgim: I think we could make the change - we'd need to implement the change on all sides though since we're all using a string currently.

amfred:
@arewm I'd like to see if we can get this merged.  Looks like there's a typo you can fix, and then can you take a pass to make sure it's up to date based on the current implementation?

ADR/0027-tekton-results-naming-convention.md
dirgim: Fyi the [STONEINTG-425](https://issues.redhat.com/browse/STONEINTG-425) is now being implemented so the `unknown` severity vulnerabilities will start to show up in this output this week.

ADR/0027-tekton-results-naming-convention.md
arewm: > I think we could make the change to use the int for the timestamp

What would be the proposed pattern to follow? `YYYY-MM-DDTHH:MM:SSZ`?

ADR/0027-tekton-results-naming-convention.md
arewm: I added this into these into the ADR.

ADR/0027-tekton-results-naming-convention.md
zregvart: My 2c, either we change the type to a number, or we change to a standard format (RFC3339).

hongweiliu17: I think RFC3339 format is more readable.

dirgim: Thinking about this some more, I also agree that we can switch to RFC3339 (So e.g. `2019-10-12T07:20:50.52Z`) and keep the string type for this field.

ADR/0013-integration-service-api-contracts.md
```diff
@@ -4,7 +4,7 @@ Date: 2023-01-30
 
 ## Status
 
-Accepted
+Deprecated by [ADR 27. Tekton Results Naming Convention](0030-tekton-results-naming-convention.html).
```
sonam1412, line 7: This should be changed from ADR 27. -> ADR 30.

sonam1412 (changes requested):

ADR/0013-integration-service-api-contracts.md
sonam1412, line 7: This should be changed from ADR 27. -> ADR 30

ADR/0014-let-pipelines-proceed.md
```diff
@@ -7,7 +7,9 @@
 
 Accepted
 
-Relates to [ADR 13. AppStudio Test Stream - API contracts](0013-integration-service-api-contracts.html)
+Relates to:
+* [ADR 13. AppStudio Test Stream - API contracts](0013-integration-service-api-contracts.html)
+* [ADR 27. Tekton Results Naming Convention](0030-tekton-results-naming-convention.html)
```
sonam1412, line 12: This should be ADR 30.

ADR/0027-tekton-results-naming-convention.md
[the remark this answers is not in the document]
arewm: This has been addressed in the most recent commit using a result named `IMAGES_PROCESSED`.

ADR/0027-tekton-results-naming-convention.md
arewm: This has been modified in the most recent commit.

ADR/0030-tekton-results-naming-convention.md
```diff
@@ -0,0 +1,257 @@
...
+    "high": 0,
+    "medium": 1,
+    "low":0,
+    "unknown":0
+  },
+  "unpatched_vulnerabilities": {
+    "critical": 0,
+    "high": 1,
+    "medium": 0,
+    "low":1
```
MartinBasti, line 115: This misses `unknown`

ADR/0030-tekton-results-naming-convention.md
```diff
@@ -0,0 +1,257 @@
...
+          "minimum": 0
+        },
+        "unknown": {
+          "type": "integer",
+          "minimum": 0
+        }
+      },
+      "required": ["critical", "high", "medium", "low"]
+    },
+    "unknown_vulnerabilities": {
```
MartinBasti, line 154:
```suggestion
    "unpatched_vulnerabilities": {
```

arewm:
@simonbaird @lcarva @sonam1412 @dirgim @rohitkrai03, can you review this ADR as it affects the integration, EC, and HAC teams? If you cannot review, can you delegate individual(s) to review on your behalf?

Notable changes in this ADR:
* Rename of `HACBS_TEST_OUTPUT` to `TEST_OUTPUT` result for test-like results
* Timestamp format changes for the `TEST_OUTPUT` result
* Rename of `CLAIR_SCAN_RESULT` to `SCAN_OUTPUT` result for scan-like results
* Addition of `unknown` severity and `unpatched_vulnerabilities` classification to `SCAN_OUTPUT`
* Creation of `IMAGES_PROCESSED` result to disambiguate the specific images that have been processed by a Task

ADR/0030-tekton-results-naming-convention.md
arewm: `unknown` is optional for backwards compatibility and was therefore intentionally not included

ADR/0030-tekton-results-naming-convention.md
```diff
@@ -0,0 +1,257 @@
...
+```
+
+In the case that the image processed is an Image Index, the result may look like the following:
+
+```
+{
+  "image": {
+    "pullspec": "quay.io/foo/bar:baz",
+    "digests": [
+      "@sha256:bb8[...]",
```
MartinBasti, line 214: Those digests don't hold any information about architecture, so basically I've to query them in advance to detect architecture and do something, so basically we can have index digest and we can do the same query

MartinBasti: Maybe I just finally get it, this is not output of build task, and input of integration task, but output from integration task, so at the same time we report, SCAN_RESULT and IMAGES_PROCESSED

ADR/0030-tekton-results-naming-convention.md
```diff
@@ -0,0 +1,257 @@
...
+### Results for Tasks that Process Multiple Images
+
+When scans and tests ingest an image reference as input, the actual content that is tested or scanned is ambiguous due to the potential presence of multi-arch images. If the image reference is an Image Manifest, for example, then that image would be known to be scanned. If the image reference is an Image Index, however, the Task could be processing all images referenced (i.e. all platforms/architectures), or just a single Image Manifest such as the one corresponding the the pod's architecture.
+
+In order to disambiguate the image references corresponding to the other Tekton results, a Task may output an additional result indicating the images that have been processed in order to obtain the otherwise specified result(s). The name of the result will be **IMAGES_PROCESSED**.
+
+#### Tekton Result Format for `IMAGES_PROCESSED`
+The test output of the Tekton result **IMAGES_PROCESSED** will be a JSON object that includes the context about the image references that have been processed.
+
+In the case that the image processed is an Image Manifest, the result may look like the following:
```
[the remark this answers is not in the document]
MartinBasti: nevermind

MartinBasti (approved):

ADR/0030-tekton-results-naming-convention.md
arewm: Yes, that is correct. This would surface the information about what images are processed by the test tasks. The validation that all relevant digests (i.e. corresponding to the different architectures) would be done at a later point according to a policy (i.e. via Enterprise Contract). We just need to make sure that the relevant contextual information is available to the policy engine in able for a fully informed decision to be made.

ADR/0030-tekton-results-naming-convention.md
```diff
@@ -0,0 +1,257 @@
...
+
+```
+{
+  "image": {
+    "pullspec": "quay.io/foo/bar:baz",
+    "digests": [
+      "@sha256:bb8[...]"
```
MartinBasti, line 201: digest shouldn't start with `@`

dirgim (approved):
LGTM

lcarva (approved):
This is pure gold :pinched_fingers:

karthikjeeyar (approved):
LGTM

ralphbean (approved):

PR #107, opened 2023-06-05
Common labels for constituent parts of the application

metlos:
This tries to define a common set of labels that should be used to identify supporting objects for application/environment/component.

### DIFF
--- a/ADR/0026-appstudio-object-labels.md
+++ b/ADR/0026-appstudio-object-labels.md
@@ -1,0 +1,36 @@
+# 26. Record architecture decisions
+
+Date: 2023-06-02
+
+## Status
+
+Accepted
+
+## Context
+
+There is a growing number of CRDs and other objects that together describe the application and its constituent parts. We need to use a consistent set of labels to describe to which
+part of the application certain supporting objects belong. This is not a functional requirement as such but a consistent set of labels supports troubleshooting efforts as well
+as general hygiene in the cluster (by being able to answer the questions like "what are all the objects associated with this environment?").
+
+Consistent set of labels will also enable finding related objects when co-operation between several components is required yet directly referencing the related objects in the specs or statuses of
+other objects is not feasible or practical.
+
+## Decision
+
+The deployments, pods, configmaps, secrets, etc. that together compose AppStudio itself will have the usual common suggested set of labels such as:
+* `app.kubernetes.io/name = has`
+* `app.kubernetes.io/part-of = appstudio`
+
+The more important part, though, is to identify the supporting objects defining the application model within AppStudio. This proposal establishes the 3 basic labels that can identify objects that are related to the most important parts of the model - application, environment and component:
+
+* `appstudio/application`
+* `appstudio/environment`
+* `appstudio/component`
+
+The "domain" part is kept intentionally very short yet unique not to unnecessarily add to the size of the Kubernetes object.
+
+## Consequences
+
+For this to make sense, all components that interact with and build upon the model objects need to consistently label the related objects they manage with the above labels.
+
+Additionally, having the names of the application, environment and component as label values restricts their length to 63 characters.
--- a/ADR/0026-appstudio-object-labels.md
+++ b/ADR/0026-appstudio-object-labels.md
@@ -1,4 +1,4 @@
-# 26. Record architecture decisions
+# 26. Common labels for objects associtated to Applications
 
 Date: 2023-06-02
 
--- a/ADR/0026-appstudio-object-labels.md
+++ b/ADR/0026-appstudio-object-labels.md
@@ -17,11 +17,9 @@ other objects is not feasible or practical.
 
 ## Decision
 
-The deployments, pods, configmaps, secrets, etc. that together compose AppStudio itself will have the usual common suggested set of labels such as:
-* `app.kubernetes.io/name = has`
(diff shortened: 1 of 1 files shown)


ADR/0026-appstudio-object-labels.md
```diff
@@ -0,0 +1,36 @@
...
+Consistent set of labels will also enable finding related objects when co-operation between several components is required yet directly referencing the related objects in the specs or statuses of
+other objects is not feasible or practical.
+
+## Decision
+
+The deployments, pods, configmaps, secrets, etc. that together compose AppStudio itself will have the usual common suggested set of labels such as:
+* `app.kubernetes.io/name = has`
+* `app.kubernetes.io/part-of = appstudio`
+
+The more important part, though, is to identify the supporting objects defining the application model within AppStudio. This proposal establishes the 3 basic labels that can identify objects that are related to the most important parts of the model - application, environment and component:
```
arewm, line 24: Is there any criteria that we should have if we want to expand from this core set of three basic labels?

ADR/0026-appstudio-object-labels.md
```diff
@@ -0,0 +1,36 @@
...
+* `appstudio/environment`
+* `appstudio/component`
+
+The "domain" part is kept intentionally very short yet unique not to unnecessarily add to the size of the Kubernetes object.
+
+## Consequences
+
+For this to make sense, all components that interact with and build upon the model objects need to consistently label the related objects they manage with the above labels.
+
+Additionally, having the names of the application, environment and component as label values restricts their length to 63 characters.
```
arewm, line 36: The value of the labels will be limited to 63 characters because of the length of the label?

ADR/0026-appstudio-object-labels.md
```diff
@@ -0,0 +1,36 @@
...
+
+* `appstudio/application`
+* `appstudio/environment`
+* `appstudio/component`
+
+The "domain" part is kept intentionally very short yet unique not to unnecessarily add to the size of the Kubernetes object.
+
+## Consequences
+
+For this to make sense, all components that interact with and build upon the model objects need to consistently label the related objects they manage with the above labels.
```
arewm, line 34: Are all of these labels required at all times? What is the null-value if there isn't an applicable value?

ADR/0026-appstudio-object-labels.md
```diff
@@ -0,0 +1,36 @@
+# 26. Record architecture decisions
```
arewm, line 1: Can you provide a more descriptive title?

ADR/0026-appstudio-object-labels.md
```diff
@@ -0,0 +1,36 @@
...
+Date: 2023-06-02
+
+## Status
+
+Accepted
+
+## Context
+
+There is a growing number of CRDs and other objects that together describe the application and its constituent parts. We need to use a consistent set of labels to describe to which
+part of the application certain supporting objects belong. This is not a functional requirement as such but a consistent set of labels supports troubleshooting efforts as well
```
arewm, line 12: This also relates to https://issues.redhat.com/browse/RHTAP-785 requesting a unification of labels/annotations/finalizers. Would it make sense to expand this ADR to help add additional clarity? It has been suggested that this might be similar to a follow-on to ADR-10.

arewm (changes requested):

ADR/0026-appstudio-object-labels.md
metlos: This is a limitation of Kubernetes itself. The value of any label (as opposed to the name of the label) cannot exceed 63 characters. I added a link to Kubernetes documentation that mentions this limitation.

I think HAS actually already implemented this limitation, am I right @elsony, am I right?

ADR/0026-appstudio-object-labels.md
metlos: That is a good point. I added a sentence about this not being required in all circumstances.

ADR/0026-appstudio-object-labels.md
metlos: Yep, I can :) Was so glad I finally put the ADR together that I forgot about the most important thing :) It is now updated.

ADR/0026-appstudio-object-labels.md
metlos: Thanks for pointing this out. I think this is very much related to https://issues.redhat.com/browse/RHTAP-785. I think the naming of the finalizers falls under the troubleshooting umbrella, that is part of this ADR.

metlos: As for the ADR-10, I'm not sure these two ADRs talk about the same thing. The labeling in ADR-10 is for the purpose of the monitoring of the resource usage, while this ADR is more about internal consistency and troubleshooting.

metlos: I've replied in https://issues.redhat.com/browse/RHTAP-785?focusedId=22409052&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-22409052 to explain why I think this ADR is still slightly different from what is being asked for there though.

amfred:
@metlos and @arewm I'm inclined to get this merged, and it looks like it's up to date.  What do you think?

arewm (changes requested):
The ADR number and date need to be updated. Other than that, we can go ahead and merge this.

PR #113, opened 2023-06-27
First draft of ADR on ComponentReferences

ralphbean:
This PR is related to https://github.com/redhat-appstudio/book/pull/114 as both are proposed solutions to the same underlying problem.

PR #114, opened 2023-06-28
First draft of ADR on related pipelines

arewm:
Part of the support for OLM operators within AppStudio is ensuring that component relationships can be maintained. If there are components that are dependent on each other, then a change in one image should be able to precipitate a change in another image to ensure that the initial change can be fully tested.

This PR is related to #113 as both are proposed solutions to the same underlying problem.

### DIFF
--- a/ADR/00xx-related-pipelines.md
+++ b/ADR/00xx-related-pipelines.md
@@ -1,0 +1,104 @@
+# Related Pipelines
+
+* Date 2023-06-xx
+
+## Status
+
+Proposed
+
+## Context
+
+As an AppStudio user, I want to be able to build and test multiple coupled components which depend on each other. I want that process to be easy.
+
+There are three use cases in scope for this document:
+
+As an initial draft and to simplify comparison, these use cases are not changed from the content proposed in [PR#113](https://github.com/redhat-appstudio/book/pull/113).
+
+* A user team has their own **intermediate base image**. When they propose an update to their base image with new content, they want to see if that’s going to break any of their Components that depend on it before merging. AppStudio should posit what rebuilds of those Components will look like and if they will pass their tests, and report that feedback back to the original pull request that updated content of the intermediate base image ([RHTAP-967](https://issues.redhat.com/browse/RHTAP-967)). A real example of this is in stolostron, where a [layered image component](https://github.com/stolostron/console/blob/main/Dockerfile.mce.prow#L20) refers to a [common parent image](https://github.com/stolostron/common-nodejs-parent).
+  * In this case, the app images must be rebuilt for the intermediate image update to really be evaluated in a functional way.
+  * This is a many-to-one dependency. Many component images depend on one intermediate base image.
+* A user team has an **OLM operator**. When they propose an update to one of their operands with new code, they want to see if that’s going to break their operator. AppStudio should posit what a rebuild of the operator bundle will look like and if it passes its tests and report that feedback back to the original pull request that updated the code in one of the operand images ([RHTAP-992](https://issues.redhat.com/browse/RHTAP-992)). A real example of this is in [gatekeeper](https://github.com/gatekeeper/gatekeeper-operator) where the operator repo contains both the [controller code](https://github.com/gatekeeper/gatekeeper-operator/blob/main/controllers/gatekeeper_controller.go) and the [bundle metadata](https://github.com/gatekeeper/gatekeeper-operator/blob/main/config/manifests/bases/gatekeeper-operator.clusterserviceversion.yaml), which need to be built into separate images([1](https://github.com/gatekeeper/gatekeeper-operator/blob/main/Dockerfile) and [2](https://github.com/gatekeeper/gatekeeper-operator/blob/main/bundle.Dockerfile)), separate Components in AppStudio.
+
+  * In this case, the bundle image must be rebuilt for the operand image update to be really be evaluated in a functional way.
+  * This is a one-to-many dependency. One operator bundle depends on many operand images.
+* A user team just has **two components that depend on each other in a functional way**. When they propose an update to one component, they want to submit a corresponding change to the second component, and have those tested together before merging both.
+  * In this case, on any given day, component B’s PRs may or may not have dependencies on PRs from component A. Not all changes are linked.
+  * This is a many-to-many dependency. The user may have lots of components that depend on lots of other components.
+
+In theory **any combination of the cases above** could be present in an app. They could have one intermediate base image, many operand images that depend on that, and one or more operator bundles which depend on those.
+
+Today, users work around how complicated it is to manage digests themselves by instead using *floating tags*, which have the benefit of being easy to use - no need to update - but have the problem of being unclear. It's not exactly clear what you're building against if you refer to it by tag. Potentially insecure. We want to make it easy for users to do better.
+
+## Decision
+
+In order to support linking of multiple components, any component dependency will be recorded as part of the Tekton definition to be run for pull request event types.
+
+### User flow
+
+Users will be able to specify dependencies between components by modifying the pull request Tekton definition or by adding configuration to the UI which will result in AppStudio representing the configuration in the Tekton definition. This definition will allow the specification of whether dependent components should be run in parallel or serially when combined.
+
+### Pipelines as Code (PaC)
+
+PaC will expand its design to include orchestration logic. This orchestrator will enable a plugable scheduler which is capable of identifying the dependent repositories and Tekton definitions (composing the metadata required to specify a component in AppStudio). When a component's Tekton definition indicates a dependency on another component, PaC will be able to combine the defined pipelines so that the resulting Pipeline definition builds images in the proper order.
+
+When identifying the pipeline definitions to merge, PaC will first look at the linked components for the same branch where a build was triggered from. If no matching branches are found then the default branch will be used.
+
+PaC will need to recursively expand the DAG to ensure that all dependent components are linked.
+
+### Build service
+
+The build service will be able to detect any component pull specs from earlier in a composite pipeline that differ from a previous Snapshot and replace all references of the previous pull spec with the new pull spec before building the component.
+
+### Integration service
+
+The integration service will be able to support updating multiple Component updates from a single PipelineRun such that all components built in a composite pipeline can be tested together as an Application in any configured integration tests.
+
+### Eventual consistency
+
+This flow of related changes as proposed in this ADR is limited to changes in the pull request Tekton definitions. It enables related components to be rebuilt as needed to ensure that image references are maintained and up to date. Once pull requests are merged for one component and the resulting container image is built, the resulting Snapshot will be "out of sync" in that some references in dependent images (i.e. RelatedImages in bundles or image references in a Dockerfile) will not match the latest artifact in the Snapshot. When build dependencies within the same Snapshot are enabled ([RHTAP-967](https://issues.redhat.com/browse/RHTAP-967)), this synchronization issue will be detected and relevant policies will fail.
+
+After a new Snapshot is produced, a Renovatebot will be responsible to propogating the updated image references to all other Components within the Application. After these pull requests have all been merged, consistency should be achieved again and all relevant policies should pass. 
+
+## Applied to Use Cases
+
+Let's apply the architecture to some use cases, and see how it plays out:
+
+### Common Parent Image
+
+Scenario: an application image depends on a common parent image. The user has 1 Application, and 2 Components. One of them is the parent image. The other image is built FROM that image.
+
+* The common parent image will register a dependency to all other Components
+* After the parent image is built, all other dependent Component images will be rebuilt.
+* If any of these dependent components are further dependent on images (i.e. a bundle image), then those images will also be rebuilt.
+* Intermediate image references can be replaced either by digest or by full image reference as these are used within the same snapshot. The full pull spec should match in any dependent image.
+
+### OLM Operators, with components in different repos
+
+Scenario: the user has 5 components. One of them is a “bundle” image. It contains references to the other four images by digest. None of the images are in the same repo.
+
+* Each non-bundle component declares that it depends on the bundle image.
+* When a change is introduced into any of the non-bundle images, the pullspec will be replaced in the bundle image to enable the change to be tested.
+* Since the pullspec location (i.e. registry and repository) might not be consistent with that in the snapshot due to the fact that the bundle image is targeting the _final_ location of the image, replacement will have to be performed by only finding and replacing the image digests.
+
+### OLM Operators, with components in the same repo
+
+Another OLM Operator Scenario: an operator git repo contains both the controller code and the bundle metadata.
+
+* The bundle component will not indicate any dependency in its PR definition.
+* The operator PR tekton definition will indicate a dependency on the bundle image.
+* The bundle image will be rebuilt after the operator is built to replace the pullspec referenced in the bundle image.
+* Since the pullspec location (i.e. registry and repository) might not be consistent with that in the snapshot due to the fact that the bundle image is targeting the _final_ location of the image, replacement will have to be performed by only finding and replacing the image digests.
+
+## Consequences
+
+* The creation and maintenance of PRs for builds running from AppStudio components remains unchanged (an alternative design that we considered involved creating PRs on other components to maintain the references). This means that the effect on the user managing multiple components in an application is reduced as there are fewer PRs to maintain and keep track of. 
+* The user can pin their digest references in git. AppStudio will automate maintaining them. No magical resolution of tags in the buildsystem at build time, or worse at runtime.
+* Changes will be required to multiple AppStudio components including a large change to PaC. This might be too much change across the various services to support this user flow.
+* The composite pipelines generated have the potential to rebuild many container images that will always be "throw-away." This can result in an undesired increased cost of running builds.
+* Users may have only a few Components, or they may have many (many dozens) of Components. Once they get past so many component dependencies, we suspect that users will likely change from checking that all dependent images work with the new parent image to "sharing" the verification load: building the image and pushing it out for other components/dependencies to update and test within their own PRs. With this design, the user can achieve this by deleting their dependent repositories. The parent image will be built and tested as normal. The user can still hypothetically construct their own "dependent PR" to test a particular layered component on an unmerged parent image change.
+
+## Open Questions
+
+* To reduce the time of inconsistency, can renovatebot create the PRs on dependent components immediately or should renovatebot always respect the configuration files? Does that still make sense if renovatebot is not being run on a cron basis?
+* Would there be a way for a user to be able to issue a command in a PR to enable a one-off linking of multiple components or would this type of linking need to occur via a temporary modification to the Tekton pipeline definition in the same PR?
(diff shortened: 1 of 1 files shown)


ADR/00xx-related-pipelines.md
```diff
@@ -0,0 +1,104 @@
...
+
+Users will be able to specify dependencies between components by modifying the pull request Tekton definition or by adding configuration to the UI which will result in AppStudio representing the configuration in the Tekton definition. This definition will allow the specification of whether dependent components should be run in parallel or serially when combined.
+
+### Pipelines as Code (PaC)
+
+PaC will expand its design to include orchestration logic. This orchestrator will enable a plugable scheduler which is capable of identifying the dependent repositories and Tekton definitions (composing the metadata required to specify a component in AppStudio). When a component's Tekton definition indicates a dependency on another component, PaC will be able to combine the defined pipelines so that the resulting Pipeline definition builds images in the proper order.
+
+When identifying the pipeline definitions to merge, PaC will first look at the linked components for the same branch where a build was triggered from. If no matching branches are found then the default branch will be used.
+
+PaC will need to recursively expand the DAG to ensure that all dependent components are linked.
```
Roming22, line 46: Expand the acronym on first use.

ADR/00xx-related-pipelines.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Build service
+
+The build service will be able to detect any component pull specs from earlier in a composite pipeline that differ from a previous Snapshot and replace all references of the previous pull spec with the new pull spec before building the component.
```
Roming22, line 50: that differs?

ADR/00xx-related-pipelines.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Eventual consistency
+
+This flow of related changes as proposed in this ADR is limited to changes in the pull request Tekton definitions. It enables related components to be rebuilt as needed to ensure that image references are maintained and up to date. Once pull requests are merged for one component and the resulting container image is built, the resulting Snapshot will be "out of sync" in that some references in dependent images (i.e. RelatedImages in bundles or image references in a Dockerfile) will not match the latest artifact in the Snapshot. When build dependencies within the same Snapshot are enabled ([RHTAP-967](https://issues.redhat.com/browse/RHTAP-967)), this synchronization issue will be detected and relevant policies will fail.
+
+After a new Snapshot is produced, a Renovatebot will be responsible to propogating the updated image references to all other Components within the Application. After these pull requests have all been merged, consistency should be achieved again and all relevant policies should pass.
```
Roming22, line 60: responsible to propogating -> responsible for propagating

ADR/00xx-related-pipelines.md
```diff
@@ -0,0 +1,104 @@
...
+
+* The bundle component will not indicate any dependency in its PR definition.
+* The operator PR tekton definition will indicate a dependency on the bundle image.
+* The bundle image will be rebuilt after the operator is built to replace the pullspec referenced in the bundle image.
+* Since the pullspec location (i.e. registry and repository) might not be consistent with that in the snapshot due to the fact that the bundle image is targeting the _final_ location of the image, replacement will have to be performed by only finding and replacing the image digests.
+
+## Consequences
+
+* The creation and maintenance of PRs for builds running from AppStudio components remains unchanged (an alternative design that we considered involved creating PRs on other components to maintain the references). This means that the effect on the user managing multiple components in an application is reduced as there are fewer PRs to maintain and keep track of. 
+* The user can pin their digest references in git. AppStudio will automate maintaining them. No magical resolution of tags in the buildsystem at build time, or worse at runtime.
```
Roming22, line 95: AppStudio will automate maintaining them. -.> AppStudio will automate their maintenance.
buildsystem -> build system.
That last sentence is missing a verb. Maybe start it with "There is" or "There will be".

ADR/00xx-related-pipelines.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Applied to Use Cases
+
+Let's apply the architecture to some use cases, and see how it plays out:
+
+### Common Parent Image
+
+Scenario: an application image depends on a common parent image. The user has 1 Application, and 2 Components. One of them is the parent image. The other image is built FROM that image.
+
+* The common parent image will register a dependency to all other Components
```
Roming22, line 70: I have an issue with the usage of the word "dependency". If I say "A depends on B", B can be built without A, but A needs B to be built. I.e if I have no dependencies, I can be built first.

This document talks about building a graph, and suggest that the root node(s) depend on the children nodes. This went against my expectation, and I found it confusing.

Maybe we can use graph terms? E.g. `Components directly impacted by a rebuild of the parent will be registered as children in the parent definition`?

ADR/00xx-related-pipelines.md
```diff
@@ -0,0 +1,104 @@
...
+* The composite pipelines generated have the potential to rebuild many container images that will always be "throw-away." This can result in an undesired increased cost of running builds.
+* Users may have only a few Components, or they may have many (many dozens) of Components. Once they get past so many component dependencies, we suspect that users will likely change from checking that all dependent images work with the new parent image to "sharing" the verification load: building the image and pushing it out for other components/dependencies to update and test within their own PRs. With this design, the user can achieve this by deleting their dependent repositories. The parent image will be built and tested as normal. The user can still hypothetically construct their own "dependent PR" to test a particular layered component on an unmerged parent image change.
+
+## Open Questions
+
+* To reduce the time of inconsistency, can renovatebot create the PRs on dependent components immediately or should renovatebot always respect the configuration files? Does that still make sense if renovatebot is not being run on a cron basis?
```
Roming22, line 102: You used `Renovatebot` earlier.

ADR/00xx-related-pipelines.md
```diff
@@ -0,0 +1,104 @@
...
+  * In this case, on any given day, component B’s PRs may or may not have dependencies on PRs from component A. Not all changes are linked.
+  * This is a many-to-many dependency. The user may have lots of components that depend on lots of other components.
+
+In theory **any combination of the cases above** could be present in an app. They could have one intermediate base image, many operand images that depend on that, and one or more operator bundles which depend on those.
+
+Today, users work around how complicated it is to manage digests themselves by instead using *floating tags*, which have the benefit of being easy to use - no need to update - but have the problem of being unclear. It's not exactly clear what you're building against if you refer to it by tag. Potentially insecure. We want to make it easy for users to do better.
+
+## Decision
+
+In order to support linking of multiple components, any component dependency will be recorded as part of the Tekton definition to be run for pull request event types.
```
ralphbean, line 34:
```suggestion
In order to support linking of multiple components, any component dependency will be encoded as part of the Tekton definition to be run for **pull request** event types.
```

ADR/00xx-related-pipelines.md
```diff
@@ -0,0 +1,104 @@
...
+
+### User flow
+
+Users will be able to specify dependencies between components by modifying the pull request Tekton definition or by adding configuration to the UI which will result in AppStudio representing the configuration in the Tekton definition. This definition will allow the specification of whether dependent components should be run in parallel or serially when combined.
```
ralphbean, line 38:
```suggestion
Users will be able to specify dependencies between components by modifying the pull request Tekton definition or by adding configuration to the UI which will result in AppStudio encoding the configuration in the Tekton definition. This definition will allow the specification of whether dependent components should be run in parallel or serially when combined.
```

ADR/00xx-related-pipelines.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Pipelines as Code (PaC)
+
+PaC will expand its design to include orchestration logic. This orchestrator will enable a plugable scheduler which is capable of identifying the dependent repositories and Tekton definitions (composing the metadata required to specify a component in AppStudio). When a component's Tekton definition indicates a dependency on another component, PaC will be able to combine the defined pipelines so that the resulting Pipeline definition builds images in the proper order.
```
ralphbean, line 42:
```suggestion
Pipelines as Code (PaC) will expand its design to include orchestration logic. This orchestrator will enable a plugable scheduler which is capable of identifying the dependent repositories and Tekton definitions (composing the metadata required to specify a component in AppStudio). When a component's Tekton definition indicates a dependency on another component, PaC will be able to combine the defined pipelines so that the resulting Pipeline definition builds images in the proper order.
```

ADR/00xx-related-pipelines.md
ralphbean:
```suggestion
PaC will need to recursively expand the directed acylic graph (DAG) to ensure that all dependent components are linked.
```

ADR/00xx-related-pipelines.md
ralphbean: I'm not sure how build-service will do this.

The work to be done: "replace some pullspec references with others"

* Should it be the responsibility of tasks in the composite pipeline?
* Should it be the responsibility of PaC when assembling the composite pipeline?

I don't think build-service can intervene at composite-pipeline-assembly time or at composite-pipeline-run-time.

ADR/00xx-related-pipelines.md
```diff
@@ -0,0 +1,104 @@
...
+
+PaC will need to recursively expand the DAG to ensure that all dependent components are linked.
+
+### Build service
+
+The build service will be able to detect any component pull specs from earlier in a composite pipeline that differ from a previous Snapshot and replace all references of the previous pull spec with the new pull spec before building the component.
+
+### Integration service
+
+The integration service will be able to support updating multiple Component updates from a single PipelineRun such that all components built in a composite pipeline can be tested together as an Application in any configured integration tests.
```
ralphbean, line 54:
```suggestion
The integration service will be able to support updating multiple Component updates from a single PipelineRun such that all components built in a composite pipeline can be tested together as an Application in any configured integration tests. In order to support this, the composite pipeline constructed by PaC will need to expose multiple results where integration-service can find the pullspecs of the images.
```

ADR/00xx-related-pipelines.md
ralphbean, line 60:
```suggestion
After a new Snapshot is produced, a Renovatebot instance will be responsible for propogating the updated image references to all other Components within the Application. After these pull requests have all been merged, consistency should be achieved again and all relevant policies should pass. 
```

arewm: I must have fixed this locally and not committed it. Thanks for the catch.

ADR/00xx-related-pipelines.md
```diff
@@ -0,0 +1,103 @@
...
+
+Pipelines as Code (PaC) will expand its design to include orchestration logic. This orchestrator will enable a plugable scheduler which is capable of identifying the dependent repositories and Tekton definitions (composing the metadata required to specify a component in AppStudio). When a component's Tekton definition indicates a dependency on another component, PaC will be able to combine the defined pipelines so that the resulting Pipeline definition builds images in the proper order.
+
+When identifying the pipeline definitions to merge, PaC will first look at the linked components for the same branch where a build was triggered from. If no matching branches are found then the default branch will be used.
+
+PaC will need to recursively expand the directed acyclic graph (DAG) to ensure that all dependent components are linked.
+
+### Build service
+
+The build service will be able to detect any component pull specs from earlier in a composite pipeline that differs from a previous Snapshot and replace all references of the previous pull spec with the new pull spec before building the component.
```
arewm, line 49: In https://github.com/redhat-appstudio/book/pull/114#discussion_r1254871127, @ralphbean commented:

> I'm not sure how build-service will do this.
> 
> The work to be done: "replace some pullspec references with others"
> 
> * Should it be the responsibility of tasks in the composite pipeline?
> * Should it be the responsibility of PaC when assembling the composite pipeline?
> 
> I don't think build-service can intervene at composite-pipeline-assembly time or at composite-pipeline-run-time.

ADR/00xx-related-pipelines.md
arewm: I am going to resolve this conversation as the original comment is resolved. I opened a new thread on this to continue the conversation from @ralphbean: https://github.com/redhat-appstudio/book/pull/114/files#r1256363848

ADR/00xx-related-pipelines.md
arewm: I think I might understand your issue with the word "dependency" here. I was thinking that A depends on B to test, but confusion can arise when those words are missing. I think clarity could be improved by illustration with a graph, but I also feel like the following representation is too much:

> Each component will be able to describe the builds which need to be triggered after its completion in order to properly test. When a PR build is started for a specific component, the triggered builds will be added as children to the current node. A DAG will be created by visiting each node and registering the children nodes based on the triggering relationships encoded. Once recursion has ceased, execution can proceed by building all Components in the next layer whose every parent has previously been built.

---

> This document talks about building a graph, and suggest that the root node(s) depend on the children nodes. This went against my expectation, and I found it confusing.

What is the expectation you are referring to? The use of the dependency concept again or related to the graph construction?

ADR/00xx-related-pipelines.md
arewm: I was thinking that the functionality would be somewhat similar to how Cachi2 has been implemented where the Dockerfile is modified based on the prefetched input: https://github.com/redhat-appstudio/build-definitions/blob/main/task/buildah/0.1/buildah.yaml#L146-L152

The build service would have to ingest the results of the previous tasks and replace the references matching those earlier components. For common parent images, I would expect the images from the workspace's quay organization to be leveraged so it should be easy to detect required updates. For bundle images where the image references may be present with modified pull specs (i.e. preparing for a release), the replacement may be harder unless the build service has access to the latest Snapshot or Global Candidate List so that it can determine what the _previous/latest_ digest of the rebuilt image is and then substitute solely based on digest matching.

ADR/00xx-related-pipelines.md
```diff
@@ -0,0 +1,103 @@
...
+
+Today, users work around how complicated it is to manage digests themselves by instead using *floating tags*, which have the benefit of being easy to use - no need to update - but have the problem of being unclear. It's not exactly clear what you're building against if you refer to it by tag. Potentially insecure. We want to make it easy for users to do better.
+
+## Decision
+
+In order to support linking of multiple components, any component dependency will be encoded as part of the Tekton definition to be run for **pull request** event types.
+
+### User flow
+
+Users will be able to specify dependencies between components by modifying the pull request Tekton definition or by adding configuration to the UI which will result in AppStudio encoding the configuration in the Tekton definition. This definition will allow the specification of whether dependent components should be run in parallel or serially when combined.
```
mattreid, line 37: If the relationship is encoded in the component's pipeline definition, this becomes an advanced use case that can only be utilized with a custom pipeline? That seems reasonable, but we have kicked around the idea of a team having the ability to use PaC and having githooks, without having to maintain their own pipeline definition everywhere. 

If the source of truth is in github, how would the UI get access to represent dependency information? Typically, that's trying to pull info from a CR that it can easily access.

mattreid: What is considered a valid dependency in this model? Would it have to be part of their current application? or could a dependency be from another app/namespace?

ADR/00xx-related-pipelines.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Eventual consistency
+
+This flow of related changes as proposed in this ADR is limited to changes in the pull request Tekton definitions. It enables related components to be rebuilt as needed to ensure that image references are maintained and up to date. Once pull requests are merged for one component and the resulting container image is built, the resulting Snapshot will be "out of sync" in that some references in dependent images (i.e. RelatedImages in bundles or image references in a Dockerfile) will not match the latest artifact in the Snapshot. When build dependencies within the same Snapshot are enabled ([RHTAP-967](https://issues.redhat.com/browse/RHTAP-967)), this synchronization issue will be detected and relevant policies will fail.
+
+After a new Snapshot is produced, a Renovatebot instance will be responsible for propagating the updated image references to all other Components within the Application. After these pull requests have all been merged, consistency should be achieved again and all relevant policies should pass.
```
mattreid, line 59: How would integration service know what it could try to release? or would it trigger an attempted release since code was merged, and fail until the other components are built and included in a releasable snapshot?

ADR/00xx-related-pipelines.md
mattreid: How would the child image be able to display its dependency? If I'm looking at the child image in the UI, how does HAC know to represent the dependency noted on the parent image? 

I had a similar comment to Romain's here, but on Ralph's proposal in #113 for 'depends on', in that it seemed intuitively backwards to me. From a graph perspective, it makes more sense to me to think of the child image depending on the parent, since it needs it to build, although I understand that we want a change to the parent to test the child, and the reverse wouldn't be true.

ADR/00xx-related-pipelines.md
```diff
@@ -0,0 +1,103 @@
...
+  * In this case, on any given day, component B’s PRs may or may not have dependencies on PRs from component A. Not all changes are linked.
+  * This is a many-to-many dependency. The user may have lots of components that depend on lots of other components.
+
+In theory **any combination of the cases above** could be present in an app. They could have one common parent image, many operand images that depend on that, and one or more operator bundles which depend on those.
+
+Today, users work around how complicated it is to manage digests themselves by instead using *floating tags*, which have the benefit of being easy to use - no need to update - but have the problem of being unclear. It's not exactly clear what you're building against if you refer to it by tag. Potentially insecure. We want to make it easy for users to do better.
+
+## Decision
+
+In order to support linking of multiple components, any component dependency will be encoded as part of the Tekton definition to be run for **pull request** event types.
```
rcerven, line 33: " as part of the Tekton definition"
=>
 as part of the PipelineRun Tekton definition in git in ./tekton

ADR/00xx-related-pipelines.md
rcerven, line 37: "This definition will allow the specification of whether dependent components should be run in parallel or serially when combined."

why would it be needed?

if A & B is dependent on C, why would you run A & B ever serially?

the only use case I can think of is to safe resources, but at the price of slowing this rebuild process

ADR/00xx-related-pipelines.md
```diff
@@ -0,0 +1,103 @@
...
+
+After a new Snapshot is produced, a Renovatebot instance will be responsible for propagating the updated image references to all other Components within the Application. After these pull requests have all been merged, consistency should be achieved again and all relevant policies should pass. 
+
+## Applied to Use Cases
+
+Let's apply the architecture to some use cases, and see how it plays out:
+
+### Common Parent Image
+
+Scenario: an application image depends on a common parent image. The user has 1 Application, and 2 Components. One of them is the parent image. The other image is built FROM that image.
```
rcerven, line 67: "One of them is the parent image. The other image is built FROM that image"
=>
One of them is the parent image. The other is child image which uses parent image in FROM

rcerven:
I like https://github.com/redhat-appstudio/book/pull/113 more than https://github.com/redhat-appstudio/book/pull/114

Also I think it is benefit to have dependency information under Component, which is also easily accessible in OCP

Also with UI for setting up those dependencies it would be more user friendly

Lastly probably also easier for debugging

chmouel:
paac is not going to implement rhtap specific logic in its code. 

If there is a request on a specific feature of paac out of this ADR, it would be nice to extract the exact feature you need to understand what you want to do from paac.

Trying to understand what is your goal,  is to be able to compose a pipelinerun out of multiple repository? would https://issues.redhat.com/browse/SRVKP-3020 will be helping you? where we give you the ability to give us the PipelineRun that needs to be run.

If you need something else please let us know but bear in mind that i am not aware of how appstudio works and this discussion would be easier if we can focus the exact feature needed from paac.

arewm:
Thanks for the reviews on this ADR. We decided to go with the `ComponentReferences` (#113) approach instead.

PR #115, opened 2023-06-29
STONEBLD-1455: Populate image-controller.md

tkdchen:


book/image-controller.md
```diff
@@ -1,2 +1,125 @@
 # Image Controller
 
+## Overview
```
arewm, line 3: Can you go into the more generic background a little more? Why do we need help setting up container image repositories? One example might be to increase isolation between images produced across various workspaces.

book/image-controller.md
```diff
@@ -1,2 +1,125 @@
...
+  visibility if the repository exists already.
+
+- `image.redhat.com/image`: information of created resources is set to this
+  annotation as a JSON string. Error message will also be set to here if there
+  is.
+
+## Set up image repository
+
+To request the controller to setup an image repository, annotate the Component
+CR with `image.redhat.com/generate: '{"visibility": "public"}'` or
```
arewm, line 23: Can you define CR on its first use? Is there a reference to the Component CR(D) that you can link to?

book/image-controller.md
```diff
@@ -1,2 +1,125 @@
...
+  application: city-transit
+  componentName: billing
+```
+
+As a result, when the setup succeeds,
+
+- A public image repository is created on Quay.io.
+- A robot account is created and associated with the repository properly.
+- A Kubernetes Secret is created alongside the Component CR, to which the
+  robot account token is written out.
```
arewm, line 48: Is this secret also bound to the ServiceAccount?

book/image-controller.md
```diff
@@ -1,2 +1,125 @@
...
+    application: city-transit
+    componentName: billing
+  ```
+
+Note that, annotation `image.redhat.com/generate` is removed already.
+
+## Cleanup on deletion of Component CR
+
+Image controller deletes the image repository and robot account from Quay.io
+when Component CR is requested to be deleted.
```
arewm, line 79: What are the conditions for this to happen? Does the image controller also delete the secret and unlink it or are those performed by some other action/process?

book/image-controller.md
```diff
@@ -1,2 +1,125 @@
...
+    image.redhat.com/image: "{\"image\":\"quay.io/redhat-user-workloads/image-controller-system/city-transit/billing\",\"visibility\":\"public\",\"secret\":\"secret-name\"}"
+  name: billing
+  namespace: image-controller-system
+spec:
+  application: city-transit
+  componentName: billing
+```
+
+As a result, repository becomes private on Quay.io and `visibility` changes to
+`private` in annotation `image.redhat.com/image`:
```
arewm, line 105: Is this only for new repositories that are created or will all repositories be converted to private? If all will be converted, will they be converted on their first rebuild or immediately?

book/image-controller.md
tkdchen: Yes. It is what build-service does.

book/image-controller.md
tkdchen: > What are the conditions for this to happen?

I'm not sure I get your point, but I'm trying to answer. Based on my understanding so far, image controller does not actually care about the conditions that would cause a Component to be deleted, instead it just simply clean up the related external resources when it detects the fact of "a Component is requested to be deleted".

> Does the image controller also delete the secret and unlink it or are those performed by some other action/process?

No. Ownership is built between Component and the Secret, i.e. a Component owns the Secret.

book/image-controller.md
tkdchen: I think it could be https://redhat-appstudio.github.io/book/ref/application-environment-api.html#component, or link to the https://github.com/redhat-appstudio/application-api/blob/main/api/v1alpha1/component_types.go#L75 ?

image-controller just cares about the annotation, not the CR definition. I use term "Component CR" in order to keep consistency with other documents, like build-service.md.

 > Can you define CR on its first use?

image-controller starts to handle only when both `image.redhat.com/generate` and `component.Status.Devfile` are set.

book/image-controller.md
```diff
@@ -1,2 +1,141 @@
 # Image Controller
 
+## Overview
+
+The Image Controller for AppStudio helps set up container image repositories
```
Michkov, line 5: it's not just helping it's setting up it directly.

book/image-controller.md
```diff
@@ -1,2 +1,141 @@
...
+for AppStudio Components. With image controller, AppStudio ensures one image
+repository per Component and each build pipeline will have dedicated secret to
```
Michkov, line 7: not for each build pipeline, each pipeline/pipelineRun for the compont is using the same repository.

book/image-controller.md
```diff
@@ -1,2 +1,141 @@
...
+kind: Component
+metadata:
+  annotations:
+    image.redhat.com/generate: '{"visibility": "public"}'
+  name: billing
+  namespace: image-controller-system
+spec:
+  application: city-transit
+  componentName: billing
+```
```
Michkov, line 44: I would add line with three dots `...` to show that the CR definition is not complete.

book/image-controller.md
```diff
@@ -1,2 +1,141 @@
...
+
+Image controller then creates following resources:
+
+- A public image repository is created on Quay.io.
+- A robot account is created and associated with the repository properly.
```
Michkov, line 49: I would remove that `properly`

book/image-controller.md
Michkov: immediately, that's configuration on repository level so it's immediatelly applied to all tags in the repository.

book/image-controller.md
```diff
@@ -1,2 +1,141 @@
...
+- `image.redhat.com/image`: information of created resources is set to this
+  annotation as a JSON string. Error message will also be set to here if there
+  is.
+
+## Set up image repository
+
+To request the controller to setup an image repository, annotate the Component
+CR with `image.redhat.com/generate: '{"visibility": "public"}'` or
+`image.redhat.com/generate: '{"visibility": "private"}'` depending on desired
+repository visibility.
```
tisutisu, line 28: Do we need to add any other secret when setting `visibility to private` ? How the secret related to private repo is handled?

tkdchen: It's irrelevant. Secret is not required when setting `visibility` to `private` in the annotation.

book/image-controller.md
arewm: You indicate this at the end of the section. Instead, you can you pull that out and explain it in the overview? If you give a brief explanation of the image controller (i.e. what it is responsible for) and then indicate how it interfaces with other services, then that would be helpful. You would be able to explain conceptually the whole process without the json/yaml blocks in between to separate the contexts.

book/image-controller.md
arewm: Thanks. Is the default public or private?

book/image-controller.md
arewm: By defining CR, I just meant defining it as Custom Resource. It is a nit, I haven't looked to see if other documents define it.

Are those two references in sync with each other? If you refer to something, it would be better to refer to another place in the book over the code.

mmorhun (approved):
LGTM.
We are working on new API (CRD) for Image Controller. It would be nice to merge this doc that describes current state, so we could better describe the transition later.

book/image-controller.md
tkdchen: > Are those two references in sync with each other? If you refer to something, it would be better to refer to another place in the book over the code.

I was told the API pages https://redhat-appstudio.github.io/book/ref/index.html are generated from the types defined in Go. I don't know if there is a mechanism to sync the possible changes regularly. If no, probably this is a point that can be improved, but the API should seldom change.

book/image-controller.md
tkdchen: This part will be updated according to the new ImageRepository CRD. The initial visibility depends on `spec.image.visibility` within the first creation of that CRD. If omitted, public is the default.

book/image-controller.md
```diff
@@ -1,2 +1,144 @@
...
+
+- `image.redhat.com/image`: information of created resources is set to this
+  annotation as a JSON string. Error message will also be set to here if there
+  is.
+
+## Set up image repository
+
+To request the controller to setup an image repository, annotate the Component
+CR with `image.redhat.com/generate: '{"visibility": "public"}'` or
+`image.redhat.com/generate: '{"visibility": "private"}'` depending on desired
```
rcerven, line 26: isn't it now in in this annotation ? image.redhat.com/image

book/image-controller.md
tisutisu: Got it. Thanks.

book/image-controller.md
tkdchen: Thanks for your comment. PTAL.

mmorhun (approved):

book/image-controller.md
```diff
@@ -1,2 +1,36 @@
...
+
+  Here is an example that requests a private image repository:
+
+  ```yaml
+  image.redhat.com/generate: '{"visibility": "private"}'
+  ```
+
+- `image.redhat.com/image`: image controller provides information of prepared resources to external services via this annotation, which includes the image repository URL, the visibility of that repository, and a secret name pointing to the created Kubernetes Secret.
+
+  If something went wrong during the requested operation, this annotation will include a field `Message` with a corresponding error message.
```
tisutisu, line 30: Will it be good to add an example here like `"message":"failed to generete image repository"`?

tkdchen: I would not add such example in this document. Instead, image-controller's own technial document like the README is proper to include necessary examples in detail.

Michkov (approved):

book/image-controller.md
```diff
@@ -1,2 +1,36 @@
 # Image Controller
 
+# Overview
+Image controller sets up and manages container image repository for an application's component. With this ability, AppStudio provides an isolated environment for image builds of a component that is each component has its own image repository and secret used to push every built image to that repository.
```
arewm, line 4:
```suggestion
Image controller sets up and manages container image repositories for an application's components. This enables greater component isolation within AppStudio where each component has its own image repository and secret for pushing images built via AppStudio.```

File: book/image-controller.md [hunk 14]
```diff
@@ -1,2 +1,36 @@
...
+
+Image controller takes these actions by watching on either specific annotation's change or deletion event of a [Component CR](https://redhat-appstudio.github.io/book/ref/application-environment-api.html#component)
+
+- Setup image repository. Image controller creates an image repository for the Component CR in a remote image registry as well as a robot account which is specific to that repository for image push. A Kubernetes Secret object is also created with that robot account token in order to make it available for build PipelineRun.
+
+- Visibility switch. Image controller is able to switch image repository's visibility between public and private.
+
+- Cleanup. When a Component CR is requested to be deleted, image controller will remove component's image repository and roboto account from the remote registry. The Kubernetes Secret will be removed alongwith the Component CR eventually due to the ownership established between them.
```
arewm, line 12:
```suggestion
The image controller can perform three actions on image repositories by watching for either specific annotation changes or deletion events of a [Component CR](https://redhat-appstudio.github.io/book/ref/application-environment-api.html#component):

- **Setup image repository**: Image controller creates an image repository for the Component CR in a remote image registry as well as a robot account which is specific to that repository for image push. A Kubernetes Secret object is also created with that robot account token in order to make it available for build PipelineRun.

- **Modify visibility**: Image controller is able to switch an image repository's visibility between public and private.

- **Cleanup**: When a Component CR is requested to be deleted, image controller will remove component's image repository and robot account from the remote registry. The Kubernetes Secret will be removed along with the Component CR eventually due to the ownership established between them.
```

arewm (changes requested):

arewm (approved):
LGTM, we can always clarify later as needed as well.

PR #118, opened 2023-07-11
chore(RHTAPWATCH-359): create ADR for availability probe framework

yftacherzog:
This change defines the convention according to which appstudio components' availability will be reported.

### DIFF
--- a/ADR/0027-availability-probe-framework.md
+++ b/ADR/0027-availability-probe-framework.md
@@ -1,0 +1,131 @@
+# 27. Availability Probe Framework
+
+* Date 2023-07-06
+
+## Status
+
+Accepted
+
+## Context
+
+As an AppStudio developer building functionality for the platform, I want to be able to
+easily visualize and comprehend the stability and availability of deployed systems in
+order to inform and influence future work towards improving the overall system
+reliability.
+  
+Such indication should tell us the overall uptime of AppStudio with respect to services
+under the control of AppStudio developers.
+
+AppStudio is defined to be available at a given moment if all of its components are
+reporting to be available at that given moment, and unavailable otherwise.
+
+A component is defined to be available at a given moment if all of its availability
+probes are reporting to be available at that given moment, and unavailable otherwise.
+
+A convention is required for providing the availability of a probe.
+
+Once this is in place, those indicators can be aggregated in order to report the overall
+availability of AppStudio.
+
+## Decision
+
+Probes' availability will be provided as a Prometheus metric. The metric will contain
+metric labels to allow differentiating between the different probes. When exported out
+of the cluster of origin, additional labels will be attached to the metric to mark the
+cluster in which the metric was generated.
+
+### Details
+
+The availability Prometheus metric will be computed for each component based on the exit
+status of the latest execution of the CronJobs evaluating the component's availability.
+Component owners will provide the implementation for each component's CronJobs. By
+adhering to a specific standard
+([see naming convention below](#Probes-Naming-Convention)),
+results will be aggregated into a standardized Prometheus metric to report on
+availability (i.e. component owners will not be required to provide the translation
+mechanism).
+
+It is up for each team to define what it means for its component(s) to be available.
+Fundamentally, a component should be reported as available as long as it's capable of
+providing the service it aims to provide, and unavailable otherwise.
+
+Each team will define CronJobs that will test the availability of their components.
+The Job started by each CronJob will terminate successfully if the test completes
+successfully, and will terminate with an error in case the test fails.
+
+Kube-state-metrics is a Prometheus exporter generating metrics based on the Kubernetes
+API server. It generates
+[Jobs](https://github.com/kubernetes/kube-state-metrics/blob/main/docs/job-metrics.md)
+and
+[CronJob](https://github.com/kubernetes/kube-state-metrics/blob/main/docs/cronjob-metrics.md)
+metrics that will be processed using Prometheus recording rules in order to
+generate AppStudio's availability Prometheus metric.
+
+A single set of rules will be defined globally which will apply for all CronJobs.
+
+The resulting Prometheus metric will identify each probe based on its CronJob name and
+the namespace in which the CronJob was defined.
+
+#### Probes Requirements
+
+A Job running for a probe is required to:
+
+* Evaluate the availability of the component being probed.
+* Exit with status `Failed` if the component was evaluated to be unavailable or with
+  status `Complete` if it was evaluated to be available (referring to the status field
+  of the Kubernetes Job resource).
+* Be able to exhaust its `backoffLimit` by the time it's due to run again (e.g. if the
+  cronjob is defined to run every 10 minutes and can take up to 2 minutes to execute
+  it cannot have a `backoffLimit` larger than 4).
+* Clean up all resources generated during its run.
+
+#### Probes Naming Convention
+
+To allow generating the Prometheus metric only for relevant CronJobs, probe CronJob
+names should have a standardized format:
+
+* `appstudio-probe-<probe_name>`
+
+`appstudio-probe-` being a literal to be used in order to capture only the relevant
+CronJobs, and `<probe_name>` is a variable to be translated to a label in the resulting
+Prometheus metric that will correlate the value to the individual probe or check.
+
+To allow aggregating the Prometheus availability metric per component, the namespaces
+in which the CronJobs will be created should have a standardized format:
+
+* `appstudio-probe-<component_name>`
+
+`<component_name>` is the name of the component under which probes will be aggregated.
+
+The nature and size of each service and component will dictate the number of probes it
+should have. E.g. some services may have multiple components, while some others may have
+just one. Some components may require multiple probes while others may require just one.
+
+> **_NOTE:_** The probe-name part of the CronJob name should be unique in the context of
+the CronJob's **namespace**.
+
+#### Probes Design Considerations
+
+Considerations for defining probes' CronJobs:
+
+* Where should it run in order to provide reliable evaluation?
+    * Which namespace?
+    * Which clusters?
+* What sort of permissions does it require to have?
+    * Would that be a good-enough representation of what it aims to evaluate?
+* What sort of APIs does it need to access?
+* How often should it run?
+* Does it affect performance in a reasonable manner?
+* How resources are to be cleaned up?
+    * Upon startup? completion? failure?
+    * Using another CronJob? Finalizers?
+
+## Consequences
+
+* The different teams for all AppStudio services will define the CronJobs required for
+  testing their components' availability, and will name them according to the naming
+  convention.
+* A single set of Prometheus recording rules will be defined for transforming the
+  CronJob results into availability Prometheus metric time series.
+* Existing Prometheus alerting rules should be examined and adjusted so that they do not
+  generate unnecessary alerts caused by CronJobs failures.


ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,68 @@
...
+
+As a product team, we want a composite system availability metric for both Production
+and Staging environments, so that we can make roadmap decisions with easy-to-understand
+feedback about the reliability of the system in mind.
+  
+This metric should tell us the overall up time of RHTAP with respect to services under
+the control of Red Hat.
+
+RHTAP is defined to be available at a given moment if all of its components are
+reporting to be available at that given moment.
```
eisraeli, line 19: If we want to visualize this metric as a composition of it's services, we might want to create something similar to this page:
https://www.githubstatus.com

yftacherzog: I had this one in mind as well. I don't think we should commit to it at this stage.

eisraeli: Probably not. It's just a thought I wanted to share.

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,68 @@
...
+## Decision
+
+Probes' availability will be provided as a Prometheus metric. The metric will contain
+labels to allow differentiating between the different probes. When exported out of the
+cluster of origin, additional labels will be attached to the metric to mark the cluster
+in which the metric was generated.
+
+### Delivery Method
+
+It is up for each team to define what it means for its component(s) to be available.
```
eisraeli, line 38: We might want to provide some basic guidelines to support the definition of "availability".

eisraeli:
Very informative.
LGTM.

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,68 @@
...
+
+Each team will define CronJobs that will test the availability of their components.
+The Job started by each CronJob will terminate successfully if the test complete
+successfully, and will terminate with an error in case the test fails.
```
ifireball, line 42: So this means we don't get anything until all teams manage to write some probes? Perhaps we need to define a fall-back to existing passive data such as the Prometehus "up" metrics?

Omeramsc (approved):
LGTM

ADR/0027-availability-probe-framework.md
yftacherzog: I added something. Have a look.

ADR/0027-availability-probe-framework.md
yftacherzog: I suppose we can do something along those lines for the time being, but I don't think it should go inside the ADR.

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,70 @@
...
+cluster of origin, additional labels will be attached to the metric to mark the cluster
+in which the metric was generated.
+
+### Delivery Method
+
+It is up for each team to define what it means for its component(s) to be available.
+Fundamentally, a component should be reported as available as long as it's capable of
+providing the service it aims to provide, and unavailable otherwise.
+
+Each team will define CronJobs that will test the availability of their components.
```
Michkov, line 42: Let's say that we have 30 components. So we will endup with 30 Cronjobs which will create job/pod every eg. minute?

When it's already decided (in decision section) that availablity probes are Prometheus metrics, and since prometheus is already configurated to aggregate metrics from each component why we don't have just one cronjob/service which would check all services and provide aggregated result?

yftacherzog: > and since prometheus is already configurated to aggregate metrics from each component

What are you referring to here?

The approach of having a single check that will examine RHTAP's availability end to end is something that was considered during [refinement](https://docs.google.com/document/d/1ar8PJX4MWmw7fpEgB4cxK7d8lXOpkojdTKrJPsok1A4/edit?usp=sharing), and eventually it was decided, at least for now, to go on something closer to blackbox monitoring. The cronjob approach is now suggested as a way to simplify the requirement for the individual teams so they don't have to start writing exporters.

I would say, that for a given service, if it's easier and granular enough to have a single job that will check the availability of multiple components within it, then I think that should be good enough.

> Let's say that we have 30 components. So we will endup with 30 Cronjobs which will create job/pod every eg. minute?

Yes. The frequency is something to be tweaked based on how much time it takes the job to complete and how frequent we want the status to be reported, given the resources the test consumes, but yes, that's the suggested idea.

yftacherzog: That was not the intention. The idea is for the cronjob to run some sort of checks against the component that will evaluate its availability (e.g. try to do something that the component knows to provide and report if it was able to do it or not).

Do you have in mind a case in which a component/service already reports metrics that can be translated into an availability probe (up/down indicator)?

Michkov: That confuses me, in decision section is that availability is provided by Prometheus metric. But now I can see that we should use CronJob which can receive availability even by different way.

yftacherzog: The idea of the framework is to let teams write cronjobs and have the o11y team translate them to metrics,  instead of having the teams writing test services that export metrics + service monitors to collect those metrics.

The team writes cronjob(s) that test the availability of their components (e.g. trying to trigger something within the component and verifying it was completed successfully).

The o11y team translates the results (success/failure) of all relevant cronjob runs and convert that to the availability metric.

So, the availability is provided by a Prometheus metric, and to deliver that the different teams need to come up with cronjobs to test the availability of their components, and the o11y will come up with means to transform those results in to the required metric.

Let me know if that makes sense, and also let me know if you can think of a way to make it clearer within the ADR itself.

yftacherzog:
@eisraeli @amisstea I added the points mentioned on our call.

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,99 @@
...
+Prometheus recording rules in order to generate RHTAP's availability metric.
+
+The RHTAP Observability team will construct the necessary rules.
+
+The resulting metric will identify each probe based on its CronJob name and the
+namespace in which the CronJob was defined:
+
+* The name of the CronJob will be translated to a label denoting the name of the probe.
+* The namespace in which the CronJob is defined will be translated to a label denoting
+  the probe's namesapce.
```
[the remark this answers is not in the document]
yftacherzog: done

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,99 @@
...
+  of the Kubernetes Job resource).
+* Be able to exhaust its `backoffLimit` by the time it's due to run again (e.g. if the
+  cronjob is defined to run every 10 minutes and can take up to 2 minutes to execute
+  it cannot have a `backoffLimit` larger than 4).
+* Clean up all resources generated during its run.
+
+Considerations for defining probes' CronJobs:
+
+* Where should it run in order to provide reliable evaluation?
+    * Which namesapce?
```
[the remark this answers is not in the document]
yftacherzog: done

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,99 @@
...
+    * Which cluster?
+* What sort of permissions does it require to have?
+    * Would that be a good-enough representation of what it's aim to evaluate?
```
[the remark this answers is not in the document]
yftacherzog: done

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,99 @@
...
+* Evaluate the availability of the component being probed.
+* Exit with status `Failed` if the component was evaluated to be unavailable or with
+  status `Complete` if it was evaluated to be available (referring to the status field
+  of the Kubernetes Job resource).
+* Be able to exhaust its `backoffLimit` by the time it's due to run again (e.g. if the
+  cronjob is defined to run every 10 minutes and can take up to 2 minutes to execute
+  it cannot have a `backoffLimit` larger than 4).
+* Clean up all resources generated during its run.
+
+Considerations for defining probes' CronJobs:
```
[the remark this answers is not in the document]
yftacherzog: Added. Have a look and let me know your thoughts.

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,102 @@
...
+## Context
+
+As a product team, we want a composite system availability metric for both Production
+and Staging environments, so that we can make roadmap decisions with easy-to-understand
+feedback about the reliability of the system in mind.
+  
+This metric should tell us the overall up time of RHTAP with respect to services under
+the control of Red Hat.
+
+RHTAP is defined to be available at a given moment if all of its components are
```
arewm, line 18: `s/RHTAP/AppStudio/g`

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Status
+
+Accepted
+
+## Context
+
+As a product team, we want a composite system availability metric for both Production
+and Staging environments, so that we can make roadmap decisions with easy-to-understand
+feedback about the reliability of the system in mind.
```
arewm, line 13: This might not make as much sense within the context of AppStudio itself. Maybe something like 

```suggestion
As an AppStudio developer building functionality for the platform, I want to be able to easily
visualize and comprehend the stability and availability of deployed systems in order to inform
and influence future work towards improving the overall system reliability.
```

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,102 @@
...
+  
+This metric should tell us the overall up time of RHTAP with respect to services under
```
arewm, line 15: Depending on how the previous sentence is worded, we should be more explicit here about what "this metric" is.

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,102 @@
...
+the control of Red Hat.
```
arewm, line 16: I don't think we need to word things in relation to being under control by a single entity. Can you generalize this?

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,102 @@
...
+have a standardized format:
+
+* `rhtap-probe-<probe_name>`
+
+## Consequences
+
+* The different teams will define the CronJobs required for testing their components
+  availability, and will name them according to the naming convention.
+* The RHTAP Observability team will define the Prometheus recording rules required for
+  transforming the CronJob results into availability metric time series.
```
arewm, line 102: You can generalize this. The fact that the observability team is doing the work to define the Prometheus recording rules is an implementation detail and doesn't need to be in an ADR. _If_ you want to include a similar detail here, it might be better to just state that by creating consistency in the status output, consistency can also be achieved in creating Prometheus recording rules to easily derive availability metrics.

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,102 @@
...
+    * Using another CronJob? Finalizers?
+
+#### Naming Convention
+
+To allow generating the metric only for relevant CronJobs, probe CronJob names should
+have a standardized format:
+
+* `rhtap-probe-<probe_name>`
+
+## Consequences
```
arewm, line 97: I don't see any comment in the document about a criteria for deciding which services will implement this functionality. Is it all services within AppStudio? Are there criteria that need to be met before a service is required to create this? Is it "opt-in" if a service wants to have presence in some dashboard?

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,102 @@
...
+* How resources are to be cleaned up?
+    * Upon startup? completion? failure?
+    * Using another CronJob? Finalizers?
+
+#### Naming Convention
+
+To allow generating the metric only for relevant CronJobs, probe CronJob names should
+have a standardized format:
+
+* `rhtap-probe-<probe_name>`
```
arewm, line 95: This should also not include `rhtap`. Do you expect `<probe_name>` to have a value corresponding to some service which owns the probe?

Why is a CronJob name standardization required for the metric? Which metric are you talking about specifically? The one in Prometheus?

I know you have this above, but I think I am just looking for better connection between them:

> * The name of the CronJob will be translated to a label denoting the name of the probe.

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,102 @@
...
+    * Which namespace?
+    * Which cluster?
+* What sort of permissions does it require to have?
+    * Would that be a good-enough representation of what it aims to evaluate?
+* What sort of APIs does it need to access?
+* How often should it run?
+* Does it affect performance in a reasonable manner?
+* How resources are to be cleaned up?
+    * Upon startup? completion? failure?
+    * Using another CronJob? Finalizers?
```
arewm, line 88: I am confused by these considerations. Are they supposed to be considered for implementation or are they relevant for the decision/convention being laid out in this ADR? These may be important, but are they important for the current context?

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,102 @@
...
+providing the service it aims to provide, and unavailable otherwise.
+
+Each team will define CronJobs that will test the availability of their components.
+The Job started by each CronJob will terminate successfully if the test complete
+successfully, and will terminate with an error in case the test fails.
+
+Kube-state-metrics generates Jobs and CronJob metrics that can be processed using
+Prometheus recording rules in order to generate RHTAP's availability metric.
+
+The RHTAP Observability team will construct the necessary rules.
```
arewm, line 54: Do we need to call out a specific team here (and in other places in the document if I missed them)? I am not sure if we really call out teams in ADRs, instead aggregations/associations seem to be made based on service boundaries. I know that there isn't a service boundary for o11y, but is there some better association?

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,102 @@
...
+probes are reporting to be available at that given moment.
+
+A convention is required for providing the availability of a probe.
+
+Once this is in place, those metrics can be aggregated in order to report the overall
+availability of RHTAP.
+
+## Decision
+
+Probes' availability will be provided as a Prometheus metric. The metric will contain
```
arewm, line 31: Which probes' availability? Each components'?

arewm (changes requested):
I feel like there are a lot of references with words like availability and metric. Are all uses of these words consistent? Can we define contexts for these words in the relevant places? 

If there is reuse of a common word for two applications, we can try to draw attention to the different applications of the words using techniques like bolding the word groups.

ADR/0027-availability-probe-framework.md
stuartwdouglas: Won't these cron jobs be fairly expensive, e.g. in the JVM Build Service case the only really accurate way to tell if it is working as expected is to actually build something. I would image the other components are similar, in that the only way to actually get a reliable metric is to perform the actual action for real.

ADR/0027-availability-probe-framework.md
Michkov, line 95: I would say that it should be better to use labels instead of names -> for caching/filtering options.

ADR/0027-availability-probe-framework.md
yftacherzog: @stuartwdouglas if that is the only way to tell the service is available then I think we should do it. It might be that the question in such case should be how frequent does it makes sense to run such check.

yftacherzog:
> I feel like there are a lot of references with words like availability and metric. Are all uses of these words consistent? Can we define contexts for these words in the relevant places?
> 
> If there is reuse of a common word for two applications, we can try to draw attention to the different applications of the words using techniques like bolding the word groups.

@arewm I'll look into that.
`metric` for sure often have multiple meanings in the context of monitoring. I'll try to see if there are points here which might be confusing in that sense.
Regarding `availability`, we state that it's up for the individual teams to define what available means for what they report availability for.

ADR/0027-availability-probe-framework.md
stuartwdouglas: I am definitely not against it. I just thought it was something that should be called out.

ADR/0027-availability-probe-framework.md
yftacherzog: I'm somewhat ashamed to say that I can't say why this should be AppStudio and not RHTAP. What's the reason this does not define RHTAP's availability?

ADR/0027-availability-probe-framework.md
ifireball: Hopefully a single, occasional run of any given service would be insignificant next to the amount of times it is running for our users/customers.

ADR/0027-availability-probe-framework.md
yftacherzog: > Do you expect <probe_name> to have a value corresponding to some service which owns the probe?

Yes. But it might be that they have more than one probe so it's not 1:1 matching.

> Why is a CronJob name standardization required for the metric? Which metric are you talking about specifically? The one in Prometheus?

Yes. The availability metric we plan to generate. We will have to somehow tell cronjobs created for checking availability vs. other cronjobs. In theory, Kubernetes labels can also be used here, but for Prometheus to collect metrics correlating jobs and labels to be collected, Prometheus configs need to be changed, and we don't have control over the configs of that instance that collects those metrics.

A naming convention is a rather straightforward way to maintain this I think.

> I know you have this above, but I think I am just looking for better connection between them:
>
> The name of the CronJob will be translated to a label denoting the name of the probe.

I'll look into that.

ADR/0027-availability-probe-framework.md
yftacherzog: Those are considerations to be taken into account when implementing an individual CronJob.

@amisstea would you like to comment on the need for this?

ADR/0027-availability-probe-framework.md
yftacherzog: Point taken. I'll try to generalize this throughout the document in a way that will convey to the teams that they only need to worry about the cronjob without calling out who's going to take care of the rest.

ADR/0027-availability-probe-framework.md
yftacherzog: I agree. We have some limitations that do not allow us going that way: The Prometheus instance collecting the job metrics does not necessarily collect labels used for cronjobs (for cardinality reasons, the specific labels need be whitelisted inside the Prometheus configs, which we're not the owners for).

ADR/0027-availability-probe-framework.md
yftacherzog: Yes

yftacherzog:
@arewm I tried amending the text according to all your comments. Let me know your thoughts.

ADR/0027-availability-probe-framework.md
stuartwdouglas: I think it is probably ok as is

ADR/0027-availability-probe-framework.md
amisstea: I do think there's a need for this info but perhaps this ADR is not the best place. It reads like a contribution guide so I think we can instead include it with the sample probe we're planning.

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,125 @@
...
+
+To allow aggregating the Prometheus availability metric per component, the namespaces
+in which the CronJobs will be created should have a standardized format:
+
+* `appstudio-probe-<component_name>`
+
+`<component_name>` is the name of the component under which probes will be aggregated.
+
+The nature and size of each service and component will dictate the number of probes it
+should have and how they should be aggregated. E.g. some service may have multiple
```
amisstea, line 115:
```suggestion
should have. E.g. some service may have multiple
```
It shouldn't change _how_ they get aggregated. The planned `PrometheusRules` should be capable of aggregation for a component that uses 100 probes in the same way as if there is just 1 probe.

yftacherzog: done

amisstea (approved):

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,125 @@
...
+As an AppStudio developer building functionality for the platform, I want to be able to
+easily visualize and comprehend the stability and availability of deployed systems in
+order to inform and influence future work towards improving the overall system
+reliability.
+  
+Such indication should tell us the overall up time of AppStudio with respect to services
+under the control of AppStudio developers.
+
+AppStudio is defined to be available at a given moment if all of its components are
+reporting to be available at that given moment.
```
arewm, line 20: Are we only considering available/unavailable or is there also an intermediate degraded state? What does it mean for AppStudio to not be available?

ADR/0027-availability-probe-framework.md
arewm: Above, you state:

> A component is defined to be available at a given moment if all of its availability
probes are reporting to be available at that given moment.
> 
> A convention is required for providing the availability of a probe.

This makes me think that there may be more than one probe per component. Are we trying to have per-probe availability or per-component? Is it valid for a single component to have multiple probes?

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,125 @@
...
+
+A convention is required for providing the availability of a probe.
+
+Once this is in place, those indicators can be aggregated in order to report the overall
+availability of AppStudio.
+
+## Decision
+
+Probes' availability will be provided as a Prometheus metric. The metric will contain
+metric labels to allow differentiating between the different probes. When exported out
```
arewm, line 33: If there are multiple probes per component, you need to be explicit about these probes -- whether it is differentiating between any probe or between any given component's probe(s).

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,125 @@
...
+of the cluster of origin, additional labels will be attached to the metric to mark the
+cluster in which the metric was generated.
+
+### Delivery Method
+
+The availability Prometheus metric will be computed for each component based on the exit
+status of the latest execution of the CronJobs evaluating the component's availability.
+Developer teams will provide the CronJobs' implementation. The job results will be
```
arewm, line 41: Maybe reword this to be something like this?
```suggestion
Component owners will provide the implementation for each component's CronJobs implementation. The job results will be
```

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,125 @@
...
+translated a standardized Prometheus metric in a unified fashion (i.e. developers
+will not be required to provide the translation mechanism).
```
arewm, line 43:
```suggestion
translated to a standardized Prometheus metric in a unified fashion (i.e. developers
will not be required to provide the translation mechanism).
```

This statement doesn't seem clear enough to me. What does it mean that the translation will happen in a unified fashion? Developers won't be required to provide the translation mechanism because they are adhering to some standard? If so, then just state something along those lines -- by adhering to a specific standard (be precise about the standard, link to it below), results will be aggregated into a standardized Prometheus metric to report on availability.

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,125 @@
...
+availability of AppStudio.
+
+## Decision
+
+Probes' availability will be provided as a Prometheus metric. The metric will contain
+metric labels to allow differentiating between the different probes. When exported out
+of the cluster of origin, additional labels will be attached to the metric to mark the
+cluster in which the metric was generated.
+
+### Delivery Method
```
arewm, line 37: What do you mean by delivery method? The delivery of the availability probe? By following a standard pattern which will enable automatic ingestion into a higher level availability? This section might be better named something else as I don't see any clarification of this term in the first couple paragraphs of the section.

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,125 @@
...
+
+The availability Prometheus metric will be computed for each component based on the exit
+status of the latest execution of the CronJobs evaluating the component's availability.
+Developer teams will provide the CronJobs' implementation. The job results will be
+translated a standardized Prometheus metric in a unified fashion (i.e. developers
+will not be required to provide the translation mechanism).
+
+It is up for each team to define what it means for its component(s) to be available.
+Fundamentally, a component should be reported as available as long as it's capable of
+providing the service it aims to provide, and unavailable otherwise.
```
arewm, line 47: Relating to a comment at the beginning -- again, it is either available or not. We are not considering degraded states? This is _more_ okay for individual components' availability than I think it is for AppStudio as a whole.

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,125 @@
...
+
+Each team will define CronJobs that will test the availability of their components.
+The Job started by each CronJob will terminate successfully if the test complete
+successfully, and will terminate with an error in case the test fails.
```
arewm, line 51: Will this have an effect of raising other alerts within the cluster if there are a lot of failing jobs? If so, is that a problem? It might be more in a self-managed version than in a hosted one.

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,125 @@
...
+generate AppStudio's availability Prometheus metric.
+
+A single set of rules will be defined globally which will apply for all CronJobs.
+
+The resulting Prometheus metric will identify each probe based on its CronJob name and the namespace in which the CronJob was defined:
+
+* The name of the CronJob will be translated to a Prometheus metric label denoting the
+  name of the probe.
+* The namespace in which the CronJob is defined will be translated to a Prometheus
+  metric label denoting the probe's namespace.
```
arewm, line 67: This feels like an implementation detail. Is it relevant to the architectural record? It may be able to be linked together if you word it in a way of the auto-conversion strategy.

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,125 @@
...
+* What sort of permissions does it require to have?
+    * Would that be a good-enough representation of what it aims to evaluate?
+* What sort of APIs does it need to access?
+* How often should it run?
+* Does it affect performance in a reasonable manner?
+* How resources are to be cleaned up?
+    * Upon startup? completion? failure?
+    * Using another CronJob? Finalizers?
+
+#### Naming Convention
```
arewm, line 96: This _feels_ more important than some other content in the decision and that makes me want to see it higher up in the document but I'll defer to you.

ADR/0027-availability-probe-framework.md
arewm: Are there thoughts on this? Namely, can we define what a component is that would require the implementation by code owners vs. turning on "voluntarily?"

arewm (changes requested):

ADR/0027-availability-probe-framework.md
yftacherzog: > Are we only considering available/unavailable or is there also an intermediate degraded state?

We're only considering available/unavailable, but there will be visibility into the components causing it to be unavailable.
This is per the feature refinement process: https://docs.google.com/document/d/1ar8PJX4MWmw7fpEgB4cxK7d8lXOpkojdTKrJPsok1A4/edit?usp=sharing

> What does it mean for AppStudio to not be available?

it means that there's at least one component which reports to be unavailable. In reality, it could mean any level of (un)availability - anything between dead to almost fully working.

ADR/0027-availability-probe-framework.md
yftacherzog: Interesting! There's one alert which can be affected by this, which reports on pods staying in failed state for more than 15 minutes. It might be that we'll have to tweak that alert to exclude the namespaces on the naming convention.

ADR/0027-availability-probe-framework.md
yftacherzog: > Is it valid for a single component to have multiple probes?
It's valid. Team should create as many probes as they see fit.

ADR/0027-availability-probe-framework.md
yftacherzog: done

ADR/0027-availability-probe-framework.md
yftacherzog: I renamed the section.

ADR/0027-availability-probe-framework.md
yftacherzog: I removed it now.

ADR/0027-availability-probe-framework.md
arewm: This is likely counterintuitive for those that are familiar with availability metrics, so we should ensure that we clarify this point. I added degraded to out of scope in the document.

ADR/0027-availability-probe-framework.md
arewm: I understand that you want to keep the paragraph brief, but I think that we need to ensure that the language is precise when it comes to different probes. Probes are not always on a per component basis, so if you are referencing the probes for a specific component, it needs to be plural. Also when referencing the aggregated availability, it isn't if a specific component reports being unavailable, but a specific probe.

ADR/0027-availability-probe-framework.md
arewm: If this is the case, then we can just record that possibility as a consequence in the document.

ADR/0027-availability-probe-framework.md
yftacherzog: I can't say I understand. Should we say here that components might have one or more probes?

ADR/0027-availability-probe-framework.md
yftacherzog: done

ADR/0027-availability-probe-framework.md
arewm: Yes, that is what you are telling me. I am just trying to say that we need to ensure that we are clearly communicating this relationship and that we need to ensure that we are properly considering all cases where assumptions might fall apart or be unnecessary.

For example, later when you talk about components, you cannot talk about a single probe per component. Also when discussing the probe names, what might it look like for a service which has multiple components and some components have multiple probes? Trying to differentiate between a unique probe name and a unique component name doesn't fit my mental model. Instead, would it just make sense to describe only as a probe name and indicate that some use cases might see a single probe being able to represent the status for a single component? Would there be any ability to see an aggregation of probe metrics on a component level instead of on an AppStudio level? If so, is there any abstraction that can be provided? Should the probe naming convention be scoped to have the name include the component and a probe identifier?

yftacherzog: We're saying this:
```
The nature and size of each service and component will dictate the number of probes it
should have. E.g. some services may have multiple components, while some others may have
just one. Some components may require multiple probes while others may require just one.
```

Can you point me to places where we have probe mentioned in a way that may suggest different probe-component cardinality?

>  what might it look like for a service which has multiple components and some components have multiple probes?

It will have namespaces defined per component, under each of them there will be cronjobs defined per probe.

> Would there be any ability to see an aggregation of probe metrics on a component level instead of on an AppStudio level?

That's the idea, and it's also being mentioned: `To allow aggregating the Prometheus availability metric per component...`.

> If so, is there any abstraction that can be provided?

Creating the cronjobs for a single component on a single namespace is the means to enable that.

> Should the probe naming convention be scoped to have the name include the component and a probe identifier?

I didn't want to make the teams have to coordinate probe names between the entire appstudio. For that reason, the requirement is to have the identifier of the probe within the cronjob name, and have that multiplexed by the namespace identifying the component.

yftacherzog: @arewm let me know your thoughts.

yftacherzog:
@ralphbean @adambkaplan
Can I have your reviews on this PR?
Andrew was reviewing it and we had some discussions going, but I'm hoping to have the PR approved by someone outside the o11y team even before Andrew is back.

adambkaplan (commented):
My first question regarding this ADR - why can't ArgoCD do this for us? Is this a missing feature, or something that we currently aren't doing that makes it harder for ArgoCD to figure out that our applications are synced AND operational?

More to the point - would we ask a public user to implement the same kinds of availability probes for their workloads running on RHTAP?

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,132 @@
...
+The Job started by each CronJob will terminate successfully if the test completes
+successfully, and will terminate with an error in case the test fails.
+
+Kube-state-metrics is a Prometheus exporter generating metrics based on the Kubernetes
+API server. It generates
+[Jobs](https://github.com/kubernetes/kube-state-metrics/blob/main/docs/job-metrics.md)
+and
+[CronJob](https://github.com/kubernetes/kube-state-metrics/blob/main/docs/cronjob-metrics.md)
+metrics that will be processed using Prometheus recording rules in order to
+generate AppStudio's availability Prometheus metric.
```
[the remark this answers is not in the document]
yftacherzog: Indeed. The work here will be about converting those existing metrics into our probe format.

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,132 @@
...
+* `appstudio-probe-<probe_name>`
+
+`appstudio-probe-` being a literal to be used in order to capture only the relevant
+CronJobs, and `<probe_name>` is a variable to be translated to a label in the resulting
+Prometheus metric that will correlate the value to the individual probe or check.
+
+To allow aggregating the Prometheus availability metric per component, the namespaces
+in which the CronJobs will be created should have a standardized format:
+
+* `appstudio-probe-<component_name>`
```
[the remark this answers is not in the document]
yftacherzog: We plan on providing an example probe that will require setting up some resources (service account, role binding, etc.) so it will serve as an example for the teams to follow so they don't need to come up with such solution on their own.

yftacherzog:
@adambkaplan

> My first question regarding this ADR - why can't ArgoCD do this for us? Is this a missing feature, or something that we currently aren't doing that makes it harder for ArgoCD to figure out that our applications are synced AND operational?

the main aspect here is granularity I suppose. The idea is to provide a mechanism to allow us to probe components from the outside and verify that a certain behavior is as expected. I'm not sure this is something we can do with ArgoCD.

> More to the point - would we ask a public user to implement the same kinds of availability probes for their workloads running on RHTAP?

Those probes are just for the availability of the services that constitute RHTAP. i.e. our own services.

avi-biton (approved):
LGTM

Roming22 (approved):

ADR/0027-availability-probe-framework.md
```diff
@@ -0,0 +1,131 @@
...
+Accepted
+
+## Context
+
+As an AppStudio developer building functionality for the platform, I want to be able to
+easily visualize and comprehend the stability and availability of deployed systems in
+order to inform and influence future work towards improving the overall system
+reliability.
+  
+Such indication should tell us the overall up time of AppStudio with respect to services
```
scoheb, line 16:
```suggestion
Such indication should tell us the overall uptime of AppStudio with respect to services
```

yftacherzog: done

yftacherzog:
Hi @arewm, can we merge this?

arewm (approved):
LGTM, no blocking comments. Any further clarification needed in time can be contributed back as it is discovered (to this or future ADRs).

PR #120, opened 2023-07-24
docs(STONEINTG-496): remove the link for unused tasks

hongweiliu17:
* Since sast-go and sast-java-sec-check are not used and are removed from
  build-definistion https://github.com/redhat-appstudio/build-definitions/pull/488,
  so let's remove the link of the tasks from security-tools doc.

Signed-off-by: Hongwei Liu <hongliu@redhat.com>

### DIFF
--- a/tools/security-tools.MD
+++ b/tools/security-tools.MD
@@ -39,12 +39,8 @@ _Note: [clair-in-ci](https://quay.io/repository/redhat-appstudio/clair-in-ci) is
 ### SAST Tools
 
 **_gosec_** - https://github.com/securego/gosec
- 
-_Note: AppStudio uses gosec to perform static analysis on our Golang projects. A Tekton Task is available that can be used to run gosec in your own Pipelines [here](https://github.com/redhat-appstudio/build-definitions/blob/main/task/sast-go/)._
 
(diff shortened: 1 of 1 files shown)


hongweiliu17:
@dirgim @MartinBasti @sonam1412 PTAL

dirgim (approved):
LGTM

jsztuka (approved):
lgtm

hongweiliu17:
@arewm @mattreid Could you help review this PR? Thanks!

mattreid:
My initial reaction is I wasn't expecting to still see them listed here at all, but I don't have much context on this Security Guild page, so continuing to list them may be fine. From the description at the top, it sounds like anything listed in this page has our endorsement, so listing them without any context could lead readers to think that they're ready to use. Another option could be to mention that they were previously included, but removed in favor of snyk due to performance issues that needed more investigation, that way we have a record that they were deemed acceptable at one point, and with more work, could maybe be an option for other teams. 

Andrew is on vacation starting today for 2 weeks, so if you'd like another review, @ralphbean might be able to take a look for a more technical opinion.

ralphbean (approved):

hongweiliu17:
Thanks for the comment and review.

PR #128, opened 2023-08-14
STONEBLD-1573 Document SharedRegistry

rnc:
cc @stuartwdouglas

### DIFF
--- a/book/jvm-build-service.md
+++ b/book/jvm-build-service.md
@@ -32,6 +32,10 @@ Once the `ArtifactBuild` object has been created JBS will then attempt to try an
 
 Once the source code location has been discovered a [`DependencyBuild`](#DependencyBuild) object is created. There are generally less `DependencyBuild` objects than there are `ArtifactBuild`, as multiple artifacts can come from the same build. The controller will then try and build the artifact, first it will run a build discovery pipeline, that attempts to determine possible ways of building the artifact. Once discovery is complete it uses this information to attempt to build the artifact in a trial and error manner.
 
+The dependency discovery pipeline can check both configured shared repositories and the main repository
+(for example a Quay.io repository) for pre-existing builds. If a prior build is found, the pipeline will shortcut
+to avoid building the artifact and will instead refer to the found build artifacts instead.
+
 If a build is successful the results are stored in a container image and the state is marked as complete, otherwise it is marked as failed and manual effort is required to fix the build.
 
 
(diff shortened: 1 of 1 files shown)


book/jvm-build-service.md
```diff
@@ -163,6 +168,27 @@ spec:
   enableRebuilds: "true"
 ```
 
+The `JBSConfig` is also where shared registries are configured. A shared registry is one that may be shared by many
+users to avoid multiple rebuilds of the same artifact. This assumes the first user has configured their registry to
+deploy to a shared registry. For example:
```
arewm, line 173:
```suggestion
In order to avoid multiple rebuilds of the same artifact, a user may configure a shared registry with a `JBSConfig`.
Once a user has configured a registry, the CR can be used to point to the shared registry. For example:
```

arewm: I assume that creating a registry (or describing the process/what it is) is out of scope for this document. Is there any other information that could be added here for clarity?

book/jvm-build-service.md
```diff
@@ -32,6 +32,10 @@ Once the `ArtifactBuild` object has been created JBS will then attempt to try an
 
 Once the source code location has been discovered a [`DependencyBuild`](#DependencyBuild) object is created. There are generally less `DependencyBuild` objects than there are `ArtifactBuild`, as multiple artifacts can come from the same build. The controller will then try and build the artifact, first it will run a build discovery pipeline, that attempts to determine possible ways of building the artifact. Once discovery is complete it uses this information to attempt to build the artifact in a trial and error manner.
 
+Note that the dependency discovery pipeline can check both configured shared repositories and the main repository
+(which may be e.g. a Quay.io repository) for a pre-existing build ; if found it will shortcut the pipeline avoiding the
+build and refer to those RebuiltArtifacts instead.
```
arewm, line 37:
```suggestion
The dependency discovery pipeline can check both configured shared repositories and the main repository
(for example a Quay.io repository) for a pre-existing builds. If a previous build is found, the pipeline will shortcut to avoiding
building the artifact and will instead refer to the found `RebuiltArtifact`s instead.
```

arewm: We introduce the concept of `RebuiltArtifact` later after its first use.

arewm (changes requested):

book/jvm-build-service.md
rnc: I have rephrased it slightly to try avoid the reference.

book/jvm-build-service.md
rnc: Yes, I agree that other information on JBSConfig is missing but I am slightly loath to shoehorn that into this PR as I think its a separate piece of work.  I included the summary line
```
This assumes that another user has configured their registry to deploy builds to `my-team-owner`. For example:
```

arewm (approved):

ralphbean (approved):

PR #131, opened 2023-08-25
Add initial docs on multi arch

stuartwdouglas:


book/multi-arch-controller.md
```diff
@@ -0,0 +1,114 @@
+# Multi Arch Controller
```
arewm, line 1: Would we gain any benefit if we abstract this controller away from builds specifically? In essence, it is a remote environment provisioner. In the implementation of this controller, we are exposing the fact that this is implemented for builds because of the use of `build-dir`. We are also exposing the fact that this is created for the creation of environments with multiple architectures.

This request is effectively one to implement against the Environment API, but I suspect that this would also prevent the use of the novel scheduling approach using the secrets. Maybe the Environment API is _too_ abstracted, but the approach can be slightly abstracted (i.e. renaming `build-dir` to `user-dir`).

book/multi-arch-controller.md
```diff
@@ -0,0 +1,114 @@
...
+
+Once the secret has been created then the original `TaskRun` will start, and can use the new user on the remote host to perform the build. This approach means that every build is run under a different user, and only low privilege ephemeral SSH keys end up in the user's namespace.
+
+Which hosts are allocated with which tasks are tracked by adding additional labels onto the `TaskRun` object.
+
+If the provisioning task fails then a label is added to the `TaskRun` indicating that it has failed and allocation is re-tried, excluding all hosts in the failed list. If all of them fail then the error secret is created.
+
+### Clean Up
+
+When the `TaskRun` is completed then the user is removed from the remote host via a cleanup task.
```
arewm, line 93: How is the cleanup task scheduled? Is there any way for the task to fail to run (i.e. catastrophic crashes)?

book/multi-arch-controller.md
stuartwdouglas: I can rename the directory, it can be used for any task that requires a virtual machine to be provisioned, its not just for builds.

book/multi-arch-controller.md
```diff
@@ -0,0 +1,114 @@
...
+## Overview
+
+The Multi Arch Controller is a Kubernetes Controller that can provision virtual machines for Tekton `TaskRun` instances. This allows these TaskRuns to SSH into the provided virtual machines to build images in architectures other than amd64.
+
+## The TaskRun Contract
+
+The Multi Arch Controller has no understanding of the build pipeline logic. It only understands how to provision hosts and allocate SSH keys. For the Multi Arch Controller to act on a `TaskRun` it must meet the following conditions:
+
+- It must have the `build.appstudio.redhat.com/multi-arch-required` label
+- It must mount a secret with the name `multi-arch-ssh-$(context.taskRun.name)`
```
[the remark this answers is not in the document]
stuartwdouglas: I could call the secret something else if you want to get away from the multi-arch naming, this secret and label is part of the Task definition. Labels are copied from the task into the `TaskRun`

book/multi-arch-controller.md
stuartwdouglas: Kube finalizer. The task can definitely fail, e.g. if the VM had shutdown or was off the network, or something was wrong with our kube cluster.

book/multi-arch-controller.md
```diff
@@ -0,0 +1,114 @@
...
+### Clean Up
+
+When the `TaskRun` is completed then the user is removed from the remote host via a cleanup task.
+
+
+## VM Host Requirements
+
+In order to work with the multi arch controller remote hosts need to meet the following requirements:
+
+- Have podman installed
```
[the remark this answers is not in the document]
stuartwdouglas: Just a requirement of the current build task, not the multi arch controller itself.

book/multi-arch-controller.md
```diff
@@ -0,0 +1,114 @@
...
+- `host` the host name or IP to connect to, including the username in the form `user@host`
+- `build-dir` the working directory to perform the build in
+
+Note that these SSH credentials are only valid while the `TaskRun` is executing.
+
+If the operator is unsuccessful then it will still create the secret, however it will only have a single item called `error` with the error message. The task is expected to echo this to the logs and then fail if this error is present.
+
+The reason why this works is that the `TaskRun` will create a pod that references a non-existent secret (e.g. if the `TaskRun` was called `devfile-sample-t7rd7-build-container-arm64` then the resulting pod will reference a secret called `multi-arch-ssh-devfile-sample-t7rd7-build-container-arm64`). As this secret does not exist yet, the pod will not start but instead Kubernetes will wait until the secret is ready before running the pod. This gives the controller time to provision a SSH based host, and then create the secret with the required details. The secret is created even on failure to prevent the task hanging until timeout and to allow any error messages to be propagated to the user.
```
lcarva, line 26: Is there a scheduler timeout that may kick in prematurely?

book/multi-arch-controller.md
```diff
@@ -0,0 +1,114 @@
...
+  ibmz1.concurrency: "4"
+```
+
+Each host has a unique name, and provides different parameters that can be configured:
+
+- `address` The address or host name
+- `secret` The name of a secret in the `multi-arch-controller` namespace that contains the SSH keys for this host. These keys are never copied into the user namespace.
+- `concurrency` The number of builds that can be running on the host at any one time. Note that this is tracked by the controller rather than the host, so for example if you had a host referenced by both staging and production each environment is not aware of the other environments builds.
+- `user` The username to use to connect to the host
+- `arch` The hosts architecture
```
lcarva, line 70: In OSBS, we use the concept of platform to define the combination of a CPU architecture and an OS (just like a manifest list does). You may want to consider a similar approach here. Either as a single parameter, e.g. `platform: linux/amd64`, or two parameters, e.g. `os: linux` and `arch: amd64`.

This would also impact the "The TaskRun Contract".

book/multi-arch-controller.md
```diff
@@ -0,0 +1,114 @@
...
+- Provide a user that can successfully use `sudo` to add and delete users, and copy generated SSH keys into their home directories
+- They *should* be running RHEL, however this is not strictly speaking a technical requirement
+
+### Additional Requirements
+
+There are still unanswered questions about these hosts and additional requirements that may be placed on them.
+
+- How do we handle logs from these machines? Where will they be stored?
+- Should the 'main' user be locked down so that it can only add and remove the build users? At present the main user used by the controller has root access (not the users used by the build, but the user that creates these users).
+- Do we need any kind of intrusion detection or additional monitoring software on these VMs?
```
lcarva, line 110: Yes. Given that these VMs are long-lived, they are a prime target for embedding some kind of solarwind-like malware. @zregvart has pointed me to [this](https://youtu.be/zCznRGlsYvo). We may need something like that.

book/multi-arch-controller.md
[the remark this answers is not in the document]
lcarva: Maybe remove from this list then? I could imagine another section that states the requirements for using the host to run a build.

book/multi-arch-controller.md
```diff
@@ -0,0 +1,114 @@
...
+Which hosts are allocated with which tasks are tracked by adding additional labels onto the `TaskRun` object.
+
+If the provisioning task fails then a label is added to the `TaskRun` indicating that it has failed and allocation is re-tried, excluding all hosts in the failed list. If all of them fail then the error secret is created.
+
+### Clean Up
+
+When the `TaskRun` is completed then the user is removed from the remote host via a cleanup task.
+
+
+## VM Host Requirements
```
lcarva, line 96: Should this be just "Host Requirements"? IIUC, any host would do. It doesn't have to be a VM.

book/multi-arch-controller.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Allocation Process
+
+When a `TaskRun` is identified that meets the contract above the controller will attempt to allocate a virtual machine for them.
+
+The controller looks at the current configuration, and also looks for all running tasks with a host assigned. Using this information it picks one of the hosts at random that has the most amount of free slots. This algorithm is very simple, but works fine when all hosts are the same size. If we are mixing hosts of different sizes then we may need a better algorithm.
+
+Once the host is selected the controller runs a 'provisioning' task to prepare the host. This task currently does the following:
+
+- Create a new user specifically for this build
```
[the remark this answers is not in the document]
stuartwdouglas: It is a normal non-root user. The isolation is just system provided user isolation.

book/multi-arch-controller.md
```diff
@@ -0,0 +1,114 @@
...
+- Generate a new pair of SSH keys for this user and install them
+- Copy the private key for this user back to the cluster and create the required secret with this information
+
+Once the secret has been created then the original `TaskRun` will start, and can use the new user on the remote host to perform the build. This approach means that every build is run under a different user, and only low privilege ephemeral SSH keys end up in the user's namespace.
+
+Which hosts are allocated with which tasks are tracked by adding additional labels onto the `TaskRun` object.
+
+If the provisioning task fails then a label is added to the `TaskRun` indicating that it has failed and allocation is re-tried, excluding all hosts in the failed list. If all of them fail then the error secret is created.
```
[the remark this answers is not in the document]
stuartwdouglas: The 'failed list' is per TaskRun. There is no global 'Failed List' (which would really require CRDs to manage state). Even if a host is considered failed for one task subsequent tasks will still attempt to use it, so a minor glitch will not take it down completely. 

This could definitely be improved upon.

book/multi-arch-controller.md
arewm: These seem like much more catastrophic issues -- ones that are unlikely to be taken advantage of by a malicious entity to create a persistent entry point into our infrastructure. Is that an accurate categorization? Is there any redundancy that we can do to further protect ourselves?

book/multi-arch-controller.md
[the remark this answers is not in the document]
arewm: I think it would be beneficial but we can discuss the changes before this might go into production.

book/multi-arch-controller.md
```diff
@@ -0,0 +1,114 @@
...
+- They *should* be running RHEL, however this is not strictly speaking a technical requirement
+
+### Additional Requirements
+
+There are still unanswered questions about these hosts and additional requirements that may be placed on them.
+
+- How do we handle logs from these machines? Where will they be stored?
+- Should the 'main' user be locked down so that it can only add and remove the build users? At present the main user used by the controller has root access (not the users used by the build, but the user that creates these users).
+- Do we need any kind of intrusion detection or additional monitoring software on these VMs?
+
```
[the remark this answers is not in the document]
stuartwdouglas: In that case we will need to copy credentials I think, at least initially. 

We could attempt to pull on the cluster and then copy the contents over.

book/multi-arch-controller.md
stuartwdouglas: Good point.

At the moment this is just a string, so 'linux/amd64' would work fine, as long as we configured that name in the host config and pipeline runs. 

Not sure if this is worth changing the name to 'platform' rather than arch though?

lcarva: It would be much harder to change later :wink:

book/multi-arch-controller.md
arewm: Even as @stuartwdouglas is exploring ephemeral environments, it may still be beneficial to deploy some trusted observer to the environment to help our confidence in the trustworthiness.

book/multi-arch-controller.md
stuartwdouglas: I can't really think of any way the build could affect cleanup. The build is run under a temp user, while the cleanup is done with a higher privileged user that will completely remove the users home directory. 

I guess the user could add stuff to /tmp, but I don't think that is a big issue.

We could also try and make sure there are no long running processes that belong to this user when the cleanup happens.

book/multi-arch-controller.md
stuartwdouglas: There does not seem to be. The timeout that kicks in is the TaskRun timeout.

book/multi-arch-controller.md
stuartwdouglas: @arewm What do you think? Personally I think platform makes more sense than arch, even though at the moment we only support linux.

arewm: Platform makes sense, yes.

book/multi-platform-controller.md
```diff
@@ -0,0 +1,132 @@
...
+
+For our existing builds currently it would be possible for an admin to push a bad image, or run a bad task, however this would fail the enterprise contract.
+
+To avoid this the Multi Platform Controller is going to move to a One Time Password approach to managing SSH keys. Instead of the multi platform controller creating a secret with a SSH key, instead it will create a one time password and add this to the secret. This secret can then be exchanged exactly once for a SSH key.
+
+If an attacker reads the secret and gains access the VM the task will immediately fail as it will not be able to access the VM.
+
+#### One Time Password Service
+
+The one time password service will be a http based service that provides two endpoints, one to accept a password + OTP pair combo, and another to serve the keypair when provided with the password.
```
arewm, line 108: http or https? It should be encrypted communication.

ralphbean:
@stuartwdouglas this LGTM to merge and iterate on as we go. Anything else you want to add before merging it in?

ralphbean (approved):

arewm (approved):

PR #132, opened 2023-08-27
docs(RHTAPWATCH-476): Move RHTAP information to d/s repositories

eisraeli:
For now the logging related information is removed and will be revised.


Omeramsc (approved):
LGTM

book/logging.md
```diff
@@ -114,70 +96,14 @@ allows organizations to collect, index, search, analyze, and visualize machine-g
 data, including logs from various sources. Splunk provides powerful search capabilities,
```
arewm, line 96: Can we just strike Splunk from this documentation all together? After this PR, its only presence in this document outside of this section is the very last line.

book/logging.md
```diff
@@ -114,70 +96,14 @@ allows organizations to collect, index, search, analyze, and visualize machine-g
...
-|-----------------------------------------------|:--------------------------|
-| splunk-forwarder-rhtap-production-application | [stonesoup/production/monitoring/logging/fluentd/splunk-forwarder-rhtap-production-application](https://vault.devshift.net/ui/vault/secrets/stonesoup/show/production/monitoring/logging/fluentd/splunk-forwarder-rhtap-production-application)  |
-| splunk-forwarder-rhtap-production-audit       | [stonesoup/production/monitoring/logging/fluentd/splunk-forwarder-rhtap-production-audit](https://vault.devshift.net/ui/vault/secrets/stonesoup/show/production/monitoring/logging/fluentd/splunk-forwarder-rhtap-production-audit) |
-| rh_rhtap_staging_vpc                          | [stonesoup/production/monitoring/logging/rh_rhtap_production_vpc](https://vault.devshift.net/ui/vault/secrets/stonesoup/show/production/monitoring/logging/rh_rhtap_production_vpc) |
-
-**Notice** Secrets are also backed-up in the o11y team's BitWarden vault.
-
-### Logs retention
-Logs are being retained for 60 days.
+### CloudWatch
```
ifireball, line 99: I don't think we should say much about cloudwatch, because we're (hopefully) going to shut it down.

book/logging.md
arewm: What is the value that we are providing in this section? You are just mentioning different analytics platforms but you are not mentioning how AppStudio is integrating with them.

Deployment to Amazon/AWS is an implementation detail of AppStudio. The only other place that I see a reference to the cloud service that is _specific_ to tying us to AWS is in the pipeline-service as AWS RDS is the database back end. 

In this case, I don't see the utility of providing the information. You are removing all other Splunk information from the page, so we don't need to introduce it as a concept. We don't go into any more detail of CloudWatch here because there isn't any configuration from AppStudio itself to enable (it is done via the cloud provider).

Generic information to various monitoring and analytics platforms doesn't _feel_ interesting to me in this document. Instead, it would be beneficial to document how logging is handled by AppStudio as a whole, i.e. summarizing the strategies for determining component availability via log aggregation or if there are any custom metrics that might be enabled for people that would be deploying or how ADR-006 might relate to the platform as a whole.

arewm: Logs and observability, while related, are very different things to me. What you have described here is _observability_ in that it highlights methods that we can see the types of logs and data that are interesting to us. This largely comes down to configuration in the deployment of AppStudio and is hard to document in a generalized fashion. You _can_ still comment on the types of data that may be interesting to an observability stack and the methods by which observability can be implemented.

What might make more sense in this document, however, is to highlight the patterns _within_ AppStudio which make the eventual observing actions easier to configure/implement. That isn't to say that all of the content that you have here should necessarily be removed. If you figure out what the direction/voice of this reference to be then you can use that to figure out how much of the additional/surrounding/tangential information to also include.

arewm: Ah, I thought I had approved already. My bad!

arewm (approved):

amisstea (approved):

PR #133, opened 2023-08-29
Internal secrets

skabashnyuk:
This functionality is meant to prevent `pull` secrets from being deleted by the user.
https://issues.redhat.com/browse/DTUX-1693


mmorhun (approved):

karthikjeeyar:
LGTM

ADR/0022-secret-mgmt-for-user-workloads.md
```diff
@@ -380,6 +380,26 @@ status:
     status: "False"
     type: DataObtained
 ```
+### Internal Secrets
+
+To distinguish secrets that were provided by end-users from secrets that are generated by RHTAP's components, a label `appstudio.redhat.com/internal: true` is introduced. The presence of this label indicates that the secret is intended for internal usage, and unauthorized changes to such secrets may lead to unpredictable results.
```
arewm, line 385: This is intended to only be used for RemoteSecrets. While that is implied because of the ADR that this is in, can we be more explicit about the scope to ensure that it isn't extrapolated out of context?

ADR/0022-secret-mgmt-for-user-workloads.md
```diff
@@ -380,6 +380,26 @@ status:
...
+
+This label helps to identify secrets that are meant for internal purposes, preventing accidental or unauthorized modifications that could disrupt the functioning of RHTAP's components.
```
arewm, line 387: Can we be more explicit about how these changes will be implemented in AppStudio? Will there be changes to permissions or a controller? What are the specific actions that would be prevented?

arewm (changes requested):
Can you add a revision date to the top of this ADR based on when the change?

ADR/0022-secret-mgmt-for-user-workloads.md
```diff
@@ -380,6 +381,27 @@ status:
...
+metadata:
+    name: test-remote-secret-secret
+    labels:
+        appstudio.redhat.com/component: m-service
+        appstudio.redhat.com/application: coffee-shop
+        appstudio.redhat.com/internal: true
+spec:
+...
+```
+It's important to note that this label is not intended to alter the behavior of the controller or permissions. The primary effect is on the UI behavior, which is designed to hide these labeled RemoteSecrets from the list.
```
arewm, line 403: This will not protect workspsace Admins from deleting these internal RemoteSecrets from the CLI as they would have the ability to be destructive with `get, list, watch, create, update, patch, delete` permissions, correct?

arewm: @mmorhun, are you aware of these plans?

arewm (approved):
LGTM. These changes will help to reduce confusion and the likelihood of unintentional deletion. We are still not preventing Admins from breaking their workspaces if they really want to. (:

xbaran4 (approved):

ADR/0022-secret-mgmt-for-user-workloads.md
mmorhun: Partially. There was a discussion with Gal, but no action items requested so far. There are pros and cons that should be discussed. Anyway, this is an off topic for this ADR.

arewm: @mmorhun @skabashnyuk , would one of you be able to create an issue for that work/investigation with necessary details so that we don't lose track of the request?

arewm: We can create a Spike for the discussion and stories after that as/if needed.

gazarenkov (approved):

PR #134, opened 2023-08-29
Basic authentication secret

skabashnyuk:
TODO

### DIFF
--- a/ADR/0029-basic-authentication-secret.md
+++ b/ADR/0029-basic-authentication-secret.md
@@ -66,7 +66,7 @@ metadata:
     name: github-com-token
     namespace: myns
     annotations:
-        appstudio.redhat.com/sp.repository:   "redhat/secret, redhat/opera"
+        appstudio.redhat.com/sp.repository: "redhat/secret,redhat/opera"
     labels:
         appstudio.redhat.com/sp.host: github.com
 spec:


ADR/0029-basic-authentication-secret.md
```diff
@@ -0,0 +1,88 @@
...
+
+Users can specify resource names through the UI, which will be added as comma-separated annotation `appstudio.redhat.com/sp.repository` to the `RemoteSecret` CR. RHTAP components will consider these annotations during `RemoteSecret` searches.
+```yaml
+apiVersion: appstudio.redhat.com/v1beta1
+kind: RemoteSecret
+metadata:
+    name: github-com-token
+    namespace: myns
+    annotations:
+        appstudio.redhat.com/sp.repository:   "redhat/secret, redhat/opera"
```
xbaran4, line 69:
```suggestion
        appstudio.redhat.com/sp.repository: "redhat/secret,redhat/opera"
```
I believe these should be separated with a comma without the whitespace. If you think we should take the whitespace into consideration let me know.

ADR/0029-basic-authentication-secret.md
```diff
@@ -0,0 +1,88 @@
...
+1. Revoke the secret externally.
+2. Generate a new credential externally.
+3. Overwrite the secret in RHTAP using the UI.
+
+
+## Consequences
+
+1. **Minimal Data Exposure**: Secrets that are not actively used by any workload remain in permanent storage without being populated in etcd, reducing potential data exposure.
+2. **Audit Trail for Unused Secrets**: In the rare case where secrets remain in etcd for an extended period, there is a potential avenue to identify the intended consumers of these secrets. This can help with security audits and potential investigations.
+3. **Extended Credential Usage**: The architecture enables the usage of credentials beyond the current namespaces. This includes scenarios like namespace-backed environments, ephemeral environments, or user-provided clusters (Bring Your Own Cluster - BYOC).
```
xbaran4, line 88: Can you elaborate on `current namespaces`?

ADR/0029-basic-authentication-secret.md
```diff
@@ -0,0 +1,88 @@
...
+        appstudio.redhat.com/sp.repository: "redhat/secret,redhat/opera"
+    labels:
+        appstudio.redhat.com/sp.host: github.com
+spec:
+    secret:
+        type: kubernetes.io/basic-auth
+```
+### Rotation of Tokens/Secrets
+
+Users can rotate tokens/secrets through these steps:
```
arewm, line 78: Will users be able to see what secrets are created so that they know which ones can be revoked externally or overwritten in the UI?

arewm: This will be achieved by the UI looking for the presence of `RemoteSecret`s as well?

What does it mean to "revoke externally" -- is this via the CLI or would this also be done via the UI somehow?

arewm: It seems like I might have been confused by the reference point. So it would be external to the system/`RemoteSecret` process... but still in-line via the UI? This ADR is written from the perspective of the remote secret and not from the perspective of AppStudio as a while?

arewm: I thin we can clarify this section. Maybe something like

```
### Rotation of Tokens/Secrets

It is not possible to delete secrets added to `RemoteSecret`s. Instead, users can revoke or rotate tokens and secrets through these steps:
1. Revoke the secret with at its origination.
2. Generate a new credential with the desired service.
3. Overwrite the secret in RHTAP using the UI.
```

PR #135, opened 2023-08-30
doc(stoneintg-586): Add ADR for SEB timeout

14rcole:


ADR/0028-handling-snapshotenvironmentbinding-errors.md
```diff
@@ -0,0 +1,14 @@
...
+Date: 2023-08-31
+
+## Status
+Accepted
+
+## Context
+It is currently not possible to determine whether a SnapshotEnvironmentBinding is stuck in an unrecoverable state.  This is a major problem when deciding if an ephemeral SEB needs to be cleaned up by the integration service's SnapshotEnvironmentBinding controller.  An inability to clean up errored SEBs can overload the cluster.
+
+## Decision
+The integration service has a reconciler function that cleans up errored SnapshotEnvironmentBindings with an ErrorOccured condition.  This condition is set to 'true' by default and is set to 'false' when the environment becomes available.  The integration service will that all SEBs with a 'true' ErrorOccured condition and a LastUpdateTime of more than five minutes ago are unrecoverable and can be cleaned up.
```
dirgim, line 11: suggestion: `The integration service will that` -> `The integration service will consider that`

sonam1412:
Would it make sense to also mention the follow-up feature request [RHTAP-1530](https://issues.redhat.com/browse/RHTAP-1530) created for the more generic approach on the cleanup of the environment resources ?
We can add an additional section with footnotes like [this](https://github.com/redhat-appstudio/book/pull/47/files#diff-382d9ba4cefc530a8d22a275cfa1cc5085a1a4436fd66a8031d42b899e802de9R45)

ADR/0028-handling-snapshotenvironmentbinding-errors.md
```diff
@@ -0,0 +1,14 @@
+# 28. Handling SnapshotEnvironmentBinding Errors
+Date: 2023-08-31
+
+## Status
+Accepted
+
+## Context
+It is currently not possible to determine whether a SnapshotEnvironmentBinding is stuck in an unrecoverable state.  This is a major problem when deciding if an ephemeral SEB needs to be cleaned up by the integration service's SnapshotEnvironmentBinding controller.  An inability to clean up errored SEBs can overload the cluster.
```
arewm, line 8:
```suggestion
It is currently not possible to determine whether a SnapshotEnvironmentBinding (SEB) is stuck in an unrecoverable state.  This is a major problem when deciding if an ephemeral SEB needs to be cleaned up by the integration service's SnapshotEnvironmentBinding controller.  An inability to clean up errored SEBs can overload the cluster.
```

ADR/0028-handling-snapshotenvironmentbinding-errors.md
arewm, line 11: Is there any data to back up the time frame of five minutes? How easily is this configurable? Would users have any control over this? Five minutes seems more reasonable for an Environment like a namespace than it does for one that is like a cluster. Is this configurable for different types of Environments?

ADR/0028-handling-snapshotenvironmentbinding-errors.md
```diff
@@ -0,0 +1,14 @@
...
+
+## Decision
+The integration service has a reconciler function that cleans up errored SnapshotEnvironmentBindings with an ErrorOccured condition.  This condition is set to 'true' by default and is set to 'false' when the environment becomes available.  The integration service will that all SEBs with a 'true' ErrorOccured condition and a LastUpdateTime of more than five minutes ago are unrecoverable and can be cleaned up.
+
+## Consequence
+- SnapshotEnvironmentBindings will be cleaned up after five minutes if they are not ready.  Environments that take a long time to provision may be cleaned up erroneously.  The timeout threshold can be adjusted if need be to accomodate this.
```
arewm, line 14: Can you clarify how the threshold can be adjusted?

arewm (changes requested):

ADR/0028-handling-snapshotenvironmentbinding-errors.md
14rcole: The timeout is hard-coded in one place in the integration service.  So it's easy enough to change with a PR but can't be adjusted by users.  Do you think this should be a user-adjustable feature?  If so, what would you suggest?  My first thought would be to add the parameter to a configmap but I don't know if users have access to any.

dirgim: The initial suggestion from the GitOps team was to set the timeout to 3 minutes as that is the expected time in which any transient error state should be resolved. We opted for a slightly more generous timeout of 5 minutes during the planning for the story.

Note that having the SnapshotEnvironmentBinding come out of the `ErrorOccured` condition and having it fully deployed are not the same things and entire provisioning process may take longer than 3 minutes. A longer provisioning time would not cause issues for the Integration service as this timeout behavior is just limited to detecting and handling the SnapshotEnvironmentBinding that stays in `ErrorOccurred` condition.

About the differences in types of environments, this decision was made to rectify the immediate issues with namespace-backed environments and the current behavior of the GitOps service. Any further timeouts and error handling that would be implemented between services are expected to be done as part of the [RHTAP-1530](https://issues.redhat.com/browse/RHTAP-1530) feature.

arewm: That makes me feel better then. As long as there will be another approach tied to future work.

14rcole: @arewm Do you think we should make the timeout user-configurable or just clarify in this ADR?

arewm: Clarifying in this ADR seems fine as we have no indication or evidence that a user configurable nature would be helpful and therefore warrant an additional level of complexity. Further iteration should be non-blocking as at worst, it can be extended further/temporarily disabled to get around a reproted issue.

arewm (approved):

dirgim (approved):
LGTM

sonam1412 (approved):
LGTM

PR #146, opened 2023-11-02
Update BuildPipelineSelector example to new syntax

chmeliik:
The BuildPipelineSelector CRD includes a pipelineRef field defined as a Tekton PipelineRef. Previously, build-service used the Tekton v1beta1 APIs. The v1beta1.PielineRef [1] object supports specifying pipelines in two different ways:

1) name + bundle:
```
pipelineRef:
  name: mypipeline
  bundle: docker.io/myrepo/mycatalog:v1.0
```
2) resolver + params:
```
pipelineRef:
  resolver: bundles
  params:
  - name: kind
    value: pipeline
  - name: name
    value: mypipeline
  - name: bundle
    value: docker.io/myrepo/mycatalog:v1.0
```
Build-service used to support only the first style. While migrating to the Tekton v1 APIs, we added support for the second style.

For now, we maintain backwards compatibility with the first style, but are planning to drop it soon. The v1.PipelineRef [2] object supports only the second style.

Update the BuildPipelineSelector example to use the second style.

[1]: https://tekton.dev/docs/pipelines/pipeline-api/#tekton.dev/v1beta1.PipelineRef
[2]: https://tekton.dev/docs/pipelines/pipeline-api/#tekton.dev/v1.PipelineRef

### DIFF
--- a/book/build-service.md
+++ b/book/build-service.md
@@ -76,8 +76,14 @@ spec:
   selectors:
   - name: my-component
     pipelineRef:
-      bundle: quay.io/my-user/my-bundle:v1.0
-      name: my-bundle
+      resolver: bundles
+      params:
+      - name: bundle
+        value: quay.io/my-user/my-bundle:v1.0
+      - name: name
+        value: my-bundle
+      - name: kind
+        value: pipeline
     pipelineParams:
     - name: test
       value: fbc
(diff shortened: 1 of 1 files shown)


chmeliik:
Follow-up to https://github.com/redhat-appstudio/build-service/pull/205

PTAL @mmorhun @mkosiarc @tkdchen @rcerven

Do we want to add some more info or is changing the example good enough?

tkdchen:
LGTM. Updating the examples is good enough from my point of view.

mkosiarc (approved):

mkosiarc:
"At least 2 approving reviews are required by reviewers with write access."

Seems that I am not one of those. Should we all have write access here?

mmorhun (approved):

sbose78 (approved):

PR #147, opened 2023-11-17
Decoupling deployment ADR

ralphbean:


ADR/0032-decoupling-deployment.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Decision
+
+We are going to decouple the deployment from the build, test, and release portions of our system.
+
+**Deployment**:
+
+- The [Environment], [SnapshotEnvironmentBinding], and [GitOpsDeploymentManagedEnvironment]
+  resources will be deprecated and eventually decomissioned.
+- The [application-service] will stop generating GitOps repo content.
```
adambkaplan, line 42: nit - missing ref link to `application-service` below.

ADR/0032-decoupling-deployment.md
```diff
@@ -0,0 +1,201 @@
...
+- HAC will stop rendering [Environments] and their status.
+- If a user wants to make use of deployment capabilities, we will promote the usage of [renovatebot]
+  to propagate _released_ images to their self-managed gitops repo as pull requests.
```
adambkaplan, line 45: Can the same be accomplished with Dependabot if the GitOps repo is on GitHub?

ADR/0032-decoupling-deployment.md
```diff
@@ -0,0 +1,201 @@
...
+
+- We will retain the [DeploymentTargetClaim], [DeploymentTarget], and [DeploymentTargetClass] APIs.
+  Collectively, call these the "DTC" APIs. These should have no dependence on other AppStudio APIs.
+  Only [integration-service] should depend on them. The controllers that support the DTC API should
+  be organized into their own project: "the [deployment-target-operator]", or something like that.
+- [integration-service] should be modified to create and destroy [DeploymentTargetClaims] directly
+  instead of [Environment] resources. Users will be expected to provide integration test pipelines
+  that accept a kubeconfig and snapshot as they do now, but those pipelines should now take initial
+  steps to *deploy the instance of their application to be tested* to the deployment target provided
+  by [integration-service] using the provided kubeconfig.
```
adambkaplan, line 57: I assume this means we will want to provide tasks in catalog. Any limits on which tooling we support for deployment? (ex: `kustomize`, `helm`, etc.)

ADR/0032-decoupling-deployment.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Status
+
+Draft
+
+## Authors
+
+- Ralph Bean
+
+## Context
```
arewm, line 13: This affects multiple other ADRs. Can we establish the links to ADRs that would change based on this? I was thinking something similar to what I did here: https://github.com/redhat-appstudio/book/pull/103/files -- mentioning what is deprecated (or partially deprecated) and what the relationship links are.

It would be helpful to consider the full impact of this ADR.

ADR/0032-decoupling-deployment.md
```diff
@@ -0,0 +1,201 @@
...
+
+- Users who expect effortless deployment of their app when onboarding to the system will be
+  dissapointed. They have more work to do to set up a deployment of their app outside the system.
+- Users will lose visibility of their applications' deployments and status in the AppStudio UI
+  (HAC). Other systems like the Argo UI are arguably better at this than we are.
+- Users who expect to provide and manage their own resources to control their app will be delighted.
+  They now no longer have to interact with an intermediary API to try to express details about their
+  deployment(s).
+- As a team, we'll be in a position to try to achieve independence for the [DeploymentTarget] APIs,
+  make it usable outside the context of AppStudio, and ideally make it attractive for collaborators.
```
adambkaplan, line 150: This is a laudable goal unto itself. I do wonder if the idea of `DeploymentTarget` runs into conflict with other projects in this space, such as [Cluster API](https://cluster-api.sigs.k8s.io/introduction) and [Open Cluster Management](https://open-cluster-management.io/).

adambkaplan (approved):
General +1 on this - a step in the right direction when it comes to decoupling.

ADR/0032-decoupling-deployment.md
arewm, line 45: What might this process look like? What type of repository would Renovate be running on for ensuring up-to-date deployment capabilities? The latency for depending on Renovate seems like it would be too high to be able to run any kind of integration test in succession with the Snapshot generation, therefore causing additional challenges for quickly testing and releasing Snapshots.

Additionally, if the system doesn't know how to deploy components, how would PR-level deployments/integration tests be triggered? Would there be different locations for Renovate-driven updates for PRs as push events?

ADR/0032-decoupling-deployment.md
```diff
@@ -0,0 +1,201 @@
...
+- HAC will stop rendering [Environments] and their status.
+- If a user wants to make use of deployment capabilities, we will promote the usage of [renovatebot]
+  to propagate _released_ images to their self-managed gitops repo as pull requests.
+
+**Test and promotion**:
+
+- We will retain the [DeploymentTargetClaim], [DeploymentTarget], and [DeploymentTargetClass] APIs.
+  Collectively, call these the "DTC" APIs. These should have no dependence on other AppStudio APIs.
+  Only [integration-service] should depend on them. The controllers that support the DTC API should
+  be organized into their own project: "the [deployment-target-operator]", or something like that.
```
arewm, line 52: Do these CRD names still make sense? If the system (i.e. AppStudio) is no longer deploying Applications on environments, then I feel like these should be abstracted further to indicate the intended functionality. Namely that there is some system of provisioning environments. The fact that these are intended for deployment by integration service is no longer relevant to the underlying processes given that we want to isolate it from AppStudio.

ADR/0032-decoupling-deployment.md
arewm, line 57: I assume that these pipelines will not have Renovate integrated into them. The pipelines themselves will have to understand how to crack open the Snapshot to separate which components need to be deployed to an environment and which components will be the "public interface" or will drive the installation.

ADR/0032-decoupling-deployment.md
```diff
@@ -0,0 +1,201 @@
...
+- [integration-service] should be modified to create and destroy [DeploymentTargetClaims] directly
+  instead of [Environment] resources. Users will be expected to provide integration test pipelines
+  that accept a kubeconfig and snapshot as they do now, but those pipelines should now take initial
+  steps to *deploy the instance of their application to be tested* to the deployment target provided
+  by [integration-service] using the provided kubeconfig.
+- We should promote [release-service] as the primary means to advertise to [renovatebot] that one or
```
arewm, line 58: Are you proposing how release service might exist on its own without AppStudio as well? Or will we not be able to separate ourselves from these links for now?

ADR/0032-decoupling-deployment.md
arewm: After reading further, I assume that my reading _here_ was misdirected. The proposal to use Renovate is _only_ for deploying to specific environments. There would need to be configurations present in whatever deployment strategy/repo (i.e. Argo/GitOps) which contains pullspecs that can be updated by Renovate in order to deploy to a long-running environment.

Renovate is not proposed for the flow including the testing of specific built artifacts within the pipeline/process.

ADR/0032-decoupling-deployment.md
```diff
@@ -0,0 +1,201 @@
...
+  a lowercase "e") by way of image tags in a registry.
+
+### Out of scope
+
+Some other interesting ideas that are floating around these days, but which should be taken up in
+other ADRs, if we take them up at all:
+
+- Decoupling [integration-service] and [release-service] from a shared [Snapshot] API. They will
+  continue to share a [Snapshot] API as of this ADR.
+- Decoupling [integration-service] from the [application-service] APIs. It will still promote images
```
arewm, line 69: Can you refer to the APIs specifically like you did with Snapshot? I assume that the primary CRD is the Component.

ADR/0032-decoupling-deployment.md
```diff
@@ -0,0 +1,201 @@
...
+### Use Case Descriptions
+
+**During onboarding**: whereas today when a user requests a new appstudio tier namespace, the tier
+template includes an [Environment] that the integration-service will promote to. Tomorrow, the
+appstudio tier template should no longer include an [Environment] which on its own will cause
+integration-service to _not_ trigger a deployment when testing completes. Instead, the appstudio
+tier template should include a [ReleasePlan] with a reference to the [push-to-registry]
+release pipeline. This new default [ReleasePlan] should carry parameters such that whenever
+a [Snapshot] is successfully tested, a [Release] is created that re-tags the images in their same
+quay repositories with a tag like `:released` or `:validated` (name tbd). The
```
arewm, line 82: This would be reusing the same image repositories that builds are currently pushed to?

ADR/0032-decoupling-deployment.md
```diff
@@ -0,0 +1,201 @@
...
+
+**For manual creation of new environments** - the user manages this directly using a combination of
+their gitops repo and argo, outside of the AppStudio member cluster.
+
+**For automated testing in ephemeral environments** - the user specifies an
+[IntegrationTestScenario] CR with a [DeploymentTargetClass] reference. After a build completes, but
+before it executes tests, the integration-service creates a new [DeploymentTargetClaim] CR with
+a [DTCLS] matching the one on the [IntegrationTestScenario]. This should cause the
+[deployment-target-operator] to provision and bind a [DeploymentTarget] matching the
+[DeploymentTargetClass]. [integration-service] then passes the kubeconfig and [Snapshot] to the
```
arewm, line 117: Nit: 
```suggestion
[DeploymentTargetClass]. [Integration-service] then passes the kubeconfig and [Snapshot] to the
```
I _assume_ that this would invalidate the common link shortcut using a lowercase `i`.

ADR/0032-decoupling-deployment.md
dirgim: We might consider either expanding the Snapshot or providing additional information that helps with this. Some things like resource details (replicas, limits etc.) and environment variables are only propagated through `SnapshotEnvironmentBindings` currently afaik.

ADR/0032-decoupling-deployment.md
ralphbean: Done in bf4520b.

ADR/0032-decoupling-deployment.md
ralphbean: Not yet, not as a part of this ADR.

ADR/0032-decoupling-deployment.md
ralphbean: Ack. I think here, we're giving up control of those details entirely to the user. If they have a helm chart which deploys their app (including specifying replicas, limits, etc) then they just invoke that as one of the first tasks in their integration test pipelinerun. If they have an openshift template which deploys their app (including specifying replicas, limits, etc) then they just process and apply that as one of the first tasks in their integration test pipelinerun.

Let's keep those details _out_ of the Snapshot. It is deliberate feature reduction.

ADR/0032-decoupling-deployment.md
ralphbean: Ack - renaming them is an option. I think they still make sense on a certain level, but the broadness of their potential usage means we could revisit that for sure.

ADR/0032-decoupling-deployment.md
ralphbean: Yes, right!

ADR/0032-decoupling-deployment.md
ralphbean: Done in cc25057.

ADR/0032-decoupling-deployment.md
ralphbean: TBH, I did not anticipate providing our own tasks to support deployment activities and I expected users to provide those - although, that sure would be nice if we did provide something here. There are some things available in artifacthub already for [helm](https://artifacthub.io/packages/search?kind=7&ts_query_web=helm&sort=relevance&page=1).

ADR/0032-decoupling-deployment.md
ralphbean: Yes! 7ea38fa

ADR/0032-decoupling-deployment.md
ralphbean: Updated in 3aa6b1e.

ADR/0032-decoupling-deployment.md
gbenhaim: I don't think that we need to add additional details to the snapshot. It already contains the git reference for each component, so the user can clone the matching manifest for the image with those.

An alternative would be to create another component that just contains manifests (in case the users doesn't want to keep manifests alongside the source code).

ADR/0032-decoupling-deployment.md
```diff
@@ -0,0 +1,211 @@
...
+flowchart TD
+    NSTierTemplate --> |provides| ReleasePlan
+    A[fa:fa-code-commit Git Commit] --> |webhook| PipelinesAsCode
+    PipelinesAsCode --> |creates| build[Build PipelineRun]
+    build --> |triggers| integration-service
+    integration-service --> |asks| decision{Is there a ReleasePlan?}
+    ReleasePlan --> |influences| decision
+    decision --> |yes| Release
+    build --> |pushes the image| quay.io/redhat-user-workloads/tenant/component
+    Release --> |tags the image for release| quay.io/redhat-user-workloads/tenant/component
```
gbenhaim, line 105: It tags the images for release or mirrors them to another quey repo?
I'm a bit worried about the fact the released image will be pulled from a repo that is intended to store CI builds.

ADR/0032-decoupling-deployment.md
```diff
@@ -0,0 +1,211 @@
...
+[push-to-registry] pipeline can use the `appstudio-pipeline` service account in the user's
+dev workspace, which already has push access to the repository in question.
+
+```mermaid
+flowchart TD
+    NSTierTemplate --> |provides| ReleasePlan
+    A[fa:fa-code-commit Git Commit] --> |webhook| PipelinesAsCode
+    PipelinesAsCode --> |creates| build[Build PipelineRun]
+    build --> |triggers| integration-service
+    integration-service --> |asks| decision{Is there a ReleasePlan?}
```
gbenhaim, line 101: I think that the flow chart should also show the flow of event for a PR.

ADR/0032-decoupling-deployment.md
```diff
@@ -0,0 +1,211 @@
...
+    ReleasePlan --> |influences| decision
+    decision --> |yes| Release
+    build --> |pushes the image| quay.io/redhat-user-workloads/tenant/component
+    Release --> |tags the image for release| quay.io/redhat-user-workloads/tenant/component
+    quay.io/redhat-user-workloads/tenant/component --> renovatebot[User provided RenovateBot]
+    renovatebot --> |updates| gitops-repo[User provided GitOps repo]
+    gitops-repo --> |causes deployment via| argocd[User provided ArgoCD]
+    argocd --> |deploys to| cluster[User provided cluster]
```
gbenhaim, line 109: Maybe emphasis that ArgoCD is just an exmaple and other gitops tools can be used. It can even update an Helm repo with the new images, the options are limitless.

ADR/0032-decoupling-deployment.md
```diff
@@ -0,0 +1,211 @@
...
+deployment environments of their choice, manually laying out their application resources in the repo
+(assisted by tools like `kam`), specifying image references by tag to match the `:released` or
+`:validated` tagging schem mentioned above, configuring Argo to deploy from their gitops repo, and
+configuring [renovatebot] to propagate image updates by digest to their gitops repo.
+
+**For manual creation of new environments** - the user manages this directly using a combination of
+their gitops repo and argo, outside of the AppStudio member cluster.
+
+**For automated testing in ephemeral environments** - the user specifies an
+[IntegrationTestScenario] CR with a [DeploymentTargetClass] reference. After a build completes, but
```
gbenhaim, line 122: The user should specify a DeploymentTargetClaim (the class is mentioned in the claim). The integration service should create the DTC as mentioned in the IntegrationTest and deletes it when the test pipeline completes.

ADR/0032-decoupling-deployment.md
```diff
@@ -0,0 +1,211 @@
...
+    PipelinesAsCode --> |creates| build[Build PipelineRun]
+    build --> |triggers| integration-service
+    integration-service --> |consults| IntegrationTestScenario
+    integration-service --> |creates| DeploymentTargetClaim
+    DeploymentTargetClaim --> |references| DeploymentTargetClass
+    DeploymentTargetClaim --> |is bound to| DeploymentTarget
+    DeploymentTarget --> |is associated with| kubeconfig[kubeconfig Secret]
+    kubeconfig --> |is provided to| test
+    integration-service --> |creates| test[Test PipelineRun]
+    DeploymentTarget --> |represents| cluster
```
gbenhaim, line 144: It can represent a namespace as well (and maybe an AMI instance in the future).

ADR/0032-decoupling-deployment.md
gbenhaim: I don't think it conflict. the DeploymentTarget API provides the logic for requesting a compute resource and register handlers that can create it. I think it complements the projects you mentioned.

ADR/0032-decoupling-deployment.md
```diff
@@ -0,0 +1,211 @@
...
+  dissapointed. They have more work to do to set up a deployment of their app outside the system.
+- Users will lose visibility of their applications' deployments and status in the AppStudio UI
+  (HAC). Other systems like the Argo UI are arguably better at this than we are.
+- Users who expect to provide and manage their own resources to control their app will be delighted.
+  They now no longer have to interact with an intermediary API to try to express details about their
+  deployment(s).
+- As a team, we'll be in a position to try to achieve independence for the [DeploymentTarget] APIs,
+  make it usable outside the context of AppStudio, and ideally make it attractive for collaborators.
+
+## Phases of implementation
```
gbenhaim, line 161: Some of those phases can be done at the same time.

ADR/0032-decoupling-deployment.md
arewm: I guess it _could_ tag the image (I assume we use the same SA that has permissions to modify the repo) for release but that will likely not be a big use case. I have always envisioned that the release process is to some other _self-managed_ lifecycle.

If it is not already clear, we need to make sure that we have it clearly documented that the images in the tenant/application/component repository are for CI use and _not_ to depend on their lifecycle for any other use cases.

ADR/0032-decoupling-deployment.md
ralphbean: Yeah, after thinking about it and sharing the idea around with others I've come to the same conclusion @gbenhaim. Thanks for raising it @adambkaplan; it's the right question to ask.

ADR/0032-decoupling-deployment.md
ralphbean: To clarify - I meant for this to be as you read it @gbenhaim: for the release-service pipeline to tag the image in the same repo where it was built. It does have the dangers you worry about.

I thought this would be preferable as a default ReleasePlan that is just "marking" which build in the CI quay repo is "latest". Perhaps the default tag name should be chosen to reflect that low-degree of trustworthiness (say just "latest" rather than "release" or "validated").

To graduate out of an onboarding use case, the user should either update their ReleasePlan to instruct their tenant release pipeline to release to a separately managed image repository, or update their ReleasePlan to associate with a ReleasePlanAdmission in a managed workspace.

ADR/0032-decoupling-deployment.md
ralphbean: Why is that? We can do it, but I feel like it may distract from the new important information in the ADR.

Something like this?

```mermaid
flowchart TD
    NSTierTemplate --> |provides| ReleasePlan
    A[fa:fa-code-commit Pull Request] --> |triggers| PR[On-PR Build PipelineRun]
    B[fa:fa-code-commit Merge] --> |triggers| build[On-Push Build PipelineRun]
    build --> |triggers| integration-service
    integration-service --> |asks| decision{Is there a ReleasePlan?}
    ReleasePlan --> |influences| decision
    decision --> |yes| Release
    PR --> |pushes and tags the image| quay.io/redhat-user-workloads/tenant/component
    build --> |pushes and tags the image| quay.io/redhat-user-workloads/tenant/component
    Release --> |re-tags the image| quay.io/redhat-user-workloads/tenant/component
    quay.io/redhat-user-workloads/tenant/component --> renovatebot[User provided RenovateBot]
    renovatebot --> |updates| gitops-repo[User provided GitOps repo]
    gitops-repo --> |causes deployment via| argocd[User provided ArgoCD]
    argocd --> |deploys to| cluster[User provided cluster]
```

ADR/0032-decoupling-deployment.md
ralphbean: I think you're right. They need to specify a DeploymentTargetClaim *spec*. The integration-service should create a real DTC from that spec.

ralphbean: Something like:

```mermaid
flowchart TD
    User --> |provides| IntegrationTestScenario
    IntegrationTestScenario --> |embeds| spec[DeploymentTargetClaim spec]
    commit[fa:fa-code-commit Git Commit] --> |webhook| PipelinesAsCode
    PipelinesAsCode --> |creates| build[Build PipelineRun]
    build --> |triggers| integration-service
    integration-service --> |consults| IntegrationTestScenario
    spec --> |is used to create| DeploymentTargetClaim
    integration-service --> |creates| DeploymentTargetClaim
    DeploymentTargetClaim --> |is bound to| DeploymentTarget
    DeploymentTarget --> |is associated with| kubeconfig[kubeconfig Secret]
    kubeconfig --> |is provided to| test
    integration-service --> |creates| test[Test PipelineRun]
    DeploymentTarget --> |represents| cluster
    test --> |deploys user app to| cluster
    test --> |executes tests against| cluster
```

ADR/0032-decoupling-deployment.md
ralphbean: Mentioned in 382bcb9.

ADR/0032-decoupling-deployment.md
ralphbean: Updated in 382bcb9.

ADR/0032-decoupling-deployment.md
ralphbean: Updated in 382bcb9.

ADR/0032-decoupling-deployment.md
ralphbean: Updated in 382bcb9.

ADR/0032-decoupling-deployment.md
```diff
@@ -0,0 +1,211 @@
...
+
+See also [RHTAP-1873](https://issues.redhat.com/browse/RHTAP-1873).
+
+## Decision
+
+We are going to decouple the deployment from the build, test, and release portions of our system.
+
+**Deployment**:
+
+- The [Environment], [SnapshotEnvironmentBinding], and [GitOpsDeploymentManagedEnvironment]
```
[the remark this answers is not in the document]
ralphbean: Updated in 382bcb9.

ADR/0032-decoupling-deployment.md
```diff
@@ -0,0 +1,211 @@
...
+  resources will be deprecated and eventually decomissioned.
+- The [application-service] will stop generating GitOps repo content.
```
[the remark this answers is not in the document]
ralphbean: Updated in 382bcb9.

ADR/0032-decoupling-deployment.md
gbenhaim: | To graduate out of an onboarding use case, the user should either update their ReleasePlan to instruct their tenant release pipeline to release to a separately managed image repository, or update their ReleasePlan to associate with a ReleasePlanAdmission in a managed workspace.

maybe we need to emphasis it here.

gbenhaim:
@ralphbean yesterday I discovered that a generic API for dynamic resource allocation was implemented in k8s [1] (it's currently an alpha feature). It's a generalization of  the PV API. I think that our future work should based on  it, instead of moving the DTC API to its own project. We would need to work on integrating Tekton with this new API k8s offers.


[1] https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/

ADR/0032-decoupling-deployment.md
```diff
@@ -0,0 +1,223 @@
...
+  repositories.
+- We will stop deploying the [gitops-service] entirely and the RDS instance it uses.
+- HAC will stop rendering [Environments] and their status.
+- If a user wants to make use of deployment capabilities, we will promote the usage of [renovatebot]
+  to propagate _released_ images to their self-managed gitops repo as pull requests. Dependabot is
+  equally viable if the user's gitops repo is on GitHub.
+
+**Test and promotion**:
+
+- We will retain the [DeploymentTargetClaim], [DeploymentTarget], and [DeploymentTargetClass] APIs.
```
gbenhaim, line 62: @ralphbean based on the discussion we had last week, we can drop this CRD's as well. In addition, the integration service won't be responsible for requesting test environments. Would you like me to update this ADR?

ralphbean: Agreed! I've got a draft update started locally. Didn't finish on Friday. I'll have it posted here before end of day.

ralphbean: Updated in 9acabdcc0e632c13c2b8772d9ac8e75597e2b725.

ADR/0032-decoupling-deployment.md
```diff
@@ -0,0 +1,223 @@
...
+See also [RHTAP-1873](https://issues.redhat.com/browse/RHTAP-1873).
+
+## Decision
+
+We are going to decouple the deployment from the build, test, and release portions of our system.
+
+**Deployment**:
+
+- The [Environment], [SnapshotEnvironmentBinding], and [GitOpsDeploymentManagedEnvironment]
+  resources will be deprecated and eventually decomissioned.
```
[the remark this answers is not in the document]
ralphbean: @skabashnyuk fyi, I just pushed an update this morning since you reviewed. See new changes in 9acabdcc0e632c13c2b8772d9ac8e75597e2b725.

> With [Environment], [SnapshotEnvironmentBinding], and [GitOpsDeploymentManagedEnvironment] depreciation and decommissioning that also means removing Deploy secrets https://github.com/redhat-appstudio/architecture/blob/main/ADR/0022-secret-mgmt-for-user-workloads.md right? We will remain only with Build secrets. The one that deploys only in current namespace/workspace .

Right.

> Any understanding that ephemeral namespaces for testing are needed any secrets?

Well, in this design change - the user is now responsible for deploying their app from their test pipeline. If they upload a build secret, SPI can make that available to the test pipeline. The test pipeline can use the secret in support of deploying a test version of the app to the provisioned compute.

ADR/0032-decoupling-deployment.md
```diff
@@ -0,0 +1,238 @@
...
+- [integration-service] should no longer create and and manage [Environments] and the related
+  [DeploymentTargetClaims]. Users will be expected to provide integration test pipelines that (somehow)
+  specify `resourceClaims`, which will cause a provisioner to provision the compute and inject a
+  kubeconfig for the target compute into the taskrun pod. The user's test pipeline should then take
+  steps to *deploy an instance of their application to be tested* to the compute provided by dynamic
+  resource allocation provisioner, using using the provided kubeconfig.
+- In the intervening time between now and when the Dynamic Resource Allocation APIs are available
+  (OpenShift 4.14 plus the time we need to implement a sandbox SpaceRequest *resource driver*), users
+  that need an ephemeral namespace for testing will need to employ a Task as the first step in their
+  pipeline that creates a SpaceRequest. They will need to use a finally task to clean up the SpaceRequest
```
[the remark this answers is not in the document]
ralphbean: In the interim solution, no. I think if the user doesn't clean it up, they will pile up.

In the long-term solution (DRA), then the resource driver should clean it up after the pod is either deleted or terminates.

ADR/0032-decoupling-deployment.md
```diff
@@ -0,0 +1,238 @@
...
+- We will stop deploying the [gitops-service] entirely and the RDS instance it uses.
+- HAC will stop rendering [Environments] and their status.
+- If a user wants to make use of deployment capabilities, we will promote the usage of [renovatebot]
+  to propagate _released_ images to their self-managed gitops repo as pull requests. Dependabot is
+  equally viable if the user's gitops repo is on GitHub.
+
+**Test and promotion**:
+
+- We will decomission the [DeploymentTargetClaim], [DeploymentTarget], and [DeploymentTargetClass]
+  APIs in favor of the new [Dynamic Resource Allocation APIs] that are an alpha feature of Kubernetes
```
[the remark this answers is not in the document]
ralphbean: Yeah, Alex did a preliminary demo last Thursday. That's definitely a weakness here. If DRA is not successful, then we need a new plan.

ADR/0032-decoupling-deployment.md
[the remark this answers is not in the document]
gbenhaim: we will set and owner reference on the SpaceRequest (the pipeline run will own it), so it will get deleted when the the PLR will get deleted. In addition we will implement a Chron job to deleted old space requests (just so we will have another safe guard).

gbenhaim (approved):

ADR/0032-decoupling-deployment.md
```diff
@@ -0,0 +1,238 @@
...
+    build --> |triggers| integration-service
+    integration-service --> |consults| IntegrationTestScenario
+    IntegrationTestScenario --> |references| testpipeline
+    testpipeline --> |is used to create| testpipelinerun
+    testpipelinerun --> |prompts| driver[Resource Driver]
+    driver --> |creates| compute[Compute - cluster, namespace, or other external resource]
+    driver --> |creates| kubeconfig
+    kubeconfig --> |is injected into pod of| testpipelinerun[Test PipelineRun]
+    integration-service --> |creates| testpipelinerun[Test PipelineRun]
+    testpipelinerun --> |deploys user app to| compute
```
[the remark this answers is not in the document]
ralphbean: I don't think we have this requirement anymore in the same way that we used to have it.

If our system is no longer deploying the user's app, and instead the user is responsible for providing instructions to deploy their app to the target cluster, then it makes sense to me that the user would also be responsible for providing instructions to apply secrets to the target cluster. Perhaps they extract those secrets from their workspace namespace, mounted to their integration test pipelinerun, before they apply them to the target cluster.

ralphbean:
I'm going to merge this now; I think it's been sufficiently circulated and discussed.

If you object post-merge, please raise it and we can open something to consider amending this.

PR #148, opened 2023-11-20
ADR-#### - Upstreaming integration-service

ralphbean:
Related to https://github.com/redhat-appstudio/book/pull/147


ADR/0033-upstreaming-integration-service.md
```diff
@@ -0,0 +1,228 @@
...
+  "default".
+- Attempts to remove the `LABELNAME` label from a [TestSubject] should be rejected (or mutated to
+  "default"?)
+- Attempts to edit the `LABELNAME` label on a [TestSubject] should be permitted.
+
+### The Global Candidate List becomes the control TestSubject, the last TestSubject to pass testing
+
+Today, the "global candidate list" is a term that we use to refer to the list of all the built-image
+pullspecs attached to every [Component] of an [Application]. Replace that implementation
+with an identifying *label* on a [TestSubject] in the namespace: `CONTROLLABEL`. Call the [TestSubject] that
```
arewm, line 96: In order to help troubleshooting issues where the GCL hasn't been updated properly, can we also introduce some status messages including:
1. Reason for not applying `CONTROLLABEL` if TestSubject is unsatisfactory
2. Link to the TestSubject which the `CONTROLLABEL` was removed from
3. Link to the TestSubject to which the `CONTROLLABEL` was moved to

ADR/0033-upstreaming-integration-service.md
```diff
@@ -0,0 +1,228 @@
...
+of theirs. Before mirroring, they want to make sure that the same test works against dozens of
+different kubernetes configurations on different public clouds. That platform engineering team first
+creates a single tekton pipeline that contains the logic to perform the test given a single cluster.
+They then configure dozens of [IntegrationTestScenario] resources, one for each kubernetes
+configuration they want to target. When a new set of pullspecs is proposed to their git repo,
+Jenkins responds. Jenkins then creates a single [TestSubject] in the kubernetes namespace,
+representing the new set of images under test. [integration-service] responds, inspecting all of the
+defined [IntegrationTestScenarios], it invokes the [deployment-target-operator] (from [Decoupling
+Deployment]) to provision each cluster configuration and it runs tests of the [TestSubject] against
+each. The aggregate results and stored in the `status` of the [TestSubject] for jenkins to monitor
```
arewm, line 146: Would it make sense to enable a flow to only re-run specific IntegrationTestScenarios? If during the creation of a TestSubject, the user knows which specific ITSs it wants to run, it seems reasonable that only those desired test scenarios could be triggered. 

This does, however, complicate the decision around moving the `CONTROLLABEL` as all required ITSs configured would need to pass, not just those that were triggered by the TestSubject. This could be achieved by copying the status of the ITSs from the most recent TestSubject which shares a common set of artifact references (which is not discussed in this ADR).

ADR/0033-upstreaming-integration-service.md
```diff
@@ -0,0 +1,228 @@
...
+### The Global Candidate List becomes the control TestSubject, the last TestSubject to pass testing
+
+Today, the "global candidate list" is a term that we use to refer to the list of all the built-image
+pullspecs attached to every [Component] of an [Application]. Replace that implementation
+with an identifying *label* on a [TestSubject] in the namespace: `CONTROLLABEL`. Call the [TestSubject] that
+carries this `CONTROLLABEL` label the **control** TestSubject for the `LABELNAME` label.
+
+[integration-service] should grow a mutating webhook that *removes* the `CONTROLLABEL` label from
+the last control [TestSubject] for the `LABELNAME` label, when the `CONTROLLABEL` label is applied
+to a new [TestSubject] bearing the `LABELNAME` label.
```
arewm, line 101: Would it be acceptable to re-apply the `CONTROLLABEL` on a TestSubject that previously had the label but has been removed? Would there need to be any additional validation on such a TestSubject?

ADR/0033-upstreaming-integration-service.md
```diff
@@ -0,0 +1,228 @@
...
+Note, in this pattern, [Release] construction looks a lot like [TestSubject] construction. The
+triggered system knows enough about the triggering system to optionally initiate itself.
+
+TODO - describe this in more detail
+
+> :memo: Another alternative could be to establish an overarching workflow operator that knows how
+> to bridge the gaps (create TestSubjects and create Releases). TODO what would this look like? Too
+> much domain knowledge in one place?
+
+## Use Cases in Detail
```
arewm, line 132: As I read through these use cases, modifying integration service into an event bus keeps coming to mind. Are there efficiencies that we can gain or use cases that we can improve if we move away from managing state solely via CRs? Triggering from external locations _seems_ more intuitive with a bus.
Would there be any critical functionality/data that we lose if we move to a bus?

It would be a new use case for the integration service and I am not sure where/how it would fit in best, but what if the integration service attested to the artifacts with the result of each ITS's run? It would likely not be an _easy_ solution today (potentially easier with the referrer's API), but would it be any easier "tomorrow" with one of the integration services's APIs over the other?

ADR/0033-upstreaming-integration-service.md
```diff
@@ -0,0 +1,228 @@
...
+representing the new set of images under test. [integration-service] responds, inspecting all of the
+defined [IntegrationTestScenarios], it invokes the [deployment-target-operator] (from [Decoupling
+Deployment]) to provision each cluster configuration and it runs tests of the [TestSubject] against
+each. The aggregate results and stored in the `status` of the [TestSubject] for jenkins to monitor
+and report back to the platform team in CI.
+
+**Using integration-service with TestSubject automation.** A hypothetical platform team might want
+to make use of [integration-service] in conjunction with tekton pipelines that they use for
+building images. Imagine, every night they have a cron trigger which rebuilds their images using
+a static Dockerfile that installs the latest deb packages available to a base image and then
```
arewm, line 152: did you intend to write `deb` or were you thinking `dep` as in dependency?

ADR/0033-upstreaming-integration-service.md
```diff
@@ -0,0 +1,228 @@
...
+that instructs [integration-service] to construct [TestSubjects] in response to our [PaC]
+PipelineRuns.
+
+## Consequences
+
+- [integration-service] now knows nothing about [Applications], [Components], [Environments], etc.
+  It will now only depend on Tekton APIs and DeploymentTarget APIs.
+- We should now be able to tell a story about how someone might want to use [integration-service] on
+  a cluster that only carries the Tekton APIs and the DeploymentTarget APIs. The purpose of
+  integration-service is to trigger *integration tests on collections of container images*. You can
```
arewm, line 173: I think that we need to understand the properties of these collections better. They are not necessarily collections of container images either.

ADR/0033-upstreaming-integration-service.md
```diff
@@ -0,0 +1,228 @@
...
+  instruct it to do this on a case by case basis by creating TestSubject resources, or you can
+  configure a [ToBeDetermined] resource to instruct it to create its own [TestSubjects].
+- [integration-service] today has only one CustomResource of its own: [IntegrationTestScenario].
+  After this change, it will have three: [IntegrationTestScenario], [TestSubject], and
+  [ToBeDetermined].
+- We lose some ability to validate references between resources owned by different AppStudio
+  controllers. Previously, if an IntegrationTestScenario was created referencing an Application
```
arewm, line 180: If we want to be able to re-introduce the concept of "approved entities" for modifying specific references, how might we be able to do this? I imagine it would be more critical for release-like activities which are gated by specific EnterpriseContractPolicies.

Is this really an issue? Could this be seen as a feature to enable greater abilities to reference artifacts produced outside of one's own ownership (i.e. cross-application/team/product testing)?

arewm (commented):
Are all of these "ToBeDetermined" identifiers (resources and controllers) the same throughout the document? If not, can you disambiguate between them?

ADR/0033-upstreaming-integration-service.md
```diff
@@ -0,0 +1,228 @@
...
+therefore some architectural assumptions need to be identified and changed in order to decouple
+them.
+
+See also [Decoupling Deployment] for related context.
+
+Today, [integration-service] does the following things:
+
+- **Snapshot Construction**: When a pipelines-as-code (PaC) PipelineRun completes,
+  [integration-service] uses the [Component] API (the "global candidate list") to construct
+  a [Snapshot], in order to trigger testing.
```
dirgim, line 26: I think that this could be expanded because Snapshot creation is primarily there to create a version of the Application which can then be tested, released and deployed.

ADR/0033-upstreaming-integration-service.md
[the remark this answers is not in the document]
ralphbean: Hm - maybe the next TestSubject to pass testing is automatically crowned the controllabel - or a user can intervene and apply the label to any TestSubject manually to put things back on track.

ADR/0033-upstreaming-integration-service.md
ralphbean: Intentionally debian packages. :)

ralphbean:
> Are all of these "ToBeDetermined" identifiers (resources and controllers) the same throughout the document? If not, can you disambiguate between them?

Yeah, they all refer to the same thing. Let's get a working name for it.  Maybe the `TestSubjectConstructor`? or .. `BuildSelector`?

ADR/0033-upstreaming-integration-service.md
```diff
@@ -0,0 +1,228 @@
...
+  actions to facilitate testing and promotion of that [Snapshot].
+- **DeploymentTarget Provisioning**: If any [IntegrationTestScenarios] specify a related
+  [Environment], then [integration-service] will clone that environment, in order to provision a new
+  bare environment for testing.
+- **App Deployment**: If any [IntegrationTestScenarios] specify a related
+  [Environment], then after cloning the environment, [integration-service] will initiate deployment
+  of a version of the application to be tested. See also [Decoupling Deployment].
+- **Snapshot Test Results Recording**: As PipelineRuns corresponding with the [IntegrationTestScenarios]
+  complete, [integration-service] updates test results on the [Snapshot].
+- **Source Repo Test Results Recording**: As PipelineRuns corresponding with the [IntegrationTestScenarios]
```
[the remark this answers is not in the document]
ralphbean: Resolved in 7c3abf2.

ADR/0033-upstreaming-integration-service.md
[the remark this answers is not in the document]
ralphbean: Yeah - or back to the "degenerate" case or something like that. Not an error state.

ADR/0033-upstreaming-integration-service.md
ralphbean: Maybe the "events" API should be used for some of these?

ralphbean: But, it could be nice to have a "next" and "previous" link in the annotations or something so you can follow the breadcrumb trail of TestSubjects that were sequentially promoted to be the control reference.

ADR/0033-upstreaming-integration-service.md
```diff
@@ -0,0 +1,228 @@
...
+- **Release Initiation** will become the responsibility of [release-service]. See more below.
+- [integration-service] will continue to own **Test Initiation**, **DeploymentTarget Provisioning**,
+  **Snapshot Test Results Recording**, and **Source Repo Test Results Recording**.
+- **Snapshot Construction** and **Global Candidate List Maintenance** will remain responsibilities
+  of [integration-service], but will undergo some changes to remove dependence on external APIs.
+
+### The Snapshot becomes the TestSubject
+
+[integration-service] will fork the [Snapshot] API to a new resource under its domain
+named [TestSubject].
```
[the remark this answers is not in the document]
ralphbean: I think that's what's going on here on a certain level. The reference to the Application is removed and replaced with the `LABELNAME` label to group common TestSubjects in the same way we group with Applications today.

ADR/0033-upstreaming-integration-service.md
```diff
@@ -0,0 +1,228 @@
...
+[TestSubject]. Meaning, when the user creates an [IntegrationTestScenario] and a [TestSubject], and
+the [selector] of the [IntegrationTestScenario] matches the attributes of [TestSubject], then together
+they represent intent for the subject to be tested. The user uses these resources to control what
+the system does.
+
+Today, control of the **Snapshot Construction** process is mixed up in the [Application] and [Component]
+model, and is partly just implied. We *always* construct [Snapshots].
+
+**TestSubject Construction** (called Snapshot Construction in our system today) should have a clear
+control plane resource: the [ToBeDetermined] resource.
```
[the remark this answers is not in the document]
ralphbean: Ack - I'm actually proposing the opposite here - that integration-service grows a new controller driven by the TestSubjectConstructor that is responsible for watching resources that match a selector and constructing new TestSubjects from them.

ADR/0033-upstreaming-integration-service.md
ralphbean: Yes, I think it would be acceptable - to reset your reference subject back to whatever it was at some point in the past. No additional validation necessary from my PoV.

ADR/0033-upstreaming-integration-service.md
```diff
@@ -0,0 +1,228 @@
...
+  complete, [integration-service] updates test results in the source repo service's API (like the
+  github "checks" API).
+- **Global Candidate List Maintenance**: When all PipelineRuns corresponding with the
+  [IntegrationTestScenarios] that are marked as non-optional complete and if the [Snapshot] is
+  designated as a post-merge [Snapshot], then [integration-service] updates the "global candidate
+  list" by way of the [Component] API.
+- **Release Initiation**: When all PipelineRuns corresponding with the [IntegrationTestScenarios]
+  that are marked as non-optional complete and if the [Snapshot] is designated as a post-merge
+  [Snapshot], and if there is a [ReleasePlan] in the namespace with auto-release enabled, then
+  [integration-service] updates the "global candidate list" by way of the [Component] API.
```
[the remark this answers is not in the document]
ralphbean: Added in af2c33f.

ADR/0033-upstreaming-integration-service.md
ralphbean: Hm, ok. After reading the rest of the ADR do you think it's still helpful to add that context here?

ADR/0033-upstreaming-integration-service.md
```diff
@@ -0,0 +1,270 @@
...
+    image_digest: '.status.results.[] | select(.name == "IMAGE_DIGEST").value'
+    name: '.metadata.labels."appstudio.openshift.io/component"'
+    source:
+      git:
+        revision: '.metadata.labels."pipelinesascode.tekton.dev/sha"'
+        url: '.metadata.labels."pipelinesascode.tekton.dev/repo-url"'
+```
+
+Don't take the fields literally. The details will likely change after the ADR has settled. It is
+here for illustration purposes.
```
arewm, line 152: Nit on the wording, maybe change to something like
```suggestion
The fields used in this illustrative example are not part of the ADR. Instead, this sample
is meant to convey the types of information that might be present in the TestSubjectConstructor.
```

ADR/0033-upstreaming-integration-service.md
```diff
@@ -0,0 +1,270 @@
...
+spec:
+  selector:
+    fields:
+      match:
+        kind: PipelineRun
+        version: tekton.dev/v1
+    labels:
+      match:
+        "appstudio.openshift.io/application": my-application
+        "pipelinesascode.tekton.dev/state": completed
```
arewm, line 140: I know that this is just illustrative, but these fields seem like they might not be narrowly defined enough. Would these also match on the IntegrationTestScenarios triggered as well as it seems like we would want to associate those with an application as well.

ADR/0033-upstreaming-integration-service.md
```diff
@@ -0,0 +1,270 @@
...
+```
+
+Don't take the fields literally. The details will likely change after the ADR has settled. It is
+here for illustration purposes.
+
+The `selector` field will be used to identify resources that should trigger the construction of new
+[TestSubjects]. When created, the *basis* for the new test subject should always be taken from the
+**control** [TestSubject]. The values extracted from the triggering resource by way of the
+expressions on the `extractor` resource should be applied to that basis in the construction of a new
+[TestSubject]. If there is no **control** [TestSubject], then the new TestSubject should be created
```
arewm, line 158: If we are loosening the dependency on Applications and Components, how do we know how to convert from a PipelineRun to a resource in the TestSubject?

ADR/0033-upstreaming-integration-service.md
```diff
@@ -0,0 +1,270 @@
...
+**Using integration-service as a remote API from Jenkins.** A hypothetical platform team might want
+to make use of [integration-service] from a jenkins pipeline. For example, perhaps their jenkins job
+performs CI on a git repository that contains pullspecs which should be mirrored to some environment
+of theirs. Before mirroring, they want to make sure that the same test works against dozens of
+different kubernetes configurations on different public clouds. That platform team first
+creates a single tekton pipeline that contains the logic to perform the test given a single cluster.
+They then configure dozens of [IntegrationTestScenario] resources, one for each kubernetes
+configuration they want to target. When a new set of pullspecs is proposed to their git repo,
+Jenkins responds. Jenkins then creates a single [TestSubject] in the kubernetes namespace,
+representing the new set of images under test. [integration-service] responds, inspecting all of the
```
arewm, line 185: This is counter to the proposed TestSubject creation above, namely that the integration service is no longer responsible for creating the TestSubjects. How can the integration service validate that a TestSubject is a legitimate request? Should the integration service care about what is _inside_ the TestSubject or should it just accept _anything_ and forward it on to users' provided ITSs?

ADR/0033-upstreaming-integration-service.md
```diff
@@ -0,0 +1,270 @@
...
+defined [IntegrationTestScenarios], it invokes the [deployment-target-operator] (from [Decoupling
+Deployment]) to provision each cluster configuration and it runs tests of the [TestSubject] against
+each. The aggregate results are stored in the `status` of the [TestSubject] for jenkins to monitor
```
arewm, line 188: This effectively provides a way for some "arbitrary" process to provision an environment and run a workload on it. This feels like a dangerous attack vector.

ADR/0033-upstreaming-integration-service.md
```diff
@@ -0,0 +1,270 @@
...
+and report back to the platform team in CI.
+
+**Using integration-service with TestSubject automation.** A hypothetical platform team might want
+to make use of [integration-service] in conjunction with tekton pipelines that they use for
+building images. Imagine, every night they have a cron trigger which rebuilds their images using
+a static Dockerfile that installs the latest debian packages available to a base image and then
+rebuilds all of their utility images on that new base image. Their pipelines expose the pullspecs of
+all of those images as results. The platform team configures a [TestSubjectConstructor] which
+instructs the [test-subject-construction-controller] to watch all PipelineRuns in their namespace with
+a certain label. As those builds finish, the [test-subject-construction-controller] constructs new
```
arewm, line 198: The TestSubjectConstructor will have to be created for every set of TestSubjects created. We are implying that there is a (potentially reused) dependency on some external pattern. If there are multiple types of artifacts that are produced across build pipelines, there are consistency requirements imposed on these pipelines by a decoupled integration service controller. 

Is there a way that we can improve this? Can we make there be a single explicit requirement/dependency that exists between the services (i.e. an Artifact CRD) instead of these implicit ones?

ADR/0033-upstreaming-integration-service.md
ralphbean: Oh yeah - that's just the point though, that automatic TestSubject creation is an optional behavior of itnegration-service which can be configured, turned off or on by the absence or presence of a TestSubjectConstructor.

What would an illegitimate TestSubject look like?

ralphbean: Remember, not just anybody can create TestSubjects. They have to have RBAC in the namespace where the ITS resides that permits them to create TestSubjects. In this jenkins case, imagine that jenkins has been provisioned a serviceaccount that has explicit rights to create TestSubjects.

Of course, if that SA token were stolen, an attack could create a TestSubject with references to an image of theirs that runs a bitcoin miner (or whatever).

ADR/0033-upstreaming-integration-service.md
ralphbean: Ah - the process must have RBAC to create TestSubjects. That's the limiting factor.

ralphbean: .. similar to RBAC which limits which accounts can create DeploymentTargetClaims, which accomplishes a similar thing more directly.

ADR/0033-upstreaming-integration-service.md
ralphbean: > The TestSubjectConstructor will have to be created for every set of TestSubjects created

I worry we are miscommunicating here; I don't understand this.

The platform team creates the TestSubjectConstructor **once** as part of setting up their process. The constructor resource should stay in place monitoring for new PipelineRuns, guiding the creation of TestSubjects night after night.

ADR/0033-upstreaming-integration-service.md
ralphbean: I think so. This is the `extractors` field in the hypothetical spec of the TestSubjectConstructor.

ADR/0033-upstreaming-integration-service.md
```diff
@@ -0,0 +1,270 @@
...
+### TestSubject Construction as optional behavior
+
+**Test Initiation** has a clear control plane: the [IntegrationTestScenario] combined with the
+[TestSubject]. Meaning, when the user creates an [IntegrationTestScenario] and a [TestSubject], and
+the [selector] of the [IntegrationTestScenario] matches the attributes of [TestSubject], then together
+they represent intent for the subject to be tested. The user uses these resources to control what
+the system does.
+
+Today, control of the **Snapshot Construction** process is mixed up in the [Application] and [Component]
+model, and is partly just implied. We *always* construct [Snapshots].
```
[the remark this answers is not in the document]
ralphbean: > Are you trying to say that today, we always construct Snapshots?

Yes.

> Who is the we -- integration service?

Yes.

> How is mixing up snapshot creation with application/component models relevant?

I guess the more important point is that it is implicit behavior. When you create Applications and Components, you get Snapshot creation in response to Component builds as an implicit side-effect today.

ADR/0033-upstreaming-integration-service.md
arewm:
```
spec:
  selector:
    fields:
      match:
        kind: PipelineRun
        version: tekton.dev/v1
    labels:
      match:
        "appstudio.openshift.io/application": my-application
        "pipelinesascode.tekton.dev/state": completed
```
In this example, the TestSubjectConstructor is scoped to matching on a specific `appstudio.openshift.io/application`, so if multiple Applications are created (i.e. mapping to multiple TestSubjects) then there will need to be multiple TestSubjectConstructors.

One way to get around this would be to convert from the application being used as the selector to instead just use the application name in the extractor (i.e. to set the value of `LABELNAME`).

Even with this generalization, there are implicit requirements imposed upon the build PipelineRuns since the extractor is consistent across the board. This implicit requirement provides no feedback back to the user during the build because the decoupling of the integration service only consumes the PipelineRuns and does not return any feedback to it. The pattern creates an opportunity where a user might want to modify the build pipeline which results in a TestSubject not being properly created if the extracted fields are missing or change behavior.

ADR/0033-upstreaming-integration-service.md
arewm: Might an IntegrationTestScenario have a label associating the related application? If the IntegrationTestScenarios are additionally triggered by PAC (or if there are non-build PAC pipelines), then we could enter into a situation where the TestSubjectConstructor reacts to more PipelineRuns than just ones that produce artifacts.

ADR/0033-upstreaming-integration-service.md
arewm: Right. This question is less relevant with the other comments that I have. I made it earlier in the review and didn't refine it.

ADR/0033-upstreaming-integration-service.md
arewm: > What would an illegitimate TestSubject look like?

I think the situation that I was considering is not likely to be realized. If TestSubjects are manually created _and_ the same TestSubjects (i.e. with the same `LABELNAME`) are created with a TestSubjectConstructor, the two might not have the same data. The manually created ones might not be consistent with extracted data. Does this matter? I assume that the answer is "no" as it would be up to the user/process creating the TestSubjects that it is properly formed for the ITS configured to run from it.

dirgim: I agree that manually created ones shouldn't necessarily need to correspond to generated ones, this kind of behavior is supported by the integration service even now where manually created Snapshots are still tested and processed.

ADR/0033-upstreaming-integration-service.md
dirgim: You're right, it's ok as-is.

ADR/0033-upstreaming-integration-service.md
```diff
@@ -0,0 +1,271 @@
...
+to make use of [integration-service] in conjunction with tekton pipelines that they use for
+building images. Imagine, every night they have a cron trigger which rebuilds their images using
+a static Dockerfile that installs the latest debian packages available to a base image and then
+rebuilds all of their utility images on that new base image. Their pipelines expose the pullspecs of
+all of those images as results. The platform team configures a [TestSubjectConstructor] which
+instructs the [test-subject-construction-controller] to watch all PipelineRuns in their namespace with
+a certain label. As those builds finish, the [test-subject-construction-controller] constructs new
+[TestSubjects] by combining the image list from the **control** [TestSubject] with the new pullspec
+found in the result with the name described in the [TestSubjectConstructor]. The platform team
+benefits from integration-service's [two-phase architecture]. They build reports on the history of
```
dirgim, line 201: We'll probably need to adjust the two phase architecture to account for the new paradigm. While the core idea will remain the same, I think we'll need to specify that we'd be comparing the new `TestSubject` with the newest **control** `TestSubject`, and detecting if any of the other component images (ones that were not part of the `TestSubject` update were changed.
This would probably only work for `TestSubjects` which were constructed with a `TestSubjectConstructor` if I'm not mistaken, since we would depend on the new `TestSubject` containing the information  on which component was changed from the original **control** `TestSubject` at the time of its construction (the `component` label in current implementation).

ADR/0033-upstreaming-integration-service.md
```diff
@@ -0,0 +1,271 @@
...
+    fields:
+      match:
+        kind: PipelineRun
+        version: tekton.dev/v1
+    labels:
+      match:
+        "appstudio.openshift.io/application": my-application
+        "pipelinesascode.tekton.dev/state": completed
+  extractor:
+    image_url: '.status.results.[] | select(.name == "IMAGE_URL").value'
```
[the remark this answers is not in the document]
MartinBasti: Sorry if this applies also here :D
```
Don't take the fields literally. The details will likely change after the ADR has settled. It is
here for illustration purposes.
```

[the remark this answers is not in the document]
ifireball: I suppose some JSON path dialect or CEL could be used here.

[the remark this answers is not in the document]
MartinBasti: That works for me :)

ADR/0033-upstreaming-integration-service.md
```diff
@@ -0,0 +1,271 @@
...
+  of [integration-service], but will undergo some changes to remove dependence on external APIs.
+
+### The Snapshot becomes the TestSubject
+
+[integration-service] will fork the [Snapshot] API to a new resource under its domain
+named [TestSubject].
+
+The purpose of this change is to make [integration-service] independent of other APIs. The name
+[TestSubject] is chosen to reflect what the [Snapshot] represented *in [integration-service]'s
+domain* in our original architecture.
```
arewm, line 70: Moving away from a Snapshot as a centralized API will create challenges for visualizing the flow of artifacts through a system. If the integration service maintains its own representation, then there should be some way for us to visualize those details within its context. 

When we release, however, since we are breaking away from the common model, there would likely be _some other_ resource created by the release service. How can we enable continuity of resources throughout all controllers? As a user of the system, I would hope to be able to see information about a specific set of artifacts ... and follow its lifecycle through the system. For example, see what test cases might have returned as well as seeing that that specific set of artifacts is released successfully.

If we consider only the integration and release services, it seems manageable because we can just add annotations or something similar to illustrate the connections. If there are more controllers in the process, however, this pattern of adding annotations might become harder.

ifireball (approved):
/lgtm

(I'd rather see some attempts at implementing this then continue to discuss details in the ADR)

PR #151, opened 2023-11-27
doc: document changes to snapshot preparation for integration service

14rcole:
The integration service previously used CHAINS-GIT_URL and CHAINS-GIT_COMMIT to test PR changes. However, CHAINS-GIT_URL points to the target repository, not the source repository.  In order to fix this the integration service switched to using `params.git-url` and `params.revision`.  This change documents those code changes.


arewm (approved):

dirgim (approved):
LGTM

PR #156, opened 2023-12-01
Rename the book to "architecture"

ralphbean:
I would rename the repo after this is merged.


arewm (approved):
Without going through the transformation myself and seeing the diff, I assume that you found the necessary changes. :)

ralphbean:
If I missed any, I'll circle back to fix after the rename of the repo itself. :)

scoheb (approved):

Issue #154, opened 2023-11-29
No documentation explaining data-plane data model


jdcasey:
We need a document describing the data-plane resources, what their lifecycle is, how they relate to one another, and how they relate to control-plane resources. For instance:

* how are they created
* how do they evolve
* what state are they intended to capture
* how do the various resources in the data plane fit together among themselves
* how do they relate to services
* how do they relate to the control plane

jdcasey:
Handled by #158

Issue #153, opened 2023-11-29
No documentation explaining control-plane data model


jdcasey:
We need a document describing the control-plane resources, what their lifecycle is, and how they relate to one another. We can infer some things from their names, but it's not 100% clear how they get created and what they describe, especially in context with one another...how do the pieces fit together?

jdcasey:
Handled by #158

PR #158, opened 2023-12-01
Summarize entity roles, lifecycles, relationships

ralphbean:
For #153 and #154.

### DIFF
--- a/architecture/index.md
+++ b/architecture/index.md
@@ -47,9 +47,9 @@ these resources.
   by an [Application]. The user provides and names their [Components] and specifies which git
   repositories and branches define how the artifact is built.  The user may add, remove, or change
   [Components] over the lifespan of an [Application].
-- A [Snapshot] represents a collection of particular OCI artifacts, mapped to the [Components] of an
-  [Application]. A [Snapshot] is owned by an [Application]. Generally, the [integration-service]
-  produces new [Snapshots] automatically in response to completed builds. It is also possible for
+- A [Snapshot] represents a collection of particular OCI artifacts, mapped to [Components].
+  A [Snapshot] is owned by an [Application]. Generally, the [integration-service] produces new
+  [Snapshots] automatically in response to completed builds. It is also possible for
   the user to construct and provide [Snapshots] to test or release deliberate combinations of
   OCI artifacts. [Snapshots] are immutable. For a user to "modify" a [Snapshot], they need to create
   a new [Snapshot] based on an old [Snapshot]. Old [Snapshots] are garbage collected if not bound to
--- a/architecture/index.md
+++ b/architecture/index.md
@@ -44,9 +44,9 @@ these resources.
   generally long-lived and don't change much after they are created.
 - A [Component] represents an OCI artifact to build from a particular git branch (and a particular context
   directory). It is owned by an [Application]. The user provides and names their [Components] and
-  specifies which git repositories and branches they relate to. The user may add, remove, or change
-  [Components] over the lifespan of an [Application].
-- A [Snapshot] represents a collection of particular OCI artifact builds, mapped to the [Components] of an
+  specifies which git repositories and branches define how the artifact is built. The user may add,
+  remove, or change [Components] over the lifespan of an [Application].
+- A [Snapshot] represents a collection of particular OCI artifact, mapped to the [Components] of an
   [Application]. A [Snapshot] is owned by an [Application]. Generally, the [integration-service]
   produces new [Snapshots] automatically in response to completed builds. It is also possible for
   the user to construct and provide [Snapshots] to test or release deliberate combinations of
@@ -125,7 +125,7 @@ Each service that makes up AppStudio is further explained in its own document.
 
 - [Namespace Metadata](../ADR/adr-0010-namespace-metadata)
 
-[integration-service promotes images]: ../ADR/0016-integration-service-promotion-logic.md
+[integration-service promotes OCI artifacts]: ../ADR/0016-integration-service-promotion-logic.md
 [application-service]: ./hybrid-application-service.md
 [gitops-service]: ./gitops-service.md
 [integration-service]: ./integration-service.md
--- a/architecture/index.md
+++ b/architecture/index.md
@@ -39,7 +39,7 @@ API resources in the second row (PipelineRun, Snapshot) should primarilly be tho
 data-plane resources. The system responds to user requests by creating and managing the lifecycle of
 these resources.
 
-- An [Application] represents a functionally coherent set of git branches that should be built,
+- An [Application] represents a functionally coherent set of [Components] that should be built,
   tested, and released together. The user provides and names their [Applications]. They are
   generally long-lived and don't change much after they are created.
 - A [Component] represents an image to build from a particular git branch (and a particular context
--- a/architecture/index.md
+++ b/architecture/index.md
@@ -42,11 +42,12 @@ these resources.
 - An [Application] represents a functionally coherent set of [Components] that should be built,
   tested, and released together. The user provides and names their [Applications]. They are
   generally long-lived and don't change much after they are created.
-- A [Component] represents an OCI artifact to build from a particular git branch (and a particular context
-  directory). It is owned by an [Application]. The user provides and names their [Components] and
-  specifies which git repositories and branches define how the artifact is built. The user may add,
-  remove, or change [Components] over the lifespan of an [Application].
-- A [Snapshot] represents a collection of particular OCI artifact, mapped to the [Components] of an
+- A [Component] represents a git branch on a git repository (and a particular context directory)
+  that should be used to build OCI artifacts from commits that appear there. A [Component] is owned
+  by an [Application]. The user provides and names their [Components] and specifies which git
+  repositories and branches define how the artifact is built.  The user may add, remove, or change
+  [Components] over the lifespan of an [Application].
+- A [Snapshot] represents a collection of particular OCI artifacts, mapped to the [Components] of an
   [Application]. A [Snapshot] is owned by an [Application]. Generally, the [integration-service]
   produces new [Snapshots] automatically in response to completed builds. It is also possible for
   the user to construct and provide [Snapshots] to test or release deliberate combinations of
--- a/architecture/index.md
+++ b/architecture/index.md
@@ -51,8 +51,9 @@ these resources.
   [Application]. A [Snapshot] is owned by an [Application]. Generally, the [integration-service]
   produces new [Snapshots] automatically in response to completed builds. It is also possible for
   the user to construct and provide [Snapshots] to test or release deliberate combinations of
-  OCI artifacts. Old [Snapshots] are garbage collected if not bound to other resources like
-  a [SnapshotEnvironmentBinding] or a [Release].
+  OCI artifacts. [Snapshots] are immutable. For a user to "modify" a [Snapshot], they need to create
+  a new [Snapshot] based on an old [Snapshot]. Old [Snapshots] are garbage collected if not bound to
+  other resources like a [SnapshotEnvironmentBinding] or a [Release].
 - An [IntegrationTestScenario] represents a test that should be run against new [Snapshots] that
   belong to a particular [Application]. It is owned by an [Application]. The user provides
   tekton pipelines that test their application and registers them with the system by creating
--- a/architecture/index.md
+++ b/architecture/index.md
@@ -88,7 +88,11 @@ these resources.
   a [Snapshot]) by particular means (represented by the release pipeline details in
   a [ReleasePlan]). The creation of a [Release] causes [release-service] to create a release
   PipelineRun in one or more workspaces depending on details in the associated [ReleasePlan] and
-  [ReleasePlanAdmission].
+  [ReleasePlanAdmission]. A [Release] can be created in one of two ways: if the [ReleasePlan] has an
+  *automated release* flag set to true, then [integration-service] will automatically create new
+  [Releases] for every [Snapshot] that successfully passes its post-merge testing. If that flag is
+  set to false, then the user is expected to create new a [Release] manually, associated with
+  [Snapshot] selected by the user manually.
 
 ## Service (Component) Context
 
--- a/architecture/index.md
+++ b/architecture/index.md
@@ -57,7 +57,7 @@ these resources.
   tekton pipelines that test their application and registers them with the system by creating
   [IntegrationTestScenarios].
 - An [Environment] represents a destination that [application-service] and the [gitops-service] can
-  deploy things to. Multiple [Applications] in the same workspace can deploy to the same
+  deploy to. Multiple [Applications] in the same workspace can deploy to the same
   [Environment]. An [Environment] may be long lived (like the `development` [Environment] provided
   by the system as part of the initialization of a user's workspace) or may be short-lived and
   ephemeral (like the temporary [Environments] created by the [integration-service] as a part of its
--- a/architecture/index.md
+++ b/architecture/index.md
@@ -39,6 +39,55 @@ API resources in the second row (PipelineRun, Snapshot) should primarilly be tho
 data-plane resources. The system responds to user requests by creating and managing the lifecycle of
 these resources.
 
+- An [Application] represents a functionally coherent set of git branches that should be built,
+  tested, and released together. The user provides and names their [Applications]. They are
+  generally long-lived and don't change much after they are created.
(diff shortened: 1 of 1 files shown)


scoheb (approved):

book/index.md
```diff
@@ -39,6 +39,55 @@ API resources in the second row (PipelineRun, Snapshot) should primarilly be tho
 data-plane resources. The system responds to user requests by creating and managing the lifecycle of
 these resources.
 
+- An [Application] represents a functionally coherent set of git branches that should be built,
```
arewm, line 42:
```suggestion
- An [Application] represents a functionally coherent set of [Components] that should be built,
```

book/index.md
```diff
@@ -39,6 +39,55 @@ API resources in the second row (PipelineRun, Snapshot) should primarilly be tho
...
+  tested, and released together. The user provides and names their [Applications]. They are
+  generally long-lived and don't change much after they are created.
```
arewm, line 44: Nit: Not changing much after they are created seems like a misclassification. They are not changing much after they have been fully defined, but it may take some time for them to be fully defined (i.e. adding components individually).

book/index.md
```diff
@@ -39,6 +39,55 @@ API resources in the second row (PipelineRun, Snapshot) should primarilly be tho
...
+- A [Component] represents an image to build from a particular git branch (and a particular context
```
arewm, line 45: Should we use "image" or some other term? OCI artifact? Can we go ahead and generalize it _now_ or should we live with this definition and update later?

(this would need to be consistent throughout the proposed change)

book/index.md
```diff
@@ -39,6 +39,55 @@ API resources in the second row (PipelineRun, Snapshot) should primarilly be tho
...
+  directory). It is owned by an [Application]. The user provides and names their [Components] and
+  specifies which git repositories and branches they relate to. The user may add, remove, or change
```
arewm, line 47: I wouldn't consider Components really _relating to_ git repos/branches. Instead, I think of it more of which repos/branches define how the underlying artifact is built.

book/index.md
```diff
@@ -39,6 +39,55 @@ API resources in the second row (PipelineRun, Snapshot) should primarilly be tho
...
+  produces new [Snapshots] automatically in response to completed builds. It is also possible for
+  the user to construct and provide [Snapshots] to test or release deliberate combinations of
+  images. Old [Snapshots] are garbage collected if not bound to other resources like
+  a [SnapshotEnvironmentBinding] or a [Release].
+- An [IntegrationTestScenario] represents a test that should be run against new [Snapshots] that
+  belong to a particular [Application]. It is owned by an [Application]. The user provides writes
+  tekton pipelines that test their application and registers them with the system by creating
+  [IntegrationTestScenarios].
+- An [Environment] represents a destination that [application-service] and the [gitops-service] can
+  deploy things to. Multiple [Applications] in the same workspace can deploy to the same
```
arewm, line 60: Can we be more specific than "things" here?

book/index.md
```diff
@@ -39,6 +39,55 @@ API resources in the second row (PipelineRun, Snapshot) should primarilly be tho
...
+  specifies which git repositories and branches they relate to. The user may add, remove, or change
+  [Components] over the lifespan of an [Application].
+- A [Snapshot] represents a collection of particular image builds, mapped to the [Components] of an
+  [Application]. A [Snapshot] is owned by an [Application]. Generally, the [integration-service]
+  produces new [Snapshots] automatically in response to completed builds. It is also possible for
+  the user to construct and provide [Snapshots] to test or release deliberate combinations of
+  images. Old [Snapshots] are garbage collected if not bound to other resources like
+  a [SnapshotEnvironmentBinding] or a [Release].
+- An [IntegrationTestScenario] represents a test that should be run against new [Snapshots] that
+  belong to a particular [Application]. It is owned by an [Application]. The user provides writes
```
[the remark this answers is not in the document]
ralphbean: Oh, the IntegrationTestScenario just has a mandatory reference to the Application in its `spec`: https://redhat-appstudio.github.io/architecture/ref/integration-service.html#integrationtestscenariospec

book/index.md
ralphbean: Oh, why not. Let's say OCI artifact.

book/index.md
```diff
@@ -39,6 +39,55 @@ API resources in the second row (PipelineRun, Snapshot) should primarilly be tho
 data-plane resources. The system responds to user requests by creating and managing the lifecycle of
 these resources.
 
+- An [Application] represents a functionally coherent set of git branches that should be built,
+  tested, and released together. The user provides and names their [Applications]. They are
+  generally long-lived and don't change much after they are created.
+- A [Component] represents an image to build from a particular git branch (and a particular context
+  directory). It is owned by an [Application]. The user provides and names their [Components] and
+  specifies which git repositories and branches they relate to. The user may add, remove, or change
+  [Components] over the lifespan of an [Application].
```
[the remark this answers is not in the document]
ralphbean: Yeah - that's too much detail here.

book/index.md
ralphbean: Agreed. Updated!

book/index.md
ralphbean: I just dropped "things", being even less specific.

I'd rather not work on this part, in light of https://github.com/redhat-appstudio/architecture/pull/147

book/index.md
```diff
@@ -39,6 +39,55 @@ API resources in the second row (PipelineRun, Snapshot) should primarilly be tho
...
+  destination, depending on the implementation of the release pipeline. A [ReleasePlan] is owned by
+  an [Application]. It can operate in two modes, one which executes a "tenant" release pipeline in
+  the user's workspace, and another when used in conjunction with a [ReleasePlanAdmission] where it
+  executes a "managed" release pipeline in a separate privileged workspace owned by another team.
+  The [ReleasePlan] is generally long-lived.
+- A [ReleasePlanAdmission] represents an *acceptance* of release pipeline content from another
+  team's workspace into *this* workspace. It is used exclusively in conjunction with a [ReleasePlan]
+  to represent agreement on details about how to release [Snapshots] across workspace boundaries.
+  The [ReleasePlanAdmission] is generally long-lived.
+- A [Release] represents a request to release a particular set of images (represented by
```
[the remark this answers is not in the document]
ralphbean: Yeah, I was going for "logical" ordering, thinking that you need to know about app and component before you know about snapshot. I think you need to know about RP and RPA before you can learn about Release.

architecture/index.md
```diff
@@ -39,6 +39,55 @@ API resources in the second row (PipelineRun, Snapshot) should primarilly be tho
...
 these resources.
 
+- An [Application] represents a functionally coherent set of [Components] that should be built,
+  tested, and released together. The user provides and names their [Applications]. They are
+  generally long-lived and don't change much after they are created.
+- A [Component] represents an OCI artifact to build from a particular git branch (and a particular context
+  directory). It is owned by an [Application]. The user provides and names their [Components] and
+  specifies which git repositories and branches define how the artifact is built. The user may add,
+  remove, or change [Components] over the lifespan of an [Application].
+- A [Snapshot] represents a collection of particular OCI artifact, mapped to the [Components] of an
```
jdcasey, line 49: It seems like "OCI artifact" is redefined here vs. line 45. I think I get the idea that this is a particular build of the artifact, but I would assume Component branches may change over time without the user redefining the Component, and those would result in multiple of these output OCI artifacts. If Snapshot is the output of building / assembling an Application from a particular set of Component builds, is there a term for the built outputs of those Components?

architecture/index.md
```diff
@@ -39,6 +39,55 @@ API resources in the second row (PipelineRun, Snapshot) should primarilly be tho
...
+  [Application]. A [Snapshot] is owned by an [Application]. Generally, the [integration-service]
```
jdcasey, line 50: This line confuses me a little bit. So, the Snapshot isn't an input to the integration testing process? If the integration service constructs the Snapshot, what does validated input to the integration service look like? (Maybe that answer belongs somewhere else...still reading.)

architecture/index.md
```diff
@@ -39,6 +39,55 @@ API resources in the second row (PipelineRun, Snapshot) should primarilly be tho
...
+  produces new [Snapshots] automatically in response to completed builds. It is also possible for
+  the user to construct and provide [Snapshots] to test or release deliberate combinations of
+  OCI artifacts. Old [Snapshots] are garbage collected if not bound to other resources like
+  a [SnapshotEnvironmentBinding] or a [Release].
+- An [IntegrationTestScenario] represents a test that should be run against new [Snapshots] that
+  belong to a particular [Application]. It is owned by an [Application]. The user provides writes
```
jdcasey, line 56: Nit: "provides" or "writes" but probably not both?

architecture/index.md
```diff
@@ -39,6 +39,55 @@ API resources in the second row (PipelineRun, Snapshot) should primarilly be tho
...
+  [IntegrationTestScenarios].
+- An [Environment] represents a destination that [application-service] and the [gitops-service] can
+  deploy to. Multiple [Applications] in the same workspace can deploy to the same
+  [Environment]. An [Environment] may be long lived (like the `development` [Environment] provided
+  by the system as part of the initialization of a user's workspace) or may be short-lived and
+  ephemeral (like the temporary [Environments] created by the [integration-service] as a part of its
+  test execution processes).
+- A [DeploymentTargetClaim] represents a request to provision a new deployment target. An
+  [Environment] is backed by a [DeploymentTargetClaim]. The [DeploymentTargetClaim] causes new
+  deployment targets to be provisioned which are represented by [DeploymentTargets] that get bound
```
jdcasey, line 67: How are Environment and DeploymentTarget related? It seems like Environment is backed by DeploymentTargetClaim, but it also seems like DeploymentTarget is a resolved / instantiated DeploymentTargetClaim. Is DeploymentTarget then the infrastructural detail where an Environment is made available?

architecture/index.md
```diff
@@ -39,6 +39,55 @@ API resources in the second row (PipelineRun, Snapshot) should primarilly be tho
...
+  to it. These are typically created and destroyed by the [integration-service] as a part of its
+  test execution processes.
+- A [SnapshotEnvironmentBinding] represents a request to deploy a particular set of OCI artifacts
+  (represented by a [Snapshot]) to a particular location (represented by an [Environment]).
+  A [SnapshotEnvironmentBinding] is owned by an [Application]. The [integration-service promotes
+  OCI artifacts] by updating the [SnapshotEnvironmentBinding] associated with an [Environment]. The
```
jdcasey, line 73: This seems to imply that Snapshots are also long-lived, and only updated as new builds of Components are created? Or, is the binding updated with new Snapshot instances?

architecture/index.md
```diff
@@ -39,6 +39,55 @@ API resources in the second row (PipelineRun, Snapshot) should primarilly be tho
...
+  [SnapshotEnvironmentBinding] is generally long-lived.
+- A [ReleasePlan] represents a release pipeline that can be used to release a [Snapshot] to some
+  destination, depending on the implementation of the release pipeline. A [ReleasePlan] is owned by
+  an [Application]. It can operate in two modes, one which executes a "tenant" release pipeline in
+  the user's workspace, and another when used in conjunction with a [ReleasePlanAdmission] where it
+  executes a "managed" release pipeline in a separate privileged workspace owned by another team.
+  The [ReleasePlan] is generally long-lived.
+- A [ReleasePlanAdmission] represents an *acceptance* of release pipeline content from another
```
jdcasey, line 81: Is this used as a gating mechanism when content is being handed off to another workspace, then? Such that the release plan can be resolved and allowed to wait for someone to accept it via the admission object?

architecture/index.md
ralphbean: Yeah, see also https://github.com/redhat-appstudio/book/pull/148.

architecture/index.md
arewm: Snapshots are long lived in that the CRs are not deleted. Some workspaces with more builds have hit the CR limits and need to be manually cleaned up. We are implementing a periodic garbage collection cron until proper CR storage can be implemented (i.e. considering something like https://github.com/adambkaplan/kubearchive).

When a new Component build is created, a Snapshot is created for testing. There is a [two-phase approach](https://github.com/redhat-appstudio/architecture/blob/main/ADR/0015-integration-service-two-phase-architecture.md) for Snapshot promotion based on the success of the required IntegrationTestScenarios.

SEBs are updated with new Snapshot CRs after successfully passing the Composite promotion phase.

architecture/index.md
```diff
@@ -39,6 +39,55 @@ API resources in the second row (PipelineRun, Snapshot) should primarilly be tho
...
+  team's workspace into *this* workspace. It is used exclusively in conjunction with a [ReleasePlan]
+  to represent agreement on details about how to release [Snapshots] across workspace boundaries.
+  The [ReleasePlanAdmission] is generally long-lived.
+- A [Release] represents a request to release a particular set of OCI artifacts (represented by
+  a [Snapshot]) by particular means (represented by the release pipeline details in
+  a [ReleasePlan]). The creation of a [Release] causes [release-service] to create a release
```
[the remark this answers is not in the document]
arewm: The Integration Service can auto-generate Releases if all required IntegrationTestScenarios pass. Otherwise, Releases can be manually created.

architecture/index.md
ralphbean: > If Snapshot is the output of building / assembling an Application from a particular set of Component builds, is there a term for the built outputs of those Components?

No - there's no formal resource on-cluster that represents that. Informally, I've heard it called a "build" or an "image". It's an OCI artifact, described by a pullspec pinned to digest.

architecture/index.md
ralphbean: Agreed! Typo. Fixed in 1768387.

architecture/index.md
ralphbean: You got it. The DeploymentTarget represents the concrete instantiated remote resource. The Environment is backed by the claim, and the claim is *bound* to the DeploymentTarget. Only one DeploymentTarget can be claimed at a time (said another way, only one DeploymentTargetClaim can be bound to a DeploymentTarget at a time). The DeploymentTargetClaim permits managing exclusive access to DeploymentTargets.

It is meant to "look like" the PersistentVolumeClaim API. We modelled it off that. In the PVC APIs, you can have a PersistentVolume that lingers and is claimed multiple times over its life. The PV represents the real storage resource somewhere. You can also have ephemeral PVs that come into and out of existence in response to PVCs being created, say provisioned via AWS APIs. Same idea with the DTC/DT APIs.

architecture/index.md
ralphbean: Ah - a missing important detail is that Snapshots are *immutable*! Updated in ad2961c.

architecture/index.md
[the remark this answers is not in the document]
ralphbean: Added clarification in 723803d.

architecture/index.md
arewm: Immutable except for their statuses.

book/index.md
```diff
@@ -39,6 +39,55 @@ API resources in the second row (PipelineRun, Snapshot) should primarilly be tho
...
 
+- An [Application] represents a functionally coherent set of git branches that should be built,
+  tested, and released together. The user provides and names their [Applications]. They are
+  generally long-lived and don't change much after they are created.
+- A [Component] represents an image to build from a particular git branch (and a particular context
+  directory). It is owned by an [Application]. The user provides and names their [Components] and
+  specifies which git repositories and branches they relate to. The user may add, remove, or change
+  [Components] over the lifespan of an [Application].
+- A [Snapshot] represents a collection of particular image builds, mapped to the [Components] of an
+  [Application]. A [Snapshot] is owned by an [Application]. Generally, the [integration-service]
```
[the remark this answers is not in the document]
ralphbean: Agreed - I took a stab at it in 5a5e057.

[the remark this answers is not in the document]
jdcasey: So, the Application / Snasphot can describe more than Components?

architecture/index.md
jdcasey: Is it fair to say that the system assumes everything will be expressed in OCI terms? I wonder if that's a safe assumption for the long term? (That's a discussion for elsewhere)

arewm: Yes, I think that the system currently assumes that any artifact would be wrapped up in an OCI artifact. One of the key drivers for this assumption is the behavior of Tekton Chains being able to create a signed provenance attestation for the artifact. Expanding to additional artifact storage formats/strategies would likely require some rearchitecting.

PR #161, opened 2024-01-11
fix(STONEINTG-598): Clarify required integration test scenarios

MartinBasti:
Clarify that only required test scenarios are blocking for processing of snapshot. If scenario is marked as optional, any results won't block further processing of snapshot.


arewm (approved):

sonam1412 (approved):

MartinBasti:
I cannot merge here, so please somebody merge it :)

PR #163, opened 2024-01-29
Update goals and constraints

ralphbean:
Recently, more people are coming around the project to learn more about it, how they might use it, and how they might contribute. The aim here is to try to give them the most helpful high level introduction possible before they get to the specifics further down in the doc and further out in the repo.


architecture/index.md
```diff
@@ -1,30 +1,40 @@
...
 - Minimize shared infrastructure among sub-services
 - Participants: onboarding new participants, the flexibility to satisfy the technology preferences of a heterogeneous set of participants. Think of this as the ability to easily create an ecosystem and the ability to support that ecosystem’s heterogeneous needs.
 - Security, Privacy, and Governance: Sensitive data is protected by fine-grained access control
 
+## Architecture Constraints
+
+- Our API server is **the kube API server**. Services expose their API as Custom Resource Definitions. This means that requests are necessarily asyncronous. This means that RBAC is implemented the same way across services. In any exceptional case that a service needs to expose its own bespoke HTTP endpoint (like tekton results), use SubjectAccessReviews to ensure RBAC is consistent.
+- **Use tekton** for anything that should be configurable by the user (building, testing, releasing). Expose as much detail via kube resources as possible. Prefer to implement native tasks to perform work on cluster, rather than calling out to third-party services.
```
ralphbean, line 31:
```suggestion
- **Use tekton** for anything that should be extended by the user (building, testing, releasing). Expose as much detail via kube resources as possible. Prefer to implement native tasks to perform work on cluster, rather than calling out to third-party services.
```

architecture/index.md
```diff
@@ -1,30 +1,40 @@
...
+- As a corollary, adding new functionality usually looks like either adding a new controller or adding new tekton tasks.
+- The **user has admin** in their workspace. This means that the user can access secrets in their workspace. This means that the system can never provide secrets to the user that are scoped beyond that user's domain. A user can exfiltrate the push secret from their workspace, build invalid content on their laptop, and push it to their buildtime registry. Such a build will be rejected as release time.
+- The cluster is our **unit of sharding**. Each cluster is independent and runs an instance of every subsystems. User workspaces are allocated to one cluster. If we lose a cluster, all workspaces on that cluster are inaccessible, but workspaces on other clusters are not impacted, limiting the blast radius. No subsystem should coordinate across clusters.
+- Artifacts built, tested, and shipped by the system are **OCI artifacts**. SBOMs, attestations, signatures, and other supporting metadata are stored in the registry alongside the artifact, tagged by the `cosign triangulate` convention.
```
arewm, line 35: Any attestation used for making a release-time decision should be provably trusted (either because it is GPG signed or its immutable reference is added to the provenance by a trusted task)

arewm (commented):
Thanks for this!

architecture/index.md
```diff
@@ -1,30 +1,40 @@
 # AppStudio
 
 ## Overview
-AppStudio is a platform for building integrated experiences that streamline, consolidate, and secure the application lifecycle.
+AppStudio is a platform for building integrated software that streamlines, consolidates, and secures the development lifecycle.
 
 
 ### Goals
```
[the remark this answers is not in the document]
ralphbean: Yes, absolutely.

architecture/index.md
```diff
@@ -1,30 +1,40 @@
...
+- Provide a way for software teams to release to destinations under the control of their SRE or release engineering team(s).
+- Rapid bootstrapping of new software projects
+- Support both existing and new projects
+- Provide APIs to manage your software lifecycle
+- Provide a surface for partners to integrate, add value
+
+## Architecture Goals
+
 - Just in time scaling: In contrast to “just in case” scaling. The system should be able to scale without capacity reserved ahead of time.
 - Static stability: the overall system continues to work when a dependency is impaired
```
scoheb, line 19:
```suggestion
- Static stability: the overall system continues to work when a dependency is impaired.
```

architecture/index.md
```diff
@@ -1,30 +1,40 @@
...
 - Participants: onboarding new participants, the flexibility to satisfy the technology preferences of a heterogeneous set of participants. Think of this as the ability to easily create an ecosystem and the ability to support that ecosystem’s heterogeneous needs.
 - Security, Privacy, and Governance: Sensitive data is protected by fine-grained access control
 
+## Architecture Constraints
+
+- Our API server is **the kube API server**. Services expose their API as Custom Resource Definitions. This means that requests are necessarily asyncronous. This means that RBAC is implemented the same way across services. In any exceptional case that a service needs to expose its own bespoke HTTP endpoint (like tekton results), use SubjectAccessReviews to ensure RBAC is consistent.
+- **Use tekton** for anything that should be configurable by the user (building, testing, releasing). Expose as much detail via kube resources as possible. Prefer to implement native tasks to perform work on cluster, rather than calling out to third-party services.
+- As a corollary, adding new functionality usually looks like either adding a new controller or adding new tekton tasks.
```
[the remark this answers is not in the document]
scoheb: I agree. good point.

architecture/index.md
```diff
@@ -1,30 +1,40 @@
...
+- The **user has admin** in their workspace. This means that the user can access secrets in their workspace. This means that the system can never provide secrets to the user that are scoped beyond that user's domain. A user can exfiltrate the push secret from their workspace, build invalid content on their laptop, and push it to their buildtime registry. Such a build will be rejected as release time.
```
scoheb, line 33:
```suggestion
- The **user has admin** in their workspace. This means that the user can access secrets in their workspace. This means that the system can never provide secrets to the user that are scoped beyond that user's domain. A user can exfiltrate the push secret from their workspace, build invalid content on their laptop, and push it to their buildtime registry. Such a build will be rejected at release time.
```

architecture/index.md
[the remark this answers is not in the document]
ralphbean: I paraphrased a bit.

arewm (approved):

architecture/index.md
```diff
@@ -1,30 +1,44 @@
...
+## Architecture Goals
+
+- Build artifacts once with enough data to determine suitability for releasing.
+- Build artifacts once that can be released to multiple locations, use cases.
+- "Shift left" the decisions for releasing into PRs; you should be able to release artifacts from a PR as soon as it is merged.
 - Just in time scaling: In contrast to “just in case” scaling. The system should be able to scale without capacity reserved ahead of time.
-- Static stability: the overall system continues to work when a dependency is impaired
-- Each subservice can fulfill its primary use cases independently, without relying on  other systems’ availability.
+- Static stability: the overall system continues to work when a dependency is impaired.
+- Each subservice can fulfill its primary use cases independently, without relying on other systems’ availability.
```
gabemontero, line 23: is this true in the case of tekton?  doesn't seem like it

if so, should we highlight this exception?

gabemontero (commented):
Neither my comment below, nor what I'm about to state here, are specific to your precise changes per se @ralphbean , but as your updating this doc to facilitate people new to Konflux, I *think* this is pertinent.

Additional thoughts:
1) shouldn't this doc have a reference to our catalog of pipelines/tasks over in build-definitions?  Or perhaps at least to the partner tasks section over at https://github.com/redhat-appstudio/architecture/blob/main/ADR/0021-partner-tasks.md ?
2) Tekton is a pretty big black box.  When we say "use tekton" my guess is that some guardrails that Konflux ended up employing organically wrt Tekton API will need to be surfaced somehow.  Like the size restriction with the results API as one example.  Or the use of PVCs given the finite amount of those we have on the cluster (even though there are attempts to work around that with the decouple deployment and alpha level k8s features under consideration).

architecture/index.md
ralphbean: Agreed! Lots of things need tekton (that's what you're saying, right?) Yeah, we should call it out.

gabemontero: > Agreed! Lots of things need tekton (that's what you're saying, right?) Yeah, we should call it out.

correct that is what I am saying

architecture/index.md
```diff
@@ -1,30 +1,44 @@
...
 - Each sub-service owns its data and logic.
 - Communication among services and participants is always asynchronous.
 - Each sub-service is owned by one team. Ownership does not mean that only one team can change the code, but the owning team has the final decision.
 - Minimize shared infrastructure among sub-services
 - Participants: onboarding new participants, the flexibility to satisfy the technology preferences of a heterogeneous set of participants. Think of this as the ability to easily create an ecosystem and the ability to support that ecosystem’s heterogeneous needs.
 - Security, Privacy, and Governance: Sensitive data is protected by fine-grained access control
 
+## Architecture Constraints
+
+- Our API server is **the kube API server**. Services expose their API as Custom Resource Definitions. This means that requests are necessarily asyncronous. This means that RBAC is implemented the same way across services. In any exceptional case that a service needs to expose its own bespoke HTTP endpoint (like tekton results), use SubjectAccessReviews to ensure RBAC is consistent.
```
gbenhaim, line 33: The registration service provided by sandbox also provides an API.
Access to the sandbox proxy is done using SSO.

gbenhaim: Do we need to mention sprayproxy? or this document is about single cluster deployment?

ralphbean: Yeah, those are good examples to add here in the parentheses.

twaugh:
Is running renovate (or something similar) an expectation that would be useful to call out here?

architecture/index.md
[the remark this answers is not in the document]
ralphbean: Any better in 5837ff3?

architecture/index.md
```diff
@@ -1,30 +1,46 @@
...
+## Architecture Goals
+
+- Build artifacts once with enough data to determine suitability for releasing.
+- Build artifacts once that can be released to multiple locations, use cases.
+- Build artifacts should be fully determined by git, not by transient state of the build system. Employ tools like [renovate](https://docs.renovatebot.com/) to manage dependency updates.
+- "Shift left" the decisions for releasing into PRs; you should be able to release artifacts from a PR as soon as it is merged.
 - Just in time scaling: In contrast to “just in case” scaling. The system should be able to scale without capacity reserved ahead of time.
-- Static stability: the overall system continues to work when a dependency is impaired
-- Each subservice can fulfill its primary use cases independently, without relying on  other systems’ availability.
+- Static stability: the overall system continues to work when a dependency is impaired.
```
arewm, line 23: To @twaugh 's comment, I think we can add something about "avoiding flag days" in here. It is also related to our decision to use Tekton (so lower might be an option too), we shouldn't make changes that can break builds ... instead we propose changes and provide a window for accepting them

architecture/index.md
ralphbean: Added in 9332e3b.

architecture/index.md
ralphbean: Added in d3bc7b3.

architecture/index.md
gabemontero: looks good

gabemontero:
> Neither my comment below, nor what I'm about to state here, are specific to your precise changes per se @ralphbean , but as your updating this doc to facilitate people new to Konflux, I _think_ this is pertinent.
> 
> Additional thoughts:
> 
>     1. shouldn't this doc have a reference to our catalog of pipelines/tasks over in build-definitions?  Or perhaps at least to the partner tasks section over at https://github.com/redhat-appstudio/architecture/blob/main/ADR/0021-partner-tasks.md ?
> 
>     2. Tekton is a pretty big black box.  When we say "use tekton" my guess is that some guardrails that Konflux ended up employing organically wrt Tekton API will need to be surfaced somehow.  Like the size restriction with the results API as one example.  Or the use of PVCs given the finite amount of those we have on the cluster (even though there are attempts to work around that with the decouple deployment and alpha level k8s features under consideration).

I see where you addressed 2) here @ralphbean but I'm not seeing activity around my suggestion in 1)

Am I missing it?  Is it still pending?  Or do you not agree with adding such a reference?  

thanks

arewm:
@gabemontero, I don't feel like a reference to the build-definitions are needed here. These updates are just for the high level architecture, goals, etc. and shouldn't _really_ delve into implementation details. Tekton _is_ a big black box and we have done a lot with it. If we need to describe that process more, then maybe we need to add an additional page or add additional context to a current page. Maybe when discussing the `custom` mode in https://github.com/redhat-appstudio/architecture/blob/main/architecture/build-service.md ?

gabemontero:
> @gabemontero, I don't feel like a reference to the build-definitions are needed here. These updates are just for the high level architecture, goals, etc. and shouldn't _really_ delve into implementation details. Tekton _is_ a big black box and we have done a lot with it. If we need to describe that process more, then maybe we need to add an additional page or add additional context to a current page. Maybe when discussing the `custom` mode in https://github.com/redhat-appstudio/architecture/blob/main/architecture/build-service.md ?

Well technically @arewm I'm talking about a reference tohttps://github.com/redhat-appstudio/architecture/blob/main/ADR/0021-partner-tasks.md not build-definitions directly

Bottom line, contributions to the project is going to most likely includes new tasks/pipelines.  Per @ralphbean description up top, we are talking project contributions.  

Somehow, maybe with the docs you noted @arewm , we need to draw a more clearer line, IMO, from this starting point if you will to recommendations on how to contribute pipelines / tasks.

Of course, if I'm in the minority in this opinion, we'll go from there.

Hopefully I at least made my real intent here clearer.

architecture/index.md
```diff
@@ -1,30 +1,48 @@
...
-
-## Architecture Goals and Constraints
-- Robust delivery automation: Establish continuous delivery practices but also deliver operational tooling.
+- Compose software that consists of multiple components, from multiple repositories.
+- Provide transparency on the software supply chain, including both what makes up the software and how it was built.
+- Provide a way for software teams to release to destinations under the control of their SRE or release engineering team(s).
+- Rapid bootstrapping of new software projects
+- Support both existing and new projects
+- Provide APIs to manage your software lifecycle
+- Provide a surface for partners to integrate, add value
```
ifireball, line 14: Perhaps we need something here about observability and visualization into the process as the GUI is an integrated part of the system

ralphbean:
@gabemontero apologies for not responding sooner. I meant to address point 1 of your feedback in 103a116d233b5c70b78ecb54d041dcb950e3dff3.

architecture/index.md
ralphbean: Added in 2735855.

gabemontero:
> @gabemontero apologies for not responding sooner. I meant to address point 1 of your feedback in [103a116](https://github.com/redhat-appstudio/architecture/commit/103a116d233b5c70b78ecb54d041dcb950e3dff3).

thanks @ralphbean for following up

at least for me, that represents the current state of affairs as best as I understand it

@arewm if you did not catch it, you might want to look at the changes in the commit @ralphbean referenced ^^

IIRC you had some thoughts long term around this area of how users provide new tasks/pipelines

ralphbean:
@redhat-appstudio/book-publishers even while there are other proposals posted to potentially change things, I'd like to get this merged to get the current state documented. Can I get one more +1?

gbenhaim (approved):

PR #166, opened 2024-02-04
feat(RHTAPWATCH-782): Add product version controller ADR

ifireball:
Add an ADR describing how to add product and version support to Konflux via a new controller

### DIFF
--- a/ADR/0033-product-version-controller.md
+++ b/ADR/0033-product-version-controller.md
@@ -9,9 +9,9 @@ Proposed.
 ## Context
 
 Konflux began its way as "App Studio", which was mainly designed to facilitate
-the development of online managed services. I order to do that a few major
-abstractions such as "Applications" and "Components" were introduced in order to
-help managing the underlying Tekton Pipeline and other lower level objects.
+the development of online managed services. In order to do that a few major
+abstractions such as "Applications" and "Components" were introduced to help
+managing the underlying Tekton Pipeline and other lower level objects.
 
 With Konflux shifting focus to serving enterprise development teams requests
 had been made to enable using other abstraction metaphors such as "Product" and
@@ -43,9 +43,9 @@ referring that repository already exist in the user's workspace.
 The intention for the solution proposed in this document is to follow the
 pattern described above as far as the Konflux "Application" and "Component"
 objects go and provide a higher-level layer for the streamlined management of
-large amount of such objects.
+large amounts of such objects.
 
-An alternative approach to managing the challenges we had described above is to
+An alternative approach to managing the challenges we have described above is to
 modify the way Application and Component object function so that for example:
 
 * A component may encompass multiple versions of similar code across different
@@ -53,7 +53,7 @@ modify the way Application and Component object function so that for example:
 * An application may include several sub groups of components so that each could
   be regarded to be defining a different version
 
-We see several benefits to the "hight level layer" approach we propose over the
+We see several benefits to the "high level layer" approach we propose over the
 alternative:
 
 * The way Components and Applications work is currently tightly coupled to the
@@ -61,7 +61,7 @@ alternative:
   across multiple services and may introduce operational risks.
 * There are proposals being discussed for decoupling the various Konflux
   services and possibly deprecating the Component and Application objects.
-  Having a separate hight-level layer would allow for adjusting it for using
+  Having a separate high-level layer would allow for adjusting it for using
   different underlying objects while continued emphasis on using Application
   and Component object would necessitate continued maintenance of them.
 
@@ -75,13 +75,13 @@ We will define two new custom resources:
 parallel development streams that may be each released as a different product
 version.
 
-**ProjectDevStream** resources will define a stream of development within a
-*Project*. It will be linked to an owning project via a [Kubernetes owner
-reference][oref].
+**ProjectDevelopmentStream** resources will define a stream of development
+within a *Project*. It will be linked to an owning project via a [Kubernetes
+owner reference][oref].
 
 *Application*, *Component*, and any other Konflux object that needs to be
 created to facilitate the development of a particular project development stream
-will be linked to a *ProjectDevStream* object via an owner reference.
+will be linked to a *ProjectDevelopmentStream* object via an owner reference.
 
 **Note:** The choice was made to use the terms "Project" and "Project
 Development Stream" as opposed to "Product" and "Version" because those can be
@@ -95,8 +95,8 @@ discussion of these terms.
 
 ### Streamlining version creation from the API/CLI.
 
-We will define the **ProjectDevStreamTemplate** custom resource. The *spec* for
-such a resource will include:
+We will define the **ProjectDevelopmentStreamTemplate** custom resource. The
+*spec* for such a resource will include:
 
 * A **project** element referring to an existing *Project* resource.
 * A **variables** element that includes a list of variable names and optional
@@ -105,11 +105,12 @@ such a resource will include:
 * A **resources** element that includes a list of resource definitions so that
   each definition may reference variables by using the Go template syntax.
 
-Here is an example for what a *ProjectDevStreamTemplate* resource may look like:
+Here is an example for what a *ProjectDevelopmentStreamTemplate* resource may
+look like:
 
 ```
 apiVersion: pvc.konflux.dev/v1alpha1
-kind: ProjectDevStreamTemplate
+kind: ProjectDevelopmentStreamTemplate
 metadata:
   name: template1
   ...
@@ -141,17 +142,18 @@ spec:
           uri: ...
 ```
 
-The *spec* section for the *ProjectDevStream* may include a **template** section
-that includes a **name** field for referring to a *ProjectDevStreamTemplate*
-resource and a **values** section to provide values for the template variables.
-When this happens, the resources described by the template get created.
+The *spec* section for the *ProjectDevelopmentStream* may include a **template**
+section that includes a **name** field for referring to a
+*ProjectDevelopmentStreamTemplate* resource and a **values** section to provide
+values for the template variables. When this happens, the resources described by
+the template get created.
 
-Here is an example for what a *ProjectDevStream* resource with a template may
-look like:
+Here is an example for what a *ProjectDevelopmentStream* resource with a
+template may look like:
 
 ```
 apiVersion: pcv.konflux.dev/v1alpha1
-kind: ProjectDevStream
+kind: ProjectDevelopmentStream
 metadata:
   ...
 spec:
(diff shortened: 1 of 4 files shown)


ADR/0033-product-version-controller.md
```diff
@@ -0,0 +1,252 @@
...
+
+### Facilitating product and version based navigation in the UI
+
+We will define two new custom resources:
+
+**Project** resources will define a large project that can have multiple
+parallel development streams that may be each released as a different product
+version.
+
+**ProjectDevStream** resources will define a stream of development within a
```
scoheb, line 78: I would advocate for not using an abbreviation here...but use  `ProjectDevelopmentStream`

ADR/0033-product-version-controller.md
```diff
@@ -0,0 +1,252 @@
...
+
+### Streamlining version cloning from the UI
+
+We will support a process for cloning a *ProjectDevStream* resource and all the
+resources included in it. While the process could be used from the API, it is
+primarily meant to be used from the Konflux UI.
+
+Several technical elements would be needed to support this process.
+
+First and foremost we will support a process for generating a
```
scoheb, line 180: I could use some CR examples like you have in the previous section. I am having a harder time following the flow here.

ADR/0033-product-version-controller.md
```diff
@@ -0,0 +1,252 @@
...
+As one can expect, creating and managing so many Konflux objects is currently
+rather cumbersome. There are a few redundant steps that need to be taken. For
+example, access credentials for a Git repository need to be filled-in every time
+a component referring to that repository is created, even if other components
+referring that repository already exist in the user's workspace.
+
+The intention for the solution proposed in this document is to follow the
+pattern described above as far as the Konflux "Application" and "Component"
+objects go and provide a higher-level layer for the streamlined management of
+large amount of such objects.
```
ralphbean, line 46:
```suggestion
large amounts of such objects.
```

ADR/0033-product-version-controller.md
```diff
@@ -0,0 +1,252 @@
...
+
+An alternative approach to managing the challenges we had described above is to
+modify the way Application and Component object function so that for example:
+
+* A component may encompass multiple versions of similar code across different
+  Git branches
+* An application may include several sub groups of components so that each could
+  be regarded to be defining a different version
+
+We see several benefits to the "hight level layer" approach we propose over the
```
ralphbean, line 56:
```suggestion
We see several benefits to the "high level layer" approach we propose over the
```

ADR/0033-product-version-controller.md
```diff
@@ -0,0 +1,252 @@
...
+alternative:
+
+* The way Components and Applications work is currently tightly coupled to the
+  way the various Konflux services work. Changing it would require changes
+  across multiple services and may introduce operational risks.
+* There are proposals being discussed for decoupling the various Konflux
+  services and possibly deprecating the Component and Application objects.
+  Having a separate hight-level layer would allow for adjusting it for using
```
ralphbean, line 64:
```suggestion
  Having a separate high-level layer would allow for adjusting it for using
```

ADR/0033-product-version-controller.md
```diff
@@ -0,0 +1,252 @@
...
+      - name: foo
+        value: some-value-here
+```
+
+The *values* section must provide values for all template variables without
+default values defined, it may also include values for variable that do have
+default values defined.
+
+The example above will cause the creation of the "some-value-here-app1"
+*Application* resource and the "some-value-here-comp1" *Component* resource.
```
ralphbean, line 170: What happens if someone tries to edit the components, or say delete one component from a new set belonging to a ProjectDevStream that was created from a template? Does the projectdevstream controller unset what the user did, does it recreate the deleted component?

Or, is this a one-pass thing? i.e., once a project dev stream with a template in its spec has been processed, something updates its status, a label, or an annotation to stop further reconciliation of those resources?

ADR/0033-product-version-controller.md
```diff
@@ -0,0 +1,252 @@
...
+4. Display a screen allowing the user to select a name fo the new
+   *ProjectDevStream*  resource and values for all the template variables. When
+   default values  exist, the screen should display then and let the user keep
+   them.
+5. Once the user confirms the screen, create a *ProjectDevStream* resource with
+   a 'template' property referring to the template and values for the variables
+   as selected by the user.
+6. Once all the child resources specified by the template are created (The state
+   field on the *ProjectDevStream* resource should indicate that) Delete the
+   template.
```
ralphbean, line 248: Is this the step that prevents further alignment of the applications and components with the projectdevstream's template?

ADR/0033-product-version-controller.md
ifireball: I think that implementation-wise it would be simpler to keep the templates and the generated objects in sync via the reconciliation loop. That may also end up being useful for some advanced users (Think: bulk-change multiple versions).

In practice when using the UI, I think each template will only get applied once and discarded.

ADR/0033-product-version-controller.md
ifireball: That, and line #3 above are in place to prevent all kinds of strange loops.

ADR/0033-product-version-controller.md
ifireball: Done.

ADR/0033-product-version-controller.md
ifireball: Added a rather large example in the bottom.

ADR/0033-product-version-controller.md
```diff
@@ -0,0 +1,493 @@
...
+alternative:
+
+* The way Components and Applications work is currently tightly coupled to the
+  way the various Konflux services work. Changing it would require changes
+  across multiple services and may introduce operational risks.
+* There are proposals being discussed for decoupling the various Konflux
+  services and possibly deprecating the Component and Application objects.
+  Having a separate high-level layer would allow for adjusting it for using
+  different underlying objects while continued emphasis on using Application
+  and Component object would necessitate continued maintenance of them.
```
arewm, line 66: How would this higher level layer benefit in the decoupling process? If we have some higher level representation, we don't want it to just become _another_ API that needs to adjust/migrate during the decoupling.

ADR/0033-product-version-controller.md
```diff
@@ -0,0 +1,493 @@
...
+a component referring to that repository is created, even if other components
+referring that repository already exist in the user's workspace.
+
+The intention for the solution proposed in this document is to follow the
+pattern described above as far as the Konflux "Application" and "Component"
+objects go and provide a higher-level layer for the streamlined management of
+large amounts of such objects.
+
+An alternative approach to managing the challenges we have described above is to
+modify the way Application and Component object function so that for example:
```
arewm, line 49: The ADR (https://github.com/redhat-appstudio/architecture/pull/167) that I submitted can be viewed as an alternative approach as well. By getting rid of our Component and Application model, we have the ability to remove many of the restrictions of the current system.

ADR/0033-product-version-controller.md
```diff
@@ -0,0 +1,493 @@
...
+
+* The new *ProjectDevelopmentStreamTemplate* will include a *version* variable
+  with no default value.
+
+* All *Application* and *Component* resources that are associated with the
+  cloned *ProjectDevelopmentStream* will be copied into the *resources* section
+  of the *ProjectDevelopmentStreamTemplate* resource, while *status* and other
+  transient properties would be stripped.
+
+* Resources names within the template will be set as following:
```
arewm, line 206:
```suggestion
* Resource names within the template will be set as following:
```

ADR/0033-product-version-controller.md
```diff
@@ -0,0 +1,493 @@
...
+
+    * If the name on the cloned resource ends with the
+      *ProjectDevelopmentStream* resource's version value as determined above,
+      the value is stripped and the "`{version}`" template string is places
```
arewm, line 210:
```suggestion
      the value is stripped and the "`{version}`" template string is used
```

ADR/0033-product-version-controller.md
ifireball: It easier to adjust a higher-level layer to changing lower-level objects - then it is to port core functionality from one set of objects to another.

ADR/0033-product-version-controller.md
ifireball: The two might be complementary - "Artifact" and "ArtifactCollection" seem to be very abstract concepts, the controller described here could be used to link them to the terminology referred by the development teams.

ADR/0033-product-version-controller.md
```diff
@@ -0,0 +1,493 @@
...
+marketable products. Those terms also more accurately describe the environment
+in which software development happens. [See here][pvsp] for a more detailed
+discussion of these terms.
+
+[oref]: https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/
+[pvsp]: https://opensource.com/article/19/11/product-vs-project
+
+### Streamlining version creation from the API/CLI.
+
+We will define the **ProjectDevelopmentStreamTemplate** custom resource. The
```
ralphbean, line 98: Assume we do want and need `Project` and `ProjectDevelopmentStream` for a moment.

Do we need `ProjectDevelopmentStreamTemplate`?

* Is it possible to re-use openshift templates rather than add another API resource here? We need our system to be able to work on kube, not just on openshift - so that may nix that option.
* Is it possible to use helm with in-cluster resources like it is with openshift templates?

ifireball: While I am looking at OpenShift templates to get ideas about how to implement this. I think the scope for ProjectDevelopmentStreamTemplates is much smaller then for Helm of OpenShift templates. I think the set of capabilities needs to be much narrower.

ADR/0033-product-version-controller.md
```diff
@@ -0,0 +1,493 @@
...
+    git:
+      context: ./
+      dockerfileUrl: Dockerfile
+      revision: fixed-rev
+      uri: git@github.com:example/comp2.git
+```
+
+## Consequences
+
+TBD.
```
ralphbean, line 493: It would be good to fill this out.

* What is potentially more difficult if we adopt this decision?
* What is potentially easier?
* Which teams/subsystems are affected and need to take action?

ADR/0033-product-version-controller.md
```diff
@@ -0,0 +1,493 @@
+# 0032 Product and Version Controller
```
ralphbean, line 1: Update the number to match the filename. :)

ADR/0033-product-version-controller.md
```diff
@@ -0,0 +1,493 @@
...
+  template:
+    name: my-cool-project-stream-template
+    values:
+    - name: version
+      value: v1.0.0
+    - name: cool-comp1-dockerfileUrl
+      value: Dockerfile-rhel10
+```
+
+Sine the *ProjectDevelopmentStream* resource is referencing a template, here are
```
twaugh, line 432:
```suggestion
Since the *ProjectDevelopmentStream* resource is referencing a template, here are
```

ADR/0033-product-version-controller.md
```diff
@@ -0,0 +1,493 @@
...
+          uri: git@github.com:example/comp2.git
+```
+
+Following is a new *ProjectDevelopmentStream* resource making use of the
+template above:
+
+```
+apiVersion: pvc.konflux.dev/v1alpha1
+kind: ProjectDevelopmentStream
+metadata:
```
twaugh, line 415: Having a `ProductDevelopmentStream` resource which is able to name a product stream in its metadata could be really useful for augmenting SBOMs with the data needed for product security.

ADR/0033-product-version-controller.md
ifireball: Reflecting conclusions from conversations that happened elsewhere - there is no readily-available solution that is commonly used across the K8s landscape and that would fit our purpose, so we'll have to make our own.

ADR/0033-product-version-controller.md
ifireball: Added content to the Consequences section.

ADR/0033-product-version-controller.md
```diff
@@ -490,4 +490,24 @@ spec:
...
+
+Using this functionality can result in the creation if large amounts of
+*Component* and *Application* resources. Navigating a UI based on those can thus
+become a bit cumbersome. This indicates the significance of having a project and
+development stream based UI.
+
+This solution is planned to be a layer on top of existing elements, to be
+optional to use, and to co-exist with using the system as it is used today. This
+also means that if we decide at a later time that this was not a sound technical
+direction, we can drop the whole thing without extra cost or risk to exiting
```
twaugh, line 512:
```suggestion
direction, we can drop the whole thing without extra cost or risk to existing
```

ifireball:
@ralphbean what is the gap to getting this merged?

scoheb:
LGTM!

scoheb (approved):

arewm (approved):

PR #167, opened 2024-02-05
Propose a new API around Artifacts and their collections

arewm:


ADR/0035-artifact-streams.md
```diff
@@ -0,0 +1,143 @@
+# 32. Decoupling Deployment
```
scoheb, line 1: copy and paste error?

ADR/0035-artifact-streams.md
```diff
@@ -0,0 +1,143 @@
...
+## Decision
+
+We are going to abstract the concepts of the built artifacts and their relationships to each other outside of
+any other controller within AppStudio. If a controller wants to maintain complete separation from others (i.e.
+in order to be fully deployable on its own), it may use an adapter pattern to convert this abstraction to its
+own relevant internal data model.
+
+### ArtifactConfig
+
+An `ArtifactConfig` describes the metadata which is used to define an `Artifact`, an `ArtifactStream`, and
```
scoheb, line 64: I eventually think some example CRs might help visualize this.

ADR/0035-artifact-streams.md
```diff
@@ -0,0 +1,143 @@
+# 32. Decoupling Deployment
+
+Date started: 2023-11-17
+Date accepted: 2023-02-01
```
johnbieren, line 4: Is this supposed to be 2024 or is the accepted date actually supposed to be after the started date?

ADR/0035-artifact-streams.md
```diff
@@ -0,0 +1,143 @@
...
+### ArtifactConfig
+
+An `ArtifactConfig` describes the metadata which is used to define an `Artifact`, an `ArtifactStream`, and
+how `Artifact`s produced update that stream.
+
+`Artifact`s do not exist in isolation to other `Artifact`s in the system, however, there are both explicit
+and implicit relationships that exist between them, for example:
+
+- (implicit) Multiple `ArtifactConfig`s can be created for the same repository and branch combinations. These
+  configurations would be allowed as long as the name of the produced artifact is unique. This supports use
```
scoheb, line 71: > These
  configurations would be allowed as long as the name of the produced artifact is unique
an insight on how this would be achieved?

ADR/0035-artifact-streams.md
arewm: Yes, thanks. I had forgotten that I copied the previous ADR for structure.

ADR/0035-artifact-streams.md
arewm: Yes, thanks. I had forgotten that I copied the previous ADR for structure.

ADR/0035-artifact-streams.md
```diff
@@ -0,0 +1,142 @@
...
+  properly functioning.
+
+## Decision
+
+We are going to abstract the concepts of the built artifacts and their relationships to each other outside of
+any other controller within AppStudio. If a controller wants to maintain complete separation from others (i.e.
+in order to be fully deployable on its own), it may use an adapter pattern to convert this abstraction to its
+own relevant internal data model.
+
+### ArtifactConfig
```
gbenhaim, line 61: It would be great to mention which CRs should be created by the users and which CRs are created by controllers as a result of a build (for example). For machine generated CRs, I think we should have a plan on how we offload those from the cluster to an external database (like we do for pipeline runs). This is needed in order to keep the history without overloading etcd and will also make migration of tenants between clusters easier.

gbenhaim (changes requested):

ADR/0035-artifact-streams.md
arewm: Yeah, I can work on these eventually. I can see how it would be great to get them written up before this ADR is approved for sure. I wanted to get discussion started before I was able to work on example CRs since those haven't been in my mind specifically.

ADR/0035-artifact-streams.md
arewm: I agree that we should offload the information from the cluster so that we don't continue to hit the problems that we encounter today. A generic tool like [KubeArchive](https://github.com/adambkaplan/kubearchive) would be critical as we will be producing multiple CRs for each artifact produced by the system.

ADR/0035-artifact-streams.md
arewm: That's a great challenge. I'll have to think about this more based on my recent push. I assumed that there would be a lazy resolution of an Artifact, so it wouldn't be until the PLR is completed that the uniqueness would be enforced. If there is an owner reference set, however, one challenge will be how to maintain that relationship without knowing the explicit owner reference.

If a single PLR is triggered that is known to produce multiple artifacts, then there could be multiple `Artifact`s created without knowledge of what the final resolved artifact is. Once the PLR is completed, the `Artifact`s will be updated to match their reference with the specific artifact that is build (i.e. one of multiple container images produced from `ko`).

ADR/0035-artifact-streams.md
gbenhaim: I is on our roadmap to maintain[KubeArchive](https://github.com/adambkaplan/kubearchive) ?

ADR/0035-artifact-streams.md
```diff
@@ -0,0 +1,150 @@
...
+- (explicit) A single `ArtifactConfig` can describe the configuration for producing the "same" `Artifact"
+  from multiple branches in the same git repository. As an `ArtifactConfig` also controls the creation of
+  an `ArtifactStream`, `Artifact`s produced on any one of the branches would update the corresponding
+  `ArtifactStream`.
+- (explicit) A single repository and branch may produce different `Artifact`s under different conditions.
+  When artifacts are produced as part of a pull request, for example, the controller will still need to
+  generate `Artifact` resources, but these should not be able to update any `ArtifactStream`s for the 
+  on-push build events
+
+### Artifact
```
ifireball, line 81: By storing Atrifact resources we make etcd effectively assume the role of a container registry or an Artifactory server except for storing the actual build result bits. That may not scale.

Do we really need that? Perhaps its sufficient to define ArtifactStrteam as a generalized pointer to registries, package repositories, etc.

ADR/0035-artifact-streams.md
```diff
@@ -0,0 +1,150 @@
...
+PipelineRun is created, an `Artifact` should be created to enable visibility into the ongoing process if
+an `ArtifactStream` is used for tracking or visualization.
+
+### ArtifactStream
+
+All `Artifact`s have an owner reference to a single `ArtifactStream`. The `ArtifactStream` keeps track of the latest
+produced `Artifact` that has been created owned by that stream (i.e. similar to a floating tag reference).
+
+A validation webhook will prevent an `Artifact` from being created unless its configuration matches that of its owning
+`ArtifactStream`.
```
ifireball, line 103: What does "configuration matches" mean in this context?

ADR/0035-artifact-streams.md
```diff
@@ -0,0 +1,150 @@
...
+## Authors
+
+- Andrew M
+
+## Context
+
+Since the beginning of AppStudio, we have had a dependency on the [Component] and [Application] model that
+originated in the Hybrid Application Service. A [Component] is a single containerized artifact that has
+information on how to build and deploy recorded in a linked [Devfile]. Multiple [Components] can be grouped
+together into a single [Application] to indicate that the should be deployed together via GitOps.
```
ralphbean, line 26:
```suggestion
together into a single [Application] to indicate that they should be deployed together via GitOps.
```

ADR/0035-artifact-streams.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Authors
+
+- Andrew M
+
+## Context
+
+Since the beginning of AppStudio, we have had a dependency on the [Component] and [Application] model that
+originated in the Hybrid Application Service. A [Component] is a single containerized artifact that has
+information on how to build and deploy recorded in a linked [Devfile]. Multiple [Components] can be grouped
```
ralphbean, line 25: In the _current_ system, they must be grouped - or "are" grouped.

```suggestion
information on how to build and deploy recorded in a linked [Devfile]. Multiple [Components] are grouped
```

ADR/0035-artifact-streams.md
ralphbean: @gbenhaim the agogos team is considering adopting it. If they do, then the answer to your question is yes.

Note, the kubearchive repo moved to https://github.com/kubearchive/dynowatch

ADR/0035-artifact-streams.md
```diff
@@ -0,0 +1,150 @@
...
+An `ArtifactConfig` describes the metadata which is used to define an `Artifact`, an `ArtifactStream`, and
+how `Artifact`s produced update that stream.
+
+`Artifact`s do not exist in isolation to other `Artifact`s in the system, however, there are both explicit
+and implicit relationships that exist between them, for example:
+
+- (implicit) Multiple `ArtifactConfig`s can be created for the same repository and branch combinations. These
+  configurations would be allowed as long as the name of the produced artifact is unique. This supports use
+  cases like `ko` where multiple `Artifact`s would be produced from a single build
+- (explicit) A single `ArtifactConfig` can describe the configuration for producing the "same" `Artifact"
```
ralphbean, line 72: What is implied by putting "same" in quotes? Do you just mean that two different commits with two different git digests would produces two different binary images, but that two different ArtifactConfigs could be used to pipe them to a single ArtifactStream? Since they both feed the same ArtifactStream they can be thought of as being "the same" in that way?

ADR/0035-artifact-streams.md
ralphbean: It may be worth elaborating what use cases the Artifact-as-a-custom-resource solves. Why not just treat OCI pullspecs as artifacts? Could rest of the proposal stand (ArtifactConfig and ArtifactStream and ArtifactCollection and ArtifactCollectionStream) even if Artifact itself is dropped?

ADR/0035-artifact-streams.md
```diff
@@ -0,0 +1,150 @@
...
+`Artifact`. An example of this constructor can resemble the `TestSubjectConstructor` as proposed in
+[Upstreaming integration-service].
+
+`Artifact`s can also store references to the process which created the object (i.e. PipelineRun). As soon as a
+PipelineRun is created, an `Artifact` should be created to enable visibility into the ongoing process if
+an `ArtifactStream` is used for tracking or visualization.
+
+### ArtifactStream
+
+All `Artifact`s have an owner reference to a single `ArtifactStream`. The `ArtifactStream` keeps track of the latest
```
ralphbean, line 99: Can Artifacts be members of multiple ArtifactStreams?

ralphbean:
Just in terms of readability, it might make sense to move ArtifactConfig to be the last CR in your list of CRs. The ArtifactStream and ArtifactCollection (and its stream) are more primitive. It's easier to try to understand what the ArtifactConfig might be representing once you know about those other pieces.

ADR/0035-artifact-streams.md
stuartwdouglas: Silly question, given that these are going to need to be stored in some kind of off cluster database anyway, did anyone consider moving this data model off cluster into a RDBMS + REST service? If controllers need information about an object they can just query the REST API directly.

This means that current and historic data would all be in the same system, and it would reduce the load on etcd. 

There may be compelling reasons to keep the current on-cluster data storage, but if we are making a big change to the architecture it might be worth considering completely different approaches.

gbenhaim: I totally agree with @stuartwdouglas. We must consider RDBMS + REST service.

arewm: I don't think that we have had an ADR behind our decision to use CRs as a data model _and_ API. I assume that this was done back when the architecture was dependent on KCP as these resources would enable data to be shared and referenced regardless of where it was generated/used.

@ralphbean and I have had discussions in the past about whether we should use resources or events as interface models. This discussion falls in a similar vein.

ADR/0035-artifact-streams.md
arewm: This use case would be to support versioning which is tied to the branches. There is likely little to no difference in how the `Artifact` is build from one branch to the next, but the content of the artifact changes.

ADR/0035-artifact-streams.md
arewm: No

ADR/0035-artifact-streams.md
twaugh, line 72:
```suggestion
- (explicit) A single `ArtifactConfig` can describe the configuration for producing the "same" `Artifact`
```

(Fix closing backquote)

ADR/0035-artifact-streams.md
```diff
@@ -0,0 +1,150 @@
...
+### ArtifactCollection
+
+An `ArtifactCollection` is a set of `Artifact`s that are grouped together. An `ArtifactCollection` can include
+references to other `ArtifactCollection`s to enable intermediate logical separation and aggregated groupings.
+Additionally, there are no restrictions around how many `ArtifactCollection`s a given `Artifact` may belong to.
+
+### ArtifactCollectionStream
+
+All `ArtifactCollection`s have an owner reference to a single `ArtifactCollectionStream`. The `ArtifactCollectionStream`
+defines the set of `ArtifactStream`s and `ArtifactCollectionStream`s that are allowed in the owned `ArtifactCollection`s.
```
mmorhun, line 114: Could you explain: `'ArtifactCollectionStream's that are allowed in the owned 'ArtifactCollection's.`?

ADR/0035-artifact-streams.md
```diff
@@ -0,0 +1,150 @@
...
+The `ArtifactCollectionStream` also defines whether any individual update to a contained stream results in the automatic
+generation of a new owned `ArtifactCollection`. The `ArtifactCollectionStream` keeps track of the latest produced
```
mmorhun, line 116: I'm afraid, that with this approach we'll end up in many separate resources created. This is similar to current snapshots problem.

ADR/0035-artifact-streams.md
```diff
@@ -0,0 +1,150 @@
...
+A validation webhook will prevent an `ArtifactCollection` from being created unless all contained `Artifact`s and
+`ArtifactCollection`s match its owning `ArtifactCollectionStream`.
+
+## Consequences
+
+- If any controller does not want to interface with the `Artifact*` CRs directly (i.e. to achieve independence
+  from other controllers within AppStudio), they will need to create an adapter pattern between the relevant
+  APIs and their own internal data model.
+- We should now have native ability to describe built artifact and their various supply chain related attestations.
+  This ability holds whether the artifact is built within the kubernetes cluster or outside of it.
```
mmorhun, line 128: Do you mean that a user can build an artifact somewhere and add a record of it in the system?

ADR/0035-artifact-streams.md
```diff
@@ -0,0 +1,150 @@
...
+- We lose the simplification of having a common [Component] and [Application] data model requirement between
+  all controllers as the relationships with `Artifact`s and related resources are loose couplings
+- We will have to produce multiple adapters for the various controllers in order to achieve data model independence
+- While each of the controllers that compose AppStudio should be able to exist on their own, they can all
+  "come together" around a common description of produced artifacts in order to create a unified build, test,
+  and delivery pipeline.
```
mmorhun, line 134: What is the rationale of the adapters? Data model independence? It looks somewhat controversially to me. We want independent controllers but creating dependencies with adapters (are they also controllers?). Will independent components be usable on its own?

ADR/0035-artifact-streams.md
```diff
@@ -0,0 +1,150 @@
...
+
+- PipelineRuns triggered by PipelinesAsCode produced a single image that is tied to a specific [Component]
+  which is assumed to be an OCI artifact.
+- The AppStudio UI (HAC) display PipelineRuns for all [Component]s in a single [Application] together.
+- [Component]s are tied to artifacts produced from a single a git repository and branch.
+
+This model presented multiple challenges including:
+
+- If a user wants to build a new branch of an already onboarded git repository, they would need to create
+  a new [Component] (with a workspace-unique name).
```
Tojaj, line 40: @arewm  what if we don't want previous branch to be built any further? Can be existing component updated to ref a new branch?

ADR/0035-artifact-streams.md
johnmcollier, line 81: How are different artifact "types" defined? I assume the `ArtifactConfig` would determine that, but can it be any type, or is it restricted to specific types (rpm, fbc, etc), which may have their own validation? (E.g. you mention some validation for RPM-based `Artifact`s farther below)

ADR/0035-artifact-streams.md
```diff
@@ -0,0 +1,150 @@
...
+  generate `Artifact` resources, but these should not be able to update any `ArtifactStream`s for the 
+  on-push build events
+
+### Artifact
+
+An `Artifact` represents a single object (binary or non-binary artifacts) produced as part of a software
+supply chain. It not only defines the storage location of the object itself but also the location of any
+additional supporting metadata like produced SBOMs, provenance, and signatures (and the public gpg keys).
+
+Since the concept of an `Artifact` exists outside of any other controller, the object may either be manually
```
johnmcollier, line 87: Is there a risk of creating duplicate`Artifact`s that refer to the same object?

ADR/0035-artifact-streams.md
```diff
@@ -0,0 +1,150 @@
...
+
+All `Artifact`s have an owner reference to a single `ArtifactStream`. The `ArtifactStream` keeps track of the latest
+produced `Artifact` that has been created owned by that stream (i.e. similar to a floating tag reference).
+
+A validation webhook will prevent an `Artifact` from being created unless its configuration matches that of its owning
+`ArtifactStream`.
+
+### ArtifactCollection
+
+An `ArtifactCollection` is a set of `Artifact`s that are grouped together. An `ArtifactCollection` can include
```
johnmcollier, line 107: What would a typical grouping of `ArtifactCollections` contain, in terms of types of `Artifact`s?


And could we achieve a similar level of grouping with labels on the resourecs, rather than creating another resource type to group them? Maybe something like:
```
labels:
   artifactCollection: <some-collection>
```

ADR/0035-artifact-streams.md
```diff
@@ -0,0 +1,150 @@
...
+## Decision
+
+We are going to abstract the concepts of the built artifacts and their relationships to each other outside of
+any other controller within AppStudio. If a controller wants to maintain complete separation from others (i.e.
+in order to be fully deployable on its own), it may use an adapter pattern to convert this abstraction to its
+own relevant internal data model.
+
+### ArtifactConfig
+
+An `ArtifactConfig` describes the metadata which is used to define an `Artifact`, an `ArtifactStream`, and
```
johnmcollier, line 63: What metadata would be used to define an Artifact type?

And agreed with other comments re: example CRs 🙂

ADR/0035-artifact-streams.md
```diff
@@ -0,0 +1,150 @@
...
+## Consequences
+
+- If any controller does not want to interface with the `Artifact*` CRs directly (i.e. to achieve independence
+  from other controllers within AppStudio), they will need to create an adapter pattern between the relevant
+  APIs and their own internal data model.
+- We should now have native ability to describe built artifact and their various supply chain related attestations.
+  This ability holds whether the artifact is built within the kubernetes cluster or outside of it.
+- We lose the simplification of having a common [Component] and [Application] data model requirement between
+  all controllers as the relationships with `Artifact`s and related resources are loose couplings
+- We will have to produce multiple adapters for the various controllers in order to achieve data model independence
```
johnmcollier, line 131: This would probably be my biggest concern, both from the amount of work / maintenance required, but also in terms of consistency. Do we need to rethink 

You mention: 
> While each of the controllers that compose AppStudio should be able to exist on their own, they can all
  "come together" around a common description of produced artifacts in order to create a unified build, test,
  and delivery pipeline

How would we ensure consistency across the various controllers that would need to implement an adapter? Having the common Application-Component data model helped to enforce that _somewhat_ (though we occasionally still had times where two controllers interpreted things differently), but that wouldn't be the case here.

johnmcollier (commented):
Generally liked the proposal here. Didn't have a ton of new comments compared to what was already left, so just left a few.

I think having some example CRs and a proposed implementation timeline/process would help a lot here.

arewm:
I am closing this proposed ADR. I think that we still need something to replace Components and Applications but the current proposal will not be sufficient for our use cases.

PR #168, opened 2024-02-14
[KONFLUX-179]: ADR for provisioning test resources

amisstea:
Here's ADR for how we can provision clusters for OLM operator testing.

### DIFF
--- a/ADR/0033-provisioning-test-resources.md
+++ b/ADR/0033-provisioning-test-resources.md
@@ -1,0 +1,279 @@
+# 33. Provisioning Test Resources
+
+Date: 2024-02-14
+
+## Status
+
+Accepted
+
+Supersedes:
+
+- [ADR 08. Environment Provisioning](0008-environment-provisioning.html)
+- Environment provisioning parts of [ADR 32. Decoupling Deployment](0032-decoupling-deployment.html)
+
+## Context
+
+This decision clarifies how integration test environments will be dynamically provisioned. In prior
+decisions it was believed that [Dynamic Resource Allocation] (DRA) would graduate out of OpenShift
+TechPreview on a timeline suitable for this project. This is no longer the case and, as such, a new
+approach to requesting compute resources for test pipelines is needed. DRA is still expected to
+become the kubernetes-native approach to managing dynamically provisioned resources across pods.
+Therefore, any interim solution should intentionally avoid introducing new barriers that might
+prevent the adoption of DRA some day.
+
+The problem of provisioning test resources/environments can be broken down into a few questions:
+
+1. How can resources (OpenShift clusters, to start) be provisioned, efficiently, without exposing
+   cloud/infra credentials to the end user?
+1. How does the user declare ephemeral resources and inject their resulting secrets into
+   `TaskRuns`/`Pods`?
+1. How can ephemeral resources be shared across `TaskRuns`/`Pods`?
+
+## Decision
+
+### Cluster Provisioning
+
+[Hive] will be used to provision OpenShift clusters. It's widely used in OpenShift CI testing,
+supports hibernated cluster pools for quick (5-7m), cost-efficient, allocation and is maintained
+and distributed by Red Hat on OperatorHub. [Cluster API] is an intriguing option for provisioning
+Kubernetes and OpenShift clusters but expected to remain in OpenShift TechPreview throughout 2024.
+It also recommends using a dedicated management cluster separate from application workloads which
+creates additional infrastructure complexity.
+
+Hive `ClusterPools` and `ClusterImageSets` will be maintained by Konflux admins (possibly with the
+aid of dependabot) and deployed to a single Konflux-managed namespace. `ClusterClaims` must reside
+in the same namespace as the `ClusterPools` but that's also where the cloud secrets are stored.
+Furthermore, the user cannot be allowed to create claims in their tenant namespace since it will
+only generate errors from Hive due to missing cluster pools. We could grant each tenant's pipeline
+service account the permission to create claims in the protected namespace but that opens up an
+attack surface where the pool of clusters could be starved by scripted creation of claims.
+Alternatively, something else could create the claims in the protected namespace on behalf of the
+user. A DRA provider would have been this actor.
+
+### Declaring Ephemeral Resources
+
+A `provision.konflux.dev/cluster-claims` annotation on integration `Pipelines`/`PipelineRuns` will
+be used to signal the intent to create claims for ephemeral OpenShift clusters. The value of this
+annotation is a reference to a `ConfigMap` in the same namespace. The `ConfigMap` keys will be used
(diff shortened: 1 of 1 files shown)


ADR/0033-provisioning-test-resources.md
```diff
@@ -0,0 +1,279 @@
...
+
+It may be possible to replace this annotation approach if Tekton ever adds support for declaring DRA
+`ResourceClaimTemplates` on a `Pipeline`. However, this isn't on any roadmap at the time of this
+writing.
+
+### Sharing Ephemeral Resources Across `TaskRuns`/`Pods`
+
+The controller will also watch the status of `ClusterClaims` as they are allocated. Upon success,
+the secret for the `ClusterDeployment` (containing a kubeconfig) will be copied to the tenant
+namespace using the same name as the `ClusterClaim`. Using this pattern allows `Pipeline` authors to
```
stuartwdouglas, line 85: Does this mean that only a single pipeline can be running at a time? Otherwise they would both get the same cluster. 

The multi-arch controller uses a similar pattern but uses the task/pipeline name in the secret, so every secret name is unique.

ADR/0033-provisioning-test-resources.md
```diff
@@ -0,0 +1,279 @@
...
+service account the permission to create claims in the protected namespace but that opens up an
+attack surface where the pool of clusters could be starved by scripted creation of claims.
+Alternatively, something else could create the claims in the protected namespace on behalf of the
+user. A DRA provider would have been this actor.
+
+### Declaring Ephemeral Resources
+
+A `provision.konflux.dev/cluster-claims` annotation on integration `Pipelines`/`PipelineRuns` will
+be used to signal the intent to create claims for ephemeral OpenShift clusters. The value of this
+annotation is a reference to a `ConfigMap` in the same namespace. The `ConfigMap` keys will be used
```
ifireball, line 57: What do we gain by introducing indirection via a ConfigMap? Could we make the annotation contain the name of a cluster pool directly instead?

ifireball (changes requested):

ADR/0033-provisioning-test-resources.md
amisstea: No, there can be many claims for a given PipelineRun. It's mentioned earlier that the claims would include the PipelineRun name (`<PipelineRunName>-<ConfigMapKey>`).

ADR/0033-provisioning-test-resources.md
amisstea: The main motivation behind the `ConfigMap` was to allow the user to customize some aspects of the `ClusterClaim`. That being said, the only field in the claim spec which might be of interest is `lifetime` and we may want to restrict what can be provided there anyway.

Looking to a future where DRA gets adopted, `ResourceClaims` have a ref to a separate `ResourceClaimParams` resource so a ConfigMap wouldn't be necessary in this case either (the user would have to be allowed to create `ResoureClaimParams`).

In order to drop the ConfigMap, we need the annotation to reference multiple `ClusterPools` while still defining some tokens (e.g. `cluster1`)  the author can use to reference the generated claims. We could do that with a bit of json/yaml or some key=value like syntax. For example:

```yaml
provision.konflux.dev/cluster-claims: |
  {
    "cluster1": "openshift-latest-ga-x86",
    "cluster2": "openshift-4.13-x86"
  }
```

```yaml
provision.konflux.dev/cluster-claims: |
  cluster1: openshift-latest-ga-x86
  cluster2: openshift-4.13-x86
```

```yaml
provision.konflux.dev/cluster-claims: cluster1=openshift-latest-ga-x86,cluster2=openshift-4.13-x86
```

ifireball: How often to we expect users to actually use multiple clusters? I'd rather see use use the Pareto principle to make the common use case dead simple while accenting some extra complexity for user use cases.

For example, we could make a very simple annotation naming a cluster pool for the single use case and a more complex one referring to a config map for more complex use cases. for example:

```
provision.konflux.dev/cluster-claim: openshift-latest-ga-x86
```

V.s.
```
provision.konflux.dev/cluster-claim-config: name-of-my-config-map
```

ifireball: Using a config map does prevent the user from fully defining thei pipeline via PAC. So we may consider one of the nested syntaxes you have suggested.

dirgim: Based on how operator testing is done currently in CVP, users almost always utilize multiple clusters in order to test compatibility with all OpenShift versions that their operator is supposed to support.

ifireball: @dirgim ok, but does that have to be in the same pipeline? could we possibly have a different pipeline per-version?

ADR/0033-provisioning-test-resources.md
```diff
@@ -0,0 +1,279 @@
...
+### Cluster Provisioning
+
+[Hive] will be used to provision OpenShift clusters. It's widely used in OpenShift CI testing,
+supports hibernated cluster pools for quick (5-7m), cost-efficient, allocation and is maintained
+and distributed by Red Hat on OperatorHub. [Cluster API] is an intriguing option for provisioning
+Kubernetes and OpenShift clusters but expected to remain in OpenShift TechPreview throughout 2024.
+It also recommends using a dedicated management cluster separate from application workloads which
+creates additional infrastructure complexity.
+
+Hive `ClusterPools` and `ClusterImageSets` will be maintained by Konflux admins (possibly with the
```
Dannyb48, line 43: IMHO, what I like about the openshift-ci platform approach is it lets me do a "bring your own cluster pool config".  Some teams might have different cluster configurations than others and may need different flavor sizes. 

What is not clear is that I can't tell if that is still possible and I just need to had these configs over to a Konflux Admin to configure which is fine or if Konflux is predefining and there isn't many options?

ADR/0033-provisioning-test-resources.md
```diff
@@ -0,0 +1,279 @@
...
+decisions it was believed that [Dynamic Resource Allocation] (DRA) would graduate out of OpenShift
+TechPreview on a timeline suitable for this project. This is no longer the case and, as such, a new
+approach to requesting compute resources for test pipelines is needed. DRA is still expected to
+become the kubernetes-native approach to managing dynamically provisioned resources across pods.
+Therefore, any interim solution should intentionally avoid introducing new barriers that might
+prevent the adoption of DRA some day.
+
+The problem of provisioning test resources/environments can be broken down into a few questions:
+
+1. How can resources (OpenShift clusters, to start) be provisioned, efficiently, without exposing
```
Dannyb48, line 26: What is also not clear to me is if it is going to be a "Bring your own cloud credentials" model? 

I think it should be. That is the model DPTP is moving to and I believe our internal Resource Hub offering is as well.

amisstea:
Closing this ADR in favor of a different approach outlined in #172.

PR #171, opened 2024-03-05
ADR to enable chaos testing of apps

yogananth-subramanian:
The objective of this ADR is to propose a way to enable the end user to run chaos tests for his app as part of integration testing.


ADR/0033-apps-continuous-chaos-testing.md
```diff
@@ -0,0 +1,47 @@
...
+
+Be proactive vs reactive. We have different types of test suites in place - unit, integration and end-to-end - that help expose bugs in code in a controlled environment. Through implementation of a chaos engineering strategy, we can discover potential causes of service degradation. We need to understand the systems’ behavior under unpredictable conditions in order to find the areas to harden, and use performance data points to size the clusters to handle failures in order to keep downtime to a minimum.
+
+### Glossary
+
+- krkn: Chaos testing framework: <https://github.com/krkn-chaos/krkn>
+
+## Decision
+
+There are two approaches to chaos testing in the CI/CD pipeline.
```
ralphbean, line 25: This is the first sentence of the decision section. To make it super clear, can you reword it as a decision? For example, is this the decision you're seeking agreement on?

> We are going to adopt chaos testing as a part of the [e2e-tests](https://github.com/redhat-appstudio/e2e-tests/) suite, and execute those tests to gate merges to main in the [infra-deployments](https://github.com/redhat-appstudio/infra-deployments) repo.

Or, is the decision something else, like:

> Because we think we will learn something from krkn, we are going to run some experiments with krkn and generate a report on where the weak areas in konflux are that need to be hardened. Depending on the result of those tests, we may submit a followup ADR recommending that we integrate krkn in our e2e-tests to gate merges to main in infra-deployments.

Or, perhaps I misunderstand, and the ADR is a recommendation to adopt krkn _as a part of the product_, perhaps in [integration-service](https://github.com/redhat-appstudio/integration-service)?  Are you arguing here that the decision should be something like:

> We recommend krkn as a testing tool for users of Konflux to use in their IntegrationTestScenarios, so that our users can reap the benefits of chaos testing on their applications - orchestrated by the integration-service in the konflux platform.

At the moment, this ADR reads more like an informative blog post (and, it is good and informative) but it doesn't feel like a decision to be recorded.

ralphbean (changes requested):
The context around the usefulness of chaos testing here is good, but kind of general. I'm not sure how this applies to konflux concretely. Can you refactor the ADR to be more specific about the decision and the impacts you think that it will have on konflux and the konflux development process?

yogananth-subramanian:
> refactor the ADR
Hello @ralphbean 
I apologize for being unclear.  The purpose of the ADR is to allow Konflux users to run chaos tests as part of IntegrationTestScenarios.. As part of the ADR, we would like to propose that the service account created as part of integration-service be used to access thanos-querier-openshift-monitoring or prometheus-k8s-openshift-monitoring endpoints. The change it would cause would be that the service account would now be able to run promQL queries.
Let me change the ADR and update the proposal.

yogananth-subramanian:
cc @psuriset @chaitanyaenr

yogananth-subramanian:
> The context around the usefulness of chaos testing here is good, but kind of general. I'm not sure how this applies to konflux concretely. Can you refactor the ADR to be more specific about the decision and the impacts you think that it will have on konflux and the konflux development process?

Hello @ralphbean 
I apologize for being unclear.  The purpose of the ADR is to allow Konflux users to run chaos tests as part of IntegrationTestScenarios.. As part of the ADR, we would like to propose that the service account created as part of integration-service be used to access thanos-querier-openshift-monitoring or prometheus-k8s-openshift-monitoring endpoints. The change it would cause would be that the service account would now be able to run promQL queries.
Let me change the ADR and update the proposal.

ADR/0033-apps-continuous-chaos-testing.md
yogananth-subramanian: Hello @ralphbean I have refactored the ADR based on your review comments. I would appreciate it if you could review the latest changes and provide your feedback.

ralphbean:
@amisstea, @dirgim can I can your review on this? It is related to both integration service and https://github.com/redhat-appstudio/architecture/pull/172

ADR/0033-apps-continuous-chaos-testing.md
ralphbean: Much clearer now - thank you!

ADR/0033-apps-continuous-chaos-testing.md
```diff
@@ -0,0 +1,41 @@
...
+
+- krkn: Chaos testing framework: <https://github.com/krkn-chaos/krkn>
+
+## Decision
+
+The Knoflux user can run chaos tests such as Krkn as part of the IntegrationTestScenarios in the ephemeral environment and can gather Prometheus metrics from Thanos-querier-openshift-monitoring or Prometheus-k8s-openshift-monitoring endpoints.
+
+## Consequences
+
+The service account associated with the ephemeral environment will possess admin-level privileges to execute CRUD operations within the ephemeral namespace space. Using the service account token, it will be possible to authenticate and query Prometheus metrics from Thanos-querier-openshift-monitoring or Prometheus-k8s-openshift-monitoring endpoints.
```
amisstea, line 41: Which resources will the chaos test need CRUD access to? We should call them out here.

Also, I believe granting the SA access to query metrics will allow (with a little "know how") the end user to access all metrics collected by that prometheus instance, not just those produced for the temp namespace. Are we sure we want to grant such access? We haven't exposed prometheus metrics to end users yet and this isn't taking the least privileged approach.

yogananth-subramanian: Hello @amisstea,
At present the serviceaccount associated with the ephemeral environment, seem to have all the required CURD access for pods,services, networkpolicies, deployments, replicastes, route.
So no additional modification would be necessary.
| **Resources**                      | **Non-Resource URLs** | **Resource Names** | **Verbs**                                                    |
|------------------------------------|-----------------------|--------------------|--------------------------------------------------------------|
| pods                               |         []            |       []           | [get list watch create delete deletecollection patch update] |
| services                           |         []            |       []           | [get create list watch delete deletecollection patch update] |
| networkpolicies.networking.k8s.io  |         []            |       []           | [create delete deletecollection patch update get list watch] |
| replicasets.apps                   |         []            |       []           | [create delete deletecollection patch update get list watch] |
| deployments.apps                   |         []            |       []           | [get list watch create delete deletecollection patch update] |
| routes                             |         []            |       []           | [create delete deletecollection get list patch update watch] |


If there exists a less privileged method for querying Prometheus metrics, I would be more than happy to provide support for such an implementation.Since it would not just benefit this ADR but also the  [0018-apps-continuous-perf-testing ](ADR/0018-apps-continuous-perf-testing.md)ADR related to performance.
The downside of not enabling the SA to query the Prometheus endpoints would be that we would miss out on executing the SLA-based Chaos scenario, but we should still be able to run the Resilience-based Chaos scenario on Konflux.

amisstea: I'm not aware of a way to make this less privileged. If I'm not mistaken, limiting prometheus access to the service account used in the krkn pod won't prevent an informed user from utilizing that SA in other workloads. @yftacherzog @avi-biton do you have any ideas?

@yogananth-subramanian please add a little more detail here about how this change will grant the service account unfettered access to query any metric within the cluster, not just those in the ephemeral namespace.

yftacherzog: What are the metrics that should be collected here? Is it some custom metrics generated by exporter on the same namespace or generic metrics generated by KSM and such?
Maybe the overhead of deploying an ephemeral prometheus instance inside the namespace could make sense?

amisstea: We discussed this problem during the arch call yesterday and reached a consensus that chaos tests should utilize ephemeral clusters (see #172) rather than ephemeral namespaces for better isolation and a more production-like testing environment. The end user will have cluster admin access on such clusters to configure RBAC permissions, prometheus instances, etc.

@yogananth-subramanian, at your convenience, please update this ADR to reflect the decision.

ralphbean: @yogananth-subramanian have you had a chance to look into update here?

yogananth-subramanian:
Hello @ralphbean @amisstea @dirgim I have revised the ADR based on the discussion during the arch call to incorporate the utilization of ephemeral clusters.
It would be wonderful if you could provide your input on the updated ADR.

ADR/0035-apps-continuous-chaos-testing.md
```diff
@@ -0,0 +1,44 @@
...
+* The user account shall have the capability to query the `Thanos-querier-openshift-monitoring` or `Prometheus-k8s-openshift-monitoring` route from the `openshift-monitoring` namespace. The account should have at least `view` level access to the `openshift-monitoring` namespace.
+* An additional openshfit feature that is recommended, though not mandatory, is the provision of [monitoring for user-defined projects](https://docs.openshift.com/container-platform/4.15/observability/monitoring/enabling-monitoring-for-user-defined-projects.html#accessing-metrics-from-outside-cluster_enabling-monitoring-for-user-defined-projects). `Monitoring-rules-edit` and `monitoring-edit` would be associated with the account to perform CURD operations related to `PrometheusRule`, `ServiceMonitor` and `PodMonitor` resources.
```
amisstea, line 44: This point seems more like an implementation detail rather than a consequence. I think it fits better under the decision section as an option to consider.

ADR/0035-apps-continuous-chaos-testing.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Consequences
+
+* The service/user account associated with the ephemeral environment will have cluster admin level privileges to execute CRUD operations (configure RBAC permissions, prometheus instances) within the ephemeral environment. At a minimum, the user should be granted administrator level privilege for a namespace on the ephemeral cluster.
+* Using the service account token, it will be feasible to authenticate and query Prometheus metrics from either `Thanos-querier-openshift-monitoring` or `Prometheus-k8s-openshift-monitoring` endpoints. The privilege could be granted by assigning a `cluster-monitoring-view` role to the associated account, which would permit the user to query prometheus metrics across all namespaces in the cluster.
+* The user account shall have the capability to query the `Thanos-querier-openshift-monitoring` or `Prometheus-k8s-openshift-monitoring` route from the `openshift-monitoring` namespace. The account should have at least `view` level access to the `openshift-monitoring` namespace.
```
amisstea, line 43: At what stage will this monitoring stack be configured? Will it be performed from a Tekton task prior to running krkn?

yogananth-subramanian: Hello @amisstea , as you have indicated, the plan is to use Tekton task to run the checks and setup before the tasks related to Chaos(krkn) are run. Therefore, a Chaos pipeline would comprise a task to verify if a user account possesses cluster admin access, and in the event that the monitoring stack [monitoring for user-defined projects](https://docs.openshift.com/container-platform/4.15/observability/monitoring/enabling-monitoring-for-user-defined-projects.html#accessing-metrics-from-outside-cluster_enabling-monitoring-for-user-defined-projects)  is not configured, enable monitoring. Subsequently, the Choas task would be initiated.

ADR/0035-apps-continuous-chaos-testing.md
yogananth-subramanian: Hello @amisstea , thank you for the input, will move it to the decision section.

yogananth-subramanian:
Hello @amisstea , I have revised the ADR based on your suggestion and have also taken out some technical implementation that might not be relevant for the decision-making.
It would be wonderful if you could provide your input on the updated ADR.

ADR/0033-apps-continuous-chaos-testing.md
amisstea: The end user will have cluster-admin permissions on the ephemeral clusters so there's no need to call out the specific resource requirements. Resolving this thread.

amisstea (approved):
/lgtm

@yogananth-subramanian you may want to update the date and status at the top before merging.

ADR/0035-apps-continuous-chaos-testing.md
```diff
@@ -0,0 +1,49 @@
...
+
+### Glossary
+
+- krkn: Chaos testing framework: <https://github.com/krkn-chaos/krkn>
+
+## Decision
+
+* Users can leverage Krkn, a chaos testing framework, to execute chaos tests within the IntegrationTestScenarios.
+
+* Ephemeral clusters ([provisioning-ephemeral-openshift-clusters](https://github.com/konflux-ci/architecture/pull/172)) will used, to provide a more isolated and production-like testing environment compared to ephemeral namespaces.
```
dirgim, line 39:
```suggestion
* Ephemeral clusters ([provisioning-ephemeral-openshift-clusters](https://github.com/konflux-ci/architecture/pull/172)) will be used for executing the tests, to provide a more isolated and production-like testing environment compared to ephemeral namespaces.
```

dirgim (approved):
LGTM aside the tiny nit

ADR/0035-apps-continuous-chaos-testing.md
yogananth-subramanian: Hello @dirgim , Thank you for catching the issue. Have fixed it as per your suggestion in the latest commit.

dirgim: Thanks, looks good!

yogananth-subramanian:
Hello @amisstea @dirgim @ralphbean I would like to thank you for your patience review and helpful comments.
Have updated the dat to today's date and set the status to `Accepted`.

ralphbean (approved):

PR #176, opened 2024-04-01
doc(RHTAPWATCH-820): Project* permissions

ifireball:
Update the roles and permissions ADR to include the permissions for the new Project Controller.

Should probably only be merged after #166 and redhat-appstudio/docs.appstudio.io#248 are merged.

### DIFF
--- a/ADR/0011-roles-and-permissions.md
+++ b/ADR/0011-roles-and-permissions.md
@@ -38,6 +38,7 @@ We will use the built-in Kubernetes RBAC system for AppStudio's role and permiss
 |               | *Service Access*        | appstudio.redhat.com      | get, list, watch                                | spiaccesstokenbindings, spiaccesschecks, spiaccesstokens, spifilecontentrequests
 |               | *Remote Secrets*        | appstudio.redhat.com      | get, list, watch                                | remotesecrets
 |               | Build Service           | appstudio.redhat.com      | get, list, watch                                | buildpipelineselectors
+|               | Project Controller      | projctl.konflux.dev       | get, list, watch                                | projects, projectdevelopmentstreams, projectdevelopmentstreamtemplates
 |               | *Configs*               |                           | get, list, watch                                | configmaps
 |               | *Secrets*               |                           |                                                 | secrets
 |               | Add User                |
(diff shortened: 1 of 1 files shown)


amisstea (approved):
LGTM

ADR/0011-roles-and-permissions.md
```diff
@@ -58,6 +59,7 @@ We will use the built-in Kubernetes RBAC system for AppStudio's role and permiss
 |               | *Service Access*        | appstudio.redhat.com      | get, list, watch, create, update, patch         | spiaccesstokenbindings, spiaccesschecks, spiaccesstokens, spifilecontentrequests, spiaccesstokendataupdates
 |               | *Remote Secrets*        | appstudio.redhat.com      | get, list, watch                                | remotesecrets
 |               | Build Service           | appstudio.redhat.com      | get, list, watch, create                        | buildpipelineselectors
+|               | Project Controller      | projctl.konflux.dev       | get, list, watch, create, update, patch, delete | projects, projectdevelopmentstreams, projectdevelopmentstreamtemplates
```
arewm, line 62: Do you want to grant `delete` permissions to maintainers? Maintainers do not have those permissions for Components or Applications.

arewm (changes requested):

ADR/0011-roles-and-permissions.md
ifireball: We did grant them in practice for now, so might as well have the docs reflect that. We can consider changing it later.

arewm: If that is the current state, would you mind including that update into this PR as well since it relates to the new permissions added?

ifireball: @arewm what changes should I make, if any?

ifireball: @arewm no, this PR reflects what the permissions for the project controller are ATM. Thos are, as you noted, slightly different the the permissions for Applications and Components

ifireball:
@arewm can you move your review from "request changes" to "approved" if this does look good to you?

arewm (approved):
Approving as this is the current state. We should open a bug to remove the ability for maintainers to delete components and applications.

PR #180, opened 2024-04-23
Replace most mentions of AppStudio with Konflux

ralphbean:


arewm:
Do you want to change the relevant file names or references within the images as well?

ralphbean:
@arewm hm, yes. Let's do that here.

ralphbean:
@arewm I updated to include diagram changes now and a few filename changes too.

arewm (approved):
I trust what you have done and won't double-check it. We can continue to replace references as we discover them.

dirgim (approved):
LGTM

PR #183, opened 2024-04-25
Restructure for hugo

ralphbean:


ralphbean:
I'm seeing really weird failures here with hugo that I don't understand.

The `theme` is currently missing. Clone it into the themes directory first.

Then, run `hugo server` to build and server locally.

Notice that these urls work:

* http://localhost:1313/decisions/ - the list of adrs
* http://localhost:1313/decisions/0000-adr-template/ - individual adrs
* http://localhost:1313/architecture/image-controller/ - individual design docs

But this url does _not_ work:

* http://localhost:1313/architecture/ - the list of design docs.

That last one should be an index of the design documents... but, it is empty. I can make it render with a list of design docs if I _remove_ what seems to be an arbitrary number of the _ADRs_ in the `decisions/` dir, which just makes no sense at all.

hugo.toml
```diff
@@ -0,0 +1,42 @@
...
+# not pluralize title pages by default
+pluralizelisttitles = false
+
+[params]
+    disabled_logo = false
+    logo_text = "Konflux"
+
+    # Style options: default (light-blue), blue, green, marsala, pink, red, turquoise, violet
+    style = "default"
+
```
avi-biton, line 24: Try adding these lines under [params] to be able to open 
http://localhost:1313/architecture/
and  http://localhost:1313/decisions

```
    mainSections = ["architecture", "decisions"]
    defaultDescription = "Site template made by Devcows using Hugo"
```

zregvart:
Also have a look at #184

ralphbean:
> Do we need the the HAS documentation?
@avi-biton let's keep it in for the restructuring. Removing it should be documented as a decision on its own.

ralphbean:
Now that the architecture repo is building successfully at https://konflux-ci.dev/architecture/ with jekyll, I'm not sure I have the motivation to finish this PR - there's still a good bit more cleanup to do to get the style to match https://konflux-ci.dev/

Do either of you want to pick up the remaining work in this PR? If so, feel free to, otherwise I think I'll drop it.

zregvart:
> Do either of you want to pick up the remaining work in this PR? If so, feel free to, otherwise I think I'll drop it.

Beside my bias towards Hugo, I'm not sure much is gained by transition from Jekyll to Hugo. Is there something we wish to gain with this?

ralphbean:
@zregvart I think the only thing at this point is alignment of the _style_ between https://konflux-ci.dev/ and https://konflux-ci.dev/architecture/

zregvart:
In https://github.com/konflux-ci/konflux-ci.github.io/pull/39 the custom CSS and images are pulled in to a theme that can be addressable via Hugo modules, i.e. changing the `theme` in `hugo.toml` something like:

```diff
diff --git a/hugo.toml b/hugo.toml
index b950f08..8172d71 100644
--- a/hugo.toml
+++ b/hugo.toml
@@ -1,7 +1,7 @@
 baseURL = 'http://localhost:1313/'
 languageCode = 'en-us'
 title = 'Standalone Konflux Architecture Docs'
-theme = 'github.com/devcows/hugo-universal-theme'
+theme = ['github.com/konflux-ci/konflux-ci.github.io/website/themes/konflux', 'github.com/devcows/hugo-universal-theme']
 
 # Site language. Available translations in the theme's `/i18n` directory.
 defaultContentLanguage = "en"

```

Would bring in the styling of [konflux-ci.dev](https://konflux-ci.dev/)

zregvart:
With #188 the overall styling of konflux-ci.dev website can be inherited here as well.

zregvart:
In #189 I've restructured the ADR decisions to add front matter and added a custom list layout for decisions, it renders a bit nicer now.

ralphbean:
@zregvart 00d755c66a03e90ac5c3be40da7d0d9147c83a3c had a side effect that I don't understand. Before it, urls like http://localhost:1313/architecture/release-service/ would render correctly, but after it they do not.

zregvart:
@ralphbean in #190 I've used `relref` for links and tweaked a number of links so that they point to the local content rather than the published site URL (used to be https://redhat-appstudio.github.io/architecture/). The link checker I use checks relative links and fragments, the only broken links with this seem to be references to removed objects from the API docs.

zregvart:
#191 was useful for previewing locally

PR #179, opened 2024-04-19
feat(STONEINTG-697): Update integraiton-service doc

hongweiliu17:
* Update integration-service architecture doc since integration don't create SEB for successful snapshot

Signed-off-by: Hongwei Liu <hongliu@redhat.com>

PR #184, opened 2024-04-26
Fixes for #183

zregvart:


zregvart:
@ralphbean here are some of the fixes for the issues you mention in the #179

PR #186, opened 2024-04-30
Add a PR size labeler

ralphbean:


arewm (commented):
The docs repo inherited a labler as well: https://github.com/konflux-ci/docs/blob/main/.github/workflows/size-label.yml

I think we should standardize on the action if possible (this PR looks like a smaller action length which is nice).

ralphbean:
@arewm from a quick look, do you have a preference on which way to go?

arewm (approved):
This one is fine!

PR #187, opened 2024-05-03
Workspace Experience

ralphbean:
This is just a stub for now - I'm hoping to create a place where we can have a discussion and collect input.

### DIFF
--- a/ADR/0034-workspace-experience.md
+++ b/ADR/0034-workspace-experience.md
@@ -26,6 +26,12 @@ Some relevant changes that have occurred or are ongoing.
 * We made the decision (at Red Hat) to stop trying to offer a managed service interface for Konflux. The user signup, waitlist, and workspace provisioning interfaces made a lot of sense there. Do they still make sense?
 * For [konflux-ci], we want everything to work on vanilla kubernetes - no OpenShift required. But, [kubesaw] brings along OpenShift Templates for workspace provisioning. Are we going to work around that?i
 
+### Some challenges
+
+* Recently, we find ourselves being asked by users to provide service accounts. Our existing proxy layer doesn't support service accounts and we're having to come up with workarounds - breaking the existing abstraction.
+* We find users requesting quota increases, but the tier system provided by our existing workspace quota management tooling doesn't permit us to adjust the quota of individual workspaces. We have to increase the whole tier, or add a new tier.
+* We worry that new contributors will find the proxy layer confusing and offputting - more complex than it needs to be. "Is something more than just kubernetes in play here?"
+
 ## Decision
 
 TBD - what do we want to do?
--- a/ADR/0034-workspace-experience.md
+++ b/ADR/0034-workspace-experience.md
@@ -1,0 +1,35 @@
+# 34. Workspace Experience
+
+## Status
+
+WiP
+
+## Context
+
+We inherited the "workspace" concept from our early days when we expected to be integrated with the [KCP] project. Long ago (over a year at the time of this writing), we decided to remove [KCP] from the architecture of the Konflux project - but, we have never revisited the workspace and defined and decided what we want it to be and what we want it to mean.
+
+Workspaces are an important concept in the Konflux user experience today. The user onboards into a workspace of their own. A team member of theirs may add them to a team workspace. As a platform administrator, if you hear about a user having trouble, you'll ask them what workspace they're working in.
+
+Workspace roles give a way to describe who has what rights in the workspace.
+
+Workspaces happen to be our mechanism for managing quota - a fact inherited from the [kubesaw] stack.
+
+Workspaces provide a kind of transparent multi-cluster for us. A user's workspace may be allocated to one cluster or another - but our proxy layer keeps the user blissfully unaware of this. Workspaces can be placed on clusters either automatically (using [kubesaw] mechanisms) or manually by a platform administrator.
+
+Workspaces also provide a kind of general user-management today. Workspaces and users are mixed up - we can't tell which workspaces correspond with real users and which workspaces are team-based workspaces. They all look like users. However, they _do_ give us a concrete representation of a user, which we can use to power the invite dialog to add other users to your workspace.
+
+### Some changes
+
+Some relevant changes that have occurred or are ongoing.
+
+* We made a decision (now long ago) to drop [KCP] from our architecture, which initially introduced the workspace concept. Both transparent multi-cluster as a feature and workspaces as a concept came from KCP. Are they still goals?
+* We made the decision (at Red Hat) to stop trying to offer a managed service interface for Konflux. The user signup, waitlist, and workspace provisioning interfaces made a lot of sense there. Do they still make sense?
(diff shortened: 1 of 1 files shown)


ADR/0034-workspace-experience.md
```diff
@@ -0,0 +1,41 @@
...
+
+* We made a decision (now long ago) to drop [KCP] from our architecture, which initially introduced the workspace concept. Both transparent multi-cluster as a feature and workspaces as a concept came from KCP. Are they still goals?
+* We made the decision (at Red Hat) to stop trying to offer a managed service interface for Konflux. The user signup, waitlist, and workspace provisioning interfaces made a lot of sense there. Do they still make sense?
+* For [konflux-ci], we want everything to work on vanilla kubernetes - no OpenShift required. But, [kubesaw] brings along OpenShift Templates for workspace provisioning. Are we going to work around that?i
+
+### Some challenges
+
+* Recently, we find ourselves being asked by users to provide service accounts. Our existing proxy layer doesn't support service accounts and we're having to come up with workarounds - breaking the existing abstraction.
+* We find users requesting quota increases, but the tier system provided by our existing workspace quota management tooling doesn't permit us to adjust the quota of individual workspaces. We have to increase the whole tier, or add a new tier.
```
alexeykazakov, line 32: Currently we, on the kubesaw side, are working on improving our tier mechanism. It will be much easier to add new tiers independently from the kubsaw codebase. I don't even see an issue with creating as many tiers as workspaces.

ADR/0034-workspace-experience.md
```diff
@@ -0,0 +1,41 @@
...
+
+Some relevant changes that have occurred or are ongoing.
+
+* We made a decision (now long ago) to drop [KCP] from our architecture, which initially introduced the workspace concept. Both transparent multi-cluster as a feature and workspaces as a concept came from KCP. Are they still goals?
+* We made the decision (at Red Hat) to stop trying to offer a managed service interface for Konflux. The user signup, waitlist, and workspace provisioning interfaces made a lot of sense there. Do they still make sense?
+* For [konflux-ci], we want everything to work on vanilla kubernetes - no OpenShift required. But, [kubesaw] brings along OpenShift Templates for workspace provisioning. Are we going to work around that?i
+
+### Some challenges
+
+* Recently, we find ourselves being asked by users to provide service accounts. Our existing proxy layer doesn't support service accounts and we're having to come up with workarounds - breaking the existing abstraction.
```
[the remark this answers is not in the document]
alexeykazakov: > I think the tokens are dependent of the endpoint you are using

Can you elaborate on this? What do you mean?

dperaza4dustbit:
In general, if we don't manage the workspace experience we have to assume user will be managing that, will be good to see how users are expected to manage their workspaces for the first release of konflux coming up without workspace support and then build the experience from there. My initial suggesting is to think of a workspace abstraction that can bring external dependencies like ArgoCD or KubeSaw as needed depending on the scale. This will keep a consistent experience for the user regardless if they are playing with Konflux in Kind or the are deploying on production with a multi-tenant and multi-cluster setup. 

Having two experiences or as a result two APIs defeats the purpose of allowing user to play and gain perspective on what to expect in production.

ADR/0034-workspace-experience.md
```diff
@@ -0,0 +1,41 @@
...
+
+Workspace roles give a way to describe who has what rights in the workspace.
+
+Workspaces happen to be our mechanism for managing quota - a fact inherited from the [kubesaw] stack.
+
+Workspaces provide a kind of transparent multi-cluster for us. A user's workspace may be allocated to one cluster or another - but our proxy layer keeps the user blissfully unaware of this. Workspaces can be placed on clusters either automatically (using [kubesaw] mechanisms) or manually by a platform administrator.
+
+Workspaces also provide a kind of general user-management today. Workspaces and users are mixed up - we can't tell which workspaces correspond with real users and which workspaces are team-based workspaces. They all look like users. However, they _do_ give us a concrete representation of a user, which we can use to power the invite dialog to add other users to your workspace.
+
+### Some changes
```
arewm, line 21: Another change is the fact that we are trying to make Konflux an upstream project. We now have to consider the mental model that new contributors and potential PEs need to contend with. This drives the challenge mentioned later where potential contributors will likely have much more familiarity with interacting with the core k8s concepts like users, serviceaccounts, roles, and rolebindings.

ADR/0034-workspace-experience.md
```diff
@@ -0,0 +1,41 @@
...
+Workspaces are an important concept in the Konflux user experience today. The user onboards into a workspace of their own. A team member of theirs may add them to a team workspace. As a platform administrator, if you hear about a user having trouble, you'll ask them what workspace they're working in.
+
+Workspace roles give a way to describe who has what rights in the workspace.
+
+Workspaces happen to be our mechanism for managing quota - a fact inherited from the [kubesaw] stack.
+
+Workspaces provide a kind of transparent multi-cluster for us. A user's workspace may be allocated to one cluster or another - but our proxy layer keeps the user blissfully unaware of this. Workspaces can be placed on clusters either automatically (using [kubesaw] mechanisms) or manually by a platform administrator.
+
+Workspaces also provide a kind of general user-management today. Workspaces and users are mixed up - we can't tell which workspaces correspond with real users and which workspaces are team-based workspaces. They all look like users. However, they _do_ give us a concrete representation of a user, which we can use to power the invite dialog to add other users to your workspace.
+
```
ifireball, line 20: Each user is allocated a workspace by default when they sign-up to Konflux for the first time. The cluster on which the workspace is allocated is decided upon when the sign-up is confirmed manually or when automated confirmation is set up. 

Team workspaces are created by creating shared "team accounts", and on-boarding them to Konflux.

ADR/0034-workspace-experience.md
```diff
@@ -0,0 +1,41 @@
...
+### Some changes
+
+Some relevant changes that have occurred or are ongoing.
+
+* We made a decision (now long ago) to drop [KCP] from our architecture, which initially introduced the workspace concept. Both transparent multi-cluster as a feature and workspaces as a concept came from KCP. Are they still goals?
+* We made the decision (at Red Hat) to stop trying to offer a managed service interface for Konflux. The user signup, waitlist, and workspace provisioning interfaces made a lot of sense there. Do they still make sense?
+* For [konflux-ci], we want everything to work on vanilla kubernetes - no OpenShift required. But, [kubesaw] brings along OpenShift Templates for workspace provisioning. Are we going to work around that?i
+
```
ifireball, line 28: Currently, each user signing up get their own workspace automatically. Should that still be the case? With so many workspaces, most of which go unused, its harder to monitor activity and resources can be wasted. Perhaps we should switch to only allowing team workspaces? Or perhaps allow personal workspaces but only on-demand and for a limited time?

We we do end up having only team workspaces, how to we associate people with their team workspaces? We have a UI for "inviting" people to a workspace by assigning them permissions. Perhaps we should enable users to brows existing workspaces and ask to be invited? Perhaps we should allow inviting people who are not (yet) Konflux users to a particular workspace?

ralphbean:
Dropping this and the conversation is being driven by Gal and David in separate documents now.

PR #190, opened 2024-05-14
Use `relref` for links

zregvart:
The benefit of using `ref`/`relref` is that broken links are detected at build time. The benefit of using `relref` is that links are relative, so files can be opened from the file system in the browser and still work.

I've used global search-replace a lot so mistakes could have been made, the only broken links that I could find are the links towards the missing fragments in the `/ref` (api docs), seems that the ADRs point to objects that are no longer present in the API.


ralphbean:
(I got distracted this week with some other work. I hope to return to this next week(!))

PR #194, opened 2024-05-27
Allow users to manage ImageRepository objects with the same access level as Components

mmorhun:
Since we are switching to the new way of managing image repositories in Konflux via a separate `ImageRepositopry` resource instead of `Component` annotations, we need to grant users permissions to work with the new resource. `ImageRepositopry` should be created together with the `Component` (by UI) in case user wants to use default image repository option.
The permissions are set in sync with the `Component` permissions of each role.

Corresponding permissions change: https://github.com/codeready-toolchain/host-operator/pull/1037


tkdchen (approved):

chmeliik (approved):
LGTM :+1:

arewm (approved):

ralphbean (approved):

PR #196, opened 2024-06-14
Update architecture docs to reflect current state of HAS

johnmcollier:
Updates the Konflux architecture docs to accurately represent the current state of HAS within Konflux: a deployment that runs Kubernetes webhooks for Applications and Component resources


ifireball (approved):
LGTM

elsony (changes requested):
There are other places in the doc that references HAS info, e.g. https://github.com/konflux-ci/architecture/blob/main/architecture/index.md. We should do a search for the entire doc to make sure we remove all references and avoid link breakage.

johnmcollier:
@elsony Good suggestion. I checked through the remaining docs for both `HAS`, `HAS/` and `application-service` and didn't find anything that was broken by the doc removal.

I do see some outdated references to HAS, so I updated those with one exception: The architecture diagram and explanation of CRs in index.md, which I left alone, as it's outside of the scope of these changes (and outside of our team's scope now).

elsony (approved):

PR #200, opened 2024-07-09
feat(STONEINTG-961): remove composite snapshots from adrs

jsztuka:
Since composite snapshot is being droped from integration-service codebase, it would be in our best interest to reflect that fact in ADRs.
Feel free to left any feedback regarding those updates.

### DIFF
--- a/ADR/0037-integration-service-promotes-to-GCL-immediately.md
+++ b/ADR/0037-integration-service-promotes-to-GCL-immediately.md
@@ -11,6 +11,10 @@ Supersedes:
 - [ADR 15. The Two-phase Architecture of the Integration Service](0015-integration-service-two-phase-architecture.html)
 - [ADR 16. Integration Service Promotion Logic](0016-integration-service-promotion-logic.html)
 
+Superseded by:
+
+- [ADR 38. Integration service removes composite snapshots and logic around them](0038-integration-service-composite-removal.html)
+
 ## Context
 
 In the initial implementation of the Integration Service, when a single component image is built, the
--- a/ADR/0038-integration-service-composite-removal.md
+++ b/ADR/0038-integration-service-composite-removal.md
@@ -1,0 +1,36 @@
+# 38. Integration service removes composite snapshots and logic around them
+
+* Date started: 2024-07-10
+
+## Status
+
+Accepted
+
+## Context
+
+Composite snapshots main goal was to prevent race condition when teams merged multiple PRs to multiple components
+of the same application at nearly the same time. This concept was confusing for users and we managed to simplify it
+by immediate promotion to GCL using override snapshots. Users also ran into the issue with GCL deadlock. In short, 
+components already in the GCL can cause the tests for a new component to fail if they are bundled into a snapshot. 
+If two or more components in an application run into this issue it can create a deadlock that prevents the release 
+of new versions of the application.
+
+
+## Decision
+
+Introduction of override snapshots should prevent this race condition and GCL deadlock with much simpler concept to understand,
+replacing the composite snapshots which served the same purpose.
+This decision led to removing of all logic regarding composite snapshots within integration-service codebase, since override snapshot
+solves same problems.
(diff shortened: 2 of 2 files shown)


arewm (changes requested):
@jsztuka , instead of modifying the prior ADR, can you create a new ADR to record this decision along with its motivation?

ADR/0038-integration-service-composite-removal.md
```diff
@@ -0,0 +1,31 @@
...
+## Status
+
+
+## Context
+
+Composite snapshots main goal was to prevent race condition when teams merged multiple PRs to multiple components
+of the same application at nearly the same time. This concept was confusing for users and we managed to simplify it
+by immediate promotion to GCL using override snapshots. Users also ran into the issue with GCL deadlock. This was 
+caused by multiple components present in GCL causing tests for a new component fail if they were bundled into a snapshot.
+This deadlock prevented the relase of new version of the application.
```
arewm, line 14: While I understand the GCL deadlock, I think that this part is confusing. 

Is it true that we would not reach deadlock if the required integration test scenarios _would have been passing_ on the GCL snapshot? We saw deadlock situations when users had an EC failure on multiple components in the GCL (i.e. the action of updating a single component in a composite snapshot would still result in a failure from a different component)?

Another situation that we could have encountered this problem is if there were two tightly coupled components where both needed to be tested together before it would be possible to get a passing ITS (i.e. if there were integration tests running).

ADR/0038-integration-service-composite-removal.md
```diff
@@ -0,0 +1,31 @@
...
+
+
+## Decision
+
+Introduction of override snapshots should prevent this race condition and GCL deadlock with much simpler concept to understand,
+basically replacing the composite snapshots logic.
```
arewm, line 20: Can you describe what an override snapshot is in any more detail?

ADR/0038-integration-service-composite-removal.md
```diff
@@ -0,0 +1,31 @@
+# 38. Integration service removes composite snapshots and logic around them
+
+* Date started: 2024-07-10
+
+## Status
+
```
arewm, line 6: You can put accepted into the status. It won't be "officially" accepted until the PR is merged.

ADR/0037-integration-service-promotes-to-GCL-immediately.md
```diff
@@ -20,7 +24,8 @@ that was newly built is updated/overwritten to complete the Snapshot creation.
 The Global Candidate List for the newly built component would only be updated once all required integration tests
 for the created Snapshot have passed successfully.
 See more about this in [ADR 16. Integration Service Promotion Logic](0016-integration-service-promotion-logic.html) and
-[ADR 15. The Two-phase Architecture of the Integration Service](0015-integration-service-two-phase-architecture.html).
+[ADR 15. The Two-phase Architecture of the Integration Service](0015-integration-service-two-phase-architecture.html) and
+[ADR 38. Integration service removes composite snapshots and logic around them](0038-integration-service-composite-removal.html).
```
arewm, line 28: I don't think you need to link to the ADR in the context, just in the superseded. The context is more for prior art/background as far as I understand it.

arewm (changes requested):

ADR/0037-integration-service-promotes-to-GCL-immediately.md
jsztuka: You are correct, let me remove it.

ADR/0038-integration-service-composite-removal.md
jsztuka: Yes that is correct, I tried to rephrase sentences about deadlock, let me know if that is making more sense now.

ADR/0038-integration-service-composite-removal.md
jsztuka: Added few sentences regarding override snapshot, let me know it that suffices.

arewm (approved):

ralphbean (approved):

PR #202, opened 2024-07-15
ADR 39: Propose an ADR to install Knative Eventing

skoved:


skoved:
This would not need to be done on Konflux production clusters _yet_, but at least installing knative eventing on testing clusters will allow integration to start if and when Konflux decides to integrate with KubeArchive

ADR/0039-install-knative-eventing.md
```diff
@@ -0,0 +1,33 @@
+# 39. Install Knative Eventing
+
+Created: 2024-07-15
+
+## Status
+
+Proposed
```
gbenhaim, line 7: I think that the ADR should describe how Konflux will integrate with Kubearchive. For example how we configure, cluster wide, which resources we want to offload from the cluster what is the api for querying them.

In addition, we need to decide  if it's going to be a hard dependency of Konflux (it means anyone who want to use Konflux will have to install KNative), including users that would like to try out kind on a local cluster such as Kind (for this use case we are striving to make the Konflux instllation as simple as possible).

In addition, all the requirement of Kubearchive should be listed, I believe it also requires a database (maybe other things I'm not aware of).

gbenhaim (changes requested):

ADR/0039-install-knative-eventing.md
brianwcook: @gbenhaim I advised @skoved to open an ADR _just for getting Knative added to Konflux_ so we could weigh the merits and issues separately from KubeArchive. If you think that is the wrong approach its probably on me.

gbenhaim: @brianwcook  why should we consider Knative separately from Kubearchive? do you see it (Knative) being used elsewhere in Konflux?

skoved: > I think that the ADR should describe how Konflux will integrate with Kubearchive. For example how we configure, cluster wide, which resources we want to offload from the cluster what is the api for querying them.
How to configure which resources will be archived cluster wide, is not set in stone yet on the KubeArchive side, but we do have namespace level configuration set

> In addition, we need to decide if it's going to be a hard dependency of Konflux (it means anyone who want to use Konflux will have to install KNative), including users that would like to try out kind on a local cluster such as Kind (for this use case we are striving to make the Konflux instllation as simple as possible).
I'll need to rely on you or other Konflux folks for this part. I'm not super familiar with all of the different components of Konflux and what they need to function correctly. If there are deployments of Konflux where leaving things like `PipelineRuns` on the cluster with using something like Tekton Results to store them off cluster then I would say that Knative Eventing does not need to be hard dependency. **HOWEVER**, KubeArchive will not function at all if Knative Eventing is not install. I also want to be clear that only Knative Eventing needs to be installed. KubeArchive **DOES NOT** depend on any other Knative projects

> In addition, all the requirement of Kubearchive should be listed, I believe it also requires a database (maybe other things I'm not aware of).
KubeArchive needs a PostgreSQL database. For production clusters Konflux will want to set up it's own database ahead of time. I'll add info about how that needs to be done. KubeArchive can be installed with a helm chart, and for non production deployments, that helm chart can optionally install a PostgreSQL database itself. KubeArchive will also need cert-manager installed ahead of time and I can add that to the ADR as well.

brianwcook: That is exactly what I want to discuss. Should knative _always_ be deployed? Should we encourage, or at least allow, Knative to be considered in new core feature designs? Or should it be an optional component which is only used when it is dragged in by something like KubeArchive?

skoved: I want to be clear that KubeArchive depends on Knative Eventing (which can be installed on its own), and **DOES NOT** depend on any other Knative Projects.

If the scope of the ADR should expand to include integration with KubeArchive, I can update this ADR to include all the steps required today for KubeArchive to be installed and configured

> I think that the ADR should describe how Konflux will integrate with Kubearchive. For example how we configure, cluster wide, which resources we want to offload from the cluster what is the api for querying them.

How to configure which resources will be archived cluster wide, is not set in stone yet on the KubeArchive side, but we do have namespace level configuration ironed out and the clusterwide configuration will like be very similar
```
---
# example archiving PipelineRuns in the namespace example
apiVersion: kubearchive.kubearchive.org/v1alpha1
kind: KubeArchiveConfig
metadata:
  name: example-kubearchiveconfig
  namespace: example
spec:
  resources:
    - apiVersion: tekton.dev/v1beta1
      kind: PipelineRun
```

> In addition, we need to decide if it's going to be a hard dependency of Konflux (it means anyone who want to use Konflux will have to install KNative), including users that would like to try out kind on a local cluster such as Kind (for this use case we are striving to make the Konflux instllation as simple as possible).

I'll need to rely on you or other Konflux folks for this part. I'm not super familiar with all of the different components of Konflux and what they need to function correctly. If there are deployments of Konflux where leaving things like `PipelineRuns` on the cluster with using something like Tekton Results to store them off cluster then I would say that Knative Eventing does not need to be hard dependency. **HOWEVER**, KubeArchive will not function at all if Knative Eventing is not install. I also want to be clear that only Knative Eventing needs to be installed. KubeArchive **DOES NOT** depend on any other Knative projects

> In addition, all the requirement of Kubearchive should be listed, I believe it also requires a database (maybe other things I'm not aware of).

KubeArchive needs a PostgreSQL database. For production clusters Konflux will want to set up it's own database ahead of time. KubeArchive can be installed with a helm chart, and for non production deployments, that helm chart can optionally install a PostgreSQL database itself. I'm not sure if Konflux would be interested in this approach for non-production clusters. KubeArchive will also need cert-manager installed ahead of time. If this ADR should expand in scope

ADR/0039-install-knative-eventing.md
```diff
@@ -0,0 +1,33 @@
...
+Results. To side step the issues with Tekton Results, Konflux may decide to integrate with
+[KubeArchive](https://github.com/kubearchive/kubearchive) to store PipelineRuns and potentially other kinds of resources
+on the cluster in a database for long term storage.
+
+## Decision
+
+To enable a smooth integration with KubeArchive in the future, Konflux will install the following KubeArchive dependecies
+in Konflux Clusters:
+
+* [Knative Evevnting](https://knative.dev/docs/eventing)
```
MartinBasti, line 22:
```suggestion
* [Knative Eventing](https://knative.dev/docs/eventing)
```

ADR/0039-install-knative-eventing.md
MartinBasti: Even unused features are using resources in dev clusters, konflux is beefy enough with mandatory parts IMO.

Does it bring any benefits or other use cases except being a dependency of kubearchive?

ADR/0039-install-knative-eventing.md
skoved: fixed the typo

skoved:
I've update the ADR to (hopefully) provide more clarity about why I'm proposing to have konflux install Knative Eventing and what Knative Eventing provides. This came from feedback during the community architecture call earlier today. I've also added additional details to what I think the consequences of installing Knative Eventing _might_ be

ADR/0039-install-knative-eventing.md
skoved: >  why should we consider Knative separately from Kubearchive? do you see it (Knative) being used elsewhere in Konflux?

> Does it bring any benefits or other use cases except being a dependency of kubearchive?

I've updated the ADR to include features of Knative Eventing as well as how it could be used by Konflux directly as well

> Should knative always be deployed?

At the very least, Knative **Eventing** must deployed anywhere that Konflux intends to deploy KubeArchive

ADR/0039-install-knative-eventing.md
```diff
@@ -0,0 +1,46 @@
...
+Knative Eventing can be installed on a cluster by running the following command
+
+```bash
+kubectl apply -f https://github.com/knative/eventing/releases/download/knative-v1.14.3/eventing-core.yaml`
+```
+
+## Consequences
+
+Konflux will install Knative Eventing. This will enable smooth integration with KubeArchive if and when Konflux decides
+to. This will require the Konflux installation process to be modified to include the installation if Knative Eventing.
```
maruiz93, line 42: I think there is a typo here:
```suggestion
to. This will require the Konflux installation process to be modified to include the installation of Knative Eventing.
```

skoved: fixed the typo

ADR/0039-install-knative-eventing.md
ifireball: We do have a lot of controllers that look at PipelineRun resources. So it can make architectural sese to use Knative eventing instead of the usual controller watch loops. The watch llops have one major advantage though, they are especially easy to code when you use Kubebuider. In Konflux we are very heavy users of Kubebuilder. When I checked a while ago, I couldn't find anything in the Knative world that comes close to being a comprehensive framework like Kubebuilder. Especially for eventing.

ADR/0039-install-knative-eventing.md
```diff
@@ -0,0 +1,46 @@
...
+resource specified on the cluster and create [Cloudevents](https://cloudevents.io) that contain the definition of the
+object on the cluster that has changed. KubeArchive subscribes to these Cloudevents from `ApiServerSource` which allows
+it to be notified of state changes for resources on the cluster.
+
+In addition to `ApiServerSource`, Knative Eventing also provides
+[`Brokers`](https://knative.dev/docs/eventing/brokers) and [`Channels`](https://knative.dev/docs/eventing/channels) to
+allow for persistence storage of Cloudevents.
+[Both `Brokers` and `Channels` provide ways handle Cloudevent delivery failure](https://knative.dev/docs/eventing/event-delivery).
+Additionally, `Brokers` use [`Triggers`](https://knative.dev/docs/eventing/triggers) provide a way to subscribe to and
+filter events, so that a Cloudevent consumer can specify the types of Cloudevents that it wants to receive.
```
ifireball, line 28: Nothing in this text seems to convey what @brianwcook said his purpose behind this discussion was.

cit1zen:
Serverless-eventing is one of the technologies we look into as a possible future basis of MintMaker (dependency management).

I think having it available in some form could open doors to considerable performance and resource utilization improvements across Konflux, especially after people earn some experience. This option becomes part of the architecture design possibilities when designing new features or redesigning existing ones.

ADR/0039-install-knative-eventing.md
skoved: If you are referring specifically to this paragraph I can remove.

If you are referring to the whole Context section, then I would disagree. It explains the motivation for wanting Knative Eventing installed.

ADR/0039-install-knative-eventing.md
skoved: I'm not as familiar with the inner workings of konflux as, so thank you for providing more insight.

Knative Eventing, more specifically cloudevents, may not have something like KubeBuilder, but the [cloudevents sdk](https://github.com/cloudevents/sdk-go) provides a really easy way to start receiving events from knative eventing. [This example is extremely simple and just printing out the event id](https://github.com/cloudevents/sdk-go?tab=readme-ov-file#receive-your-first-cloudevent) but i think it demonstrates the point. Also KubeBuilder is performing a lot boilerplate needed for starting the controller and being able to register it with Kubernetes. With Knative Eventing most of the that is offloaded to the configuration of ApiServerSource, to tell it which types of resources to create events for, and Triggers/Subscriptions so that the consumer only receives events that it is interested in. This lets the application consuming the events to only worry about the logic for processing the event and performing some set of required actions based on that event. Because of this difference I think it makes sense why there is not totally comprehensive framework. I do think that the cloudevents sdk provides comprehensive tools that make it incredibly simple to start receiving events so that can focus processing and reacting to them.

I don't think that eventing can completely replace controllers because controllers and eventing are meant to solve different problems. I do think that there are areas that could benefit from Knative Eventing. For example, a controller that is watch tekton PipelineRuns so that when they complete, some action can be performed like sending an email. I think that's an example where using eventing could make more sense. It would receive an event with a PipelineRun object as the payload, it would verify that the PipelineRun completed, grab what ever other data it needs from the object to fill in an email template and then send the email. A controller that has to synchronize the state of different resources on the cluster wouldn't necessarily benefit from something ApiServerSource as easily. However, it could send cloudevents when it performs different actions that other services could subscribe to instead of having to watch resources that this controller synchronizes.

skoved:
It's been a while since anything has happened on this PR, but I was busy with KubeArchive stuff in the second half of July and all of August.

I've updated the ADR to reflect the intent of this ADR as mentioned by @brianwcook at the July 16th Architecture meeting. My understanding of it is as follows.

To consider the usage of Knative-Eventing in Konflux separately from KubeArchive and determine if the usage of Knative-Eventing could make sense in Konflux.

If you believe that this is incorrect or incomplete, I'm willing to discuss this and update the ADR as needed. I would also appreciate if those involved in the discussion would take another look. I'm also willing to bring this as a topic for the architecture call again if folks think that it is necessary/helpful.

ralphbean:
Although we don't have it in the architecture repo docs yet, I think we are happy to use some new terms to put some boundaries around different parts of konflux:

* Konflux core - these are components that you can assume are always installed in an instance of konflux. Other components can always depend on them being present. Example, tekton.
* Konflux add-ons - these are components that may or may not be installed in an instance of konflux. Other components can depend on them being present, but they must also have defined behavior for when they are not present. Example: mintmaker.
* Konflux integrations - these are systems external to Konflux that some Konflux components may refer to, if so configured. Example: trustification. If configured, the Konflux UI will link to SBOMs in a trustification UI somewhere else.

What would the `# Decision` section look like here if written in those terms? Are we saying one of the following things?

* knative eventing is a konflux add-on. It might be installed. If installed, other add-ons can take advantage of it. Nothing in core is allowed to depend on it.
* knative eventing is in konflux core. It will always be installed, and everything is permitted to depend on it.
* knative eventing is in konflux core, and all subsystems *must* refactor themselves to depend on it.

skoved:
> knative eventing is in konflux core, and all subsystems must refactor themselves to depend on it.

I want to make clear that this is **NOT** what this ADR is proposing. It is not the goal to force anyone to use Knative-Eventing

With that out of the way:

> What would the `# Decision` section look like here if written in those terms? Are we saying one of the following things?
> 
>     * knative eventing is a konflux add-on. It might be installed. If installed, other add-ons can take advantage of it. Nothing in core is allowed to depend on it.
> 
>     * knative eventing is in konflux core. It will always be installed, and everything is permitted to depend on it.

This is something that I was thinking about, but I'm not sure that I can answer this myself. I'm not developing for konflux (directly). Knative-Eventing will be installed where ever KubeArchive gets installed. My assumption is that since KubeArchive would replace Tekton Results, KubeArchive and Knative-Eventing as a dependency for KubeArchive would fit into the same category. Unless Konflux wants to add Knative-Eventing to Konflux core. In [this comment](https://github.com/konflux-ci/architecture/pull/202#discussion_r1679772369) I list potential benefits to Knative-Eventing, but since I'm not directly involved in Konflux development, I cannot point to specific places in Konflux core that could benefit directly from Knative-Eventing. Based on the comment from @cit1zen it seems like MintMaker would be interested in Knative-Eventing if it was made available though

ralphbean:
So, (this is up for debate) I think we want KubeArchive to be a Konflux add-on. It _might_ be installed in an instance of Konflux, or it might not. If it is installed, then the UI can be configured to query it for old archived resources.

Today, I think tekton-results may actually be in konflux core, but (and this is up for debate) I think we may want to push it out to be a konflux add-on.

gbenhaim:
Additional concern that I have is that the in memory channel shouldn't be used for production purposes, so it would require users to also get a Kafka instance, and I would really prefer not to have hard dependency on Kafka.

gbenhaim:
> 2. > Additional concern that I have is that the in memory channel shouldn't be used for production purposes, so it would require users to also get a Kafka instance, and I would really prefer not to have hard dependency on Kafka.
> 
> For 2), while that's currently true if you plan to use Broker and Triggers and have reliable distributed persistence, to use only event sources, you can just use Eventing with no other dependencies. Related questions are: is Kafka and operating Kafka the problem or generally having another dependency? would you be more comfortable with a different system to store events? and if so, which one?
> 

@pierDipi yes my concern is requiring a potential Konflux user to also maintain/use managed Kafka instance. If there
is a lighter backed that can be used it would be great.

ggallen:
Given that https://github.com/konflux-ci/architecture/pull/206 has been approved, shouldn't this either be approved or closed?

PR #41, opened 2023-01-09
new ADR for a managed workspace tier

scoheb:

PR #206, opened 2024-09-24
ADR 41: Konflux should emit cloud events.

ggallen:


ADR/0039-send-cloud-events.md
```diff
@@ -0,0 +1,34 @@
...
+
+Proposed
+
+## Context
+
+Konflux had made the architectural decision to not use cloud events. However, that does not mean that
+Konflux should not emit cloud events.
+
+Emitting cloud events would allow Konflux users to easily track what is happening in the system. In addition,
+they can use these cloud events to create the own product-specific infrastructure to support their build
```
skoved, line 15:
```suggestion
they can use these cloud events to create their own product-specific infrastructure to support their build
```

ggallen: Fixed.

gbenhaim (approved):
LGTM. I just want to mention that Tekton has a feature for emitting cloud event (https://tekton.dev/vault/pipelines-v0.50.x-lts/additional-configs/#configuring-cloudevents-notifications), so at least reacting to the Konflux' pipelineruns and taskruns is already possible today.

ggallen:
> so at least reacting to the Konflux' pipelineruns and taskruns is already possible today.

The functionality is there in Tekton, but I don't believe it is enabled by default.  But one of the first things to do should this ADR be accepted is to turn on that functionality.

ADR/0039-send-cloud-events.md
```diff
@@ -0,0 +1,34 @@
...
+Konflux should not emit cloud events.
+
+Emitting cloud events would allow Konflux users to easily track what is happening in the system. In addition,
+they can use these cloud events to create their own product-specific infrastructure to support their build
+and release process.
+
+To support this, all Konflux components should be required to emit cloud events for signicant events. These
+should be documented fully and made available for users.
+
+Cloud event generation could be optional and that option could default to off. But users should be able to
```
lcarva, line 21: This probably needs clarification on what "users" are. I'm assuming this is a Konflux deployment admin, and not a user that uses Konflux to build content.

ADR/0039-send-cloud-events.md
```diff
@@ -0,0 +1,34 @@
...
+## Context
+
+Konflux had made the architectural decision to not use cloud events. However, that does not mean that
+Konflux should not emit cloud events.
+
+Emitting cloud events would allow Konflux users to easily track what is happening in the system. In addition,
+they can use these cloud events to create their own product-specific infrastructure to support their build
+and release process.
+
+To support this, all Konflux components should be required to emit cloud events for signicant events. These
```
lcarva, line 18: Let's consider if "should be required to emit" is too strong of a statement. I believe "should emit" better describes the intent.

ADR/0039-send-cloud-events.md
```diff
@@ -0,0 +1,34 @@
...
+
+Proposed
+
+## Context
+
+Konflux had made the architectural decision to not use cloud events. However, that does not mean that
+Konflux should not emit cloud events.
+
+Emitting cloud events would allow Konflux users to easily track what is happening in the system. In addition,
+they can use these cloud events to create their own product-specific infrastructure to support their build
```
lcarva, line 15: Can we use a different term other than "product-specific" throughout this document? AFAIK, Konflux is not aware of what a product is. Maybe use "application-specific"?

lcarva:
> The functionality is there in Tekton, but I don't believe it is enabled by default. But one of the first things to do should this ADR be accepted is to turn on that functionality.

However, the ADR mentions:

> Cloud event generation could be optional and that option could default to off.

Are you talking about turning on that functionality on a specific deployment of Konflux? That is perhaps outside the scope of these ADRs. As I see it, these are about the Konflux project, not necessarily any one deployment of Konflux.

ADR/0039-send-cloud-events.md
```diff
@@ -0,0 +1,34 @@
...
+itself. Rather it proposes Konflux generate cloud events to support addtional product-specific build and
+release functionality outside of Konflux.
+
+## Decision
+
+All Konflux components shall generate cloud events for significant events.
+
+## Consequences
+
+Product teams can more easily build product-specific build and release infrastructure in Konflux.
```
ralphbean, line 34: Can you add more to this section to make it more clear how, when, and why cloudevents would be used as opposed to extending build, test, or release pipelines?

ralphbean:
Some notes from the [architecture call today](https://www.youtube.com/watch?v=SAYfFlJrDxk).

* It should be possible to turn this off (and perhaps default to off) for the upstream installer to keep the footprint small.
* Tekton already exposes CloudEvents. Turning that on should just be a configuration option that someone running the platform can do today.
* There was some confusion about who the target audience for CloudEvents is. After discussion in the call (and some more afterwards) the intended audience is the end-user; the developer using Konflux to build their software.
* There's a note already in the [architecture constraints](https://github.com/konflux-ci/architecture/blob/main/architecture/index.md#architecture-constraints) that says "Use tekton for anything that should be extended by the user (building, testing, releasing)." If the purpose here is to give the end user a way to extend the system, that line needs to be revised.
* If we're revising that constraint, then we need some explanation of why, when, and how a user would choose to use cloudevents rather than extend their build, test, and release pipelines.

Arguably, anything a user might do with cloudevents they might also be able to do by extending their build, test, or release pipelines. If the idea here is that using cloudevents can be simpler for that user, I'd like to see in the Consequences section some elaboration on how that will work (what resources does the user need to create to take advantage of this, what RBAC do they need to achieve this).

ADR/0039-send-cloud-events.md
ralphbean: From async conversation - I think Greg means for this to be end-users. I.e., the developers that use Konflux to build, test and release their software. He doesn't mean platform engineers (the people running a konflux instance) or other konflux controllers.

ADR/0039-send-cloud-events.md
scoheb: I agree on the "softening"

ADR/0039-send-cloud-events.md
scoheb: or even "team-specific" or "org-specific"

ADR/0039-send-cloud-events.md
```diff
@@ -0,0 +1,34 @@
+# 40. Konflux should send cloud events all system events.
```
arewm, line 1: Can we change this to 41? I am trying to deconflict the ADRs in review

ADR/0039-send-cloud-events.md
ggallen: @lcarva , @rbean is correct.

ADR/0039-send-cloud-events.md
ggallen: Why?  I actually don't agree. This should be a requirement.  If not, you end up with a system that has no consistency. Some tools may emit events while other do not. This does not make a for a consistent, or good, user experience.

The intent of the ADR is to have Konflux emit events for everything that makes sense.

ADR/0039-send-cloud-events.md
ggallen: @arewm, I will make this change!

ADR/0039-send-cloud-events.md
ggallen: I changed to "application-specific", as application seems to be the correct term for Konflux.

ADR/0039-send-cloud-events.md
ggallen: Done. It's now #41, matching the title here.

ADR/0039-send-cloud-events.md
lcarva: In this case, I do not agree with the ADR. It's too early to make such a call. We should first experiment with emitting some cloud events. Then, once the value has been proved, we can make a follow up ADR to actually apply this across the board.

ggallen: > It's too early to make such a call.

To early in what sense?

Tekton already provides cloud events, so they have done this "experiment" and have determined it is worthwhile.

I am working on getting together some use cases.

lcarva: > To early in what sense?

I'm thinking of the Konflux-specif components, e.g. release/integration service. I'd like to see things like that emitting cloud events for some time before requiring everything in Konflux to emit cloud events.

ADR/0039-send-cloud-events.md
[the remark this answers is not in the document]
ggallen: I don't envision the UI being involved here at all. The emission of cloud events just allows users to monitor and react to any event they deem important. It similar to a CLI, and in that case I wouldn't expect the UI to display anything about any client that uses the CLI.

ggallen:
> > Cloud event generation could be optional and that option could default to off.
> 
> Are you talking about turning on that functionality on a specific deployment of Konflux? That is perhaps outside the scope of these ADRs. As I see it, these are about the Konflux project, not necessarily any one deployment of Konflux.

Konflux should allow users to turn on or turn off the emission of cloud events. Turning it on would then turn on emitting cloud events in all Konflux components. The same for turning it off.

In other words, this should be provided as a Konflux configuration option, to be turned on or off in any Konflux deployment.

ADR/0039-send-cloud-events.md
ggallen: @ralphbean, please see the section of use cases that I just added.

ggallen:
> I think this ADR could benefit from the use cases or examples of situations where cloudevents would be helpful to users. What do you envision the users wanting to do?
> 
> I believe this would help with the question I ask myself ... "why not develop whatever feature the users is resorting to using cloudevents for?"

@scoheb, I just added a section of use cases.

scoheb (approved):

PR #208, opened 2024-09-26
ADR 42: Workspace Design Proposal

dperaza4dustbit:
Proposing to create a k8s API extension to implement the concept of a konflux workspace


ADR/0040-worspace-design-proposal.md
```diff
@@ -0,0 +1,172 @@
...
+
+## Consequences
+
+Extending k8s API with the workspace concept will allow konflux user to manage their konflux workspaces in a secured, self-service manner while konflux admin can maintain multi-tenancy.
+
+Such extention of the API that wraps namepsace has precedence, for example OpenShift apis provide a Project kind the wraps a k8s namespace.
+
+### Additional benefit 
+
+We can automate the bootstraping of namespaces with extra resources a konflux admin deems proper for their specific use case.
```
gbenhaim, line 62: This is a good feature that can also achieved by existing projects such as kyverno that knows how to do it for namespaces.

ADR/0040-worspace-design-proposal.md
```diff
@@ -0,0 +1,172 @@
...
+
+We can create exension points that enable the community to use different backends of the implementation which will include the ability to scale to multiple cluster with a single instance of Konflux.
```
gbenhaim, line 64: 1. Do we have any requirement today for multi-cluster extensions or we can wait until we will have evidence that it's needed?
2. When you are saying "we", to who you refer? If I'm a contributor of Konflux, how do I develop an extension? to where I submit the code and who needs to review it? how the workspace abstraction helps me to develop an extension?

ADR/0040-worspace-design-proposal.md
```diff
@@ -0,0 +1,172 @@
...
+
+
+## Konflux Workspaces Upstream Design Proposal
+
+### Purpose
+
+This a design proposal to represent the Konflux workspace concept long term. It aims to show how the Konflux upstream project should implement a workspace abstraction that will allow the community to select their own authentication provider as well as their own namespace provisioning backend. Three namespace provisioning backends will be presented as reference implementation but it should not be limited to those as the community will be able to extend or modify to better suit their use case.
```
gbenhaim, line 71: Can you please provide more details about the authentication provider,how is it related to workspaces?
Konflux upstream is already using [oauth2-proxy](https://oauth2-proxy.github.io/oauth2-proxy/) and [dex](https://dexidp.io/) (keycloak can be used as well) for letting konflux admins configure social login. Upstream Konflux is not limited to using oauth2-proxy and dex, other implementation/services can be used as well.

ADR/0040-worspace-design-proposal.md
```diff
@@ -0,0 +1,172 @@
...
+
+Directly managing a namespace for a user has limitations imposed by Kubernetes.  First, Kubernetes lacks the granularity to allow a user to see only the namespaces it has access to.  If we want to only provide a list of namespace a user has access to, we will need another mechanism.  Second, we also will need to support provisioning of a namespace, which involves the creation of the namespace and populating it with the resources necessary for Konflux including but not limited to RoleBindings and Quotas.  Without any kind of secondary mechanism for this process, a user will need to understand how to do this process manually or delegate to an admin.  Finally, we will want to have support for setting specific limmits and restrictions on users, such as how many workspaces a user can create and own.  Collecting these metrics and determining violations will require some mechanism for enforcing these policies, which Kubernetes does not give us.
```
gbenhaim, line 85: 1. Do you see any reasons why we can't achieve it by using namespace directly + keyverno/argocd, or as mentioned above in openshift using projects?

2. Why do you think it's an issue for Konflux to solve?  for example, when I install (kubevirt)[https://kubevirt.io/], it doesn't create a namespace for me where I can create VMs in. Same goes for Tekton, I doesn't create a namespace for me where I can run piplines. TBH, I'm not aware of any extension that provides this functionality, why Konflux is special?

ADR/0040-worspace-design-proposal.md
```diff
@@ -0,0 +1,172 @@
...
+
+The diagram depicts 3 backend implementations for the Workspace Controller: KubeSaw, Local and GitOps. All backends end up managing namespaces, their initial resources required, and connectivity between these namespaces to authenticated identities.
+
+##### KubeSaw
+
+This is the current implementation downstream running today in our KubeSaw deployments. It leverages the KubeSaw host operator, member operator and registration service components. It is designed to scale with multiple member cluster support. This implementation will only be possible with OpenShift clusters, since there is no vanilla support of K8S at the moment. Implementing support for non-OpenShift Kubernetes deployments is scheduled for the future.
+
+##### Local
+
+This implementation assumes Konflux runs in single cluster mode. It is a straightforward implementation of local namespace provisioning, bootstrapping and user allocation. It will require a workspace config object that describes the default resources to set during the bootstrapping phase.
```
gbenhaim, line 107: Can we find a middle ground and have the local implementation not using the workspace abstraction

ADR/0040-worspace-design-proposal.md
```diff
@@ -0,0 +1,172 @@
...
+
+##### GitOps
+
+This implementation provisions the namespace in a target cluster, but the bootstrapping is delegated to a GitOps provider like ArgoCD. A new application will be spawned on the git repo and ArgoCD will track those resources and synchronize them with the cluster. This is the current approach using upstream at the moment.
```
gbenhaim, line 111: 1. can you please provide details about how this will be implemented? which changes would be needed to https://github.com/konflux-workspaces/workspaces ?

2. As I mentioned above, upstream Konflux isn't opinionated about the way namespace are provisioned with initial set of resources. We don't need to create any resource in a namespace for getting Konflux to operate on it.

ADR/0040-worspace-design-proposal.md
```diff
@@ -0,0 +1,172 @@
...
+
+### Features by backend
+
+The table below expands with more details on the features that each backend can provide and which features will not be available.
+
+| Feature | KubeSaw | Local | GitOps |
+| :---- | :---- | :---- | :---- |
+| Single-cluster | ✅ | ✅ | ⚠️ (ArgoCD) |
+| Multi-cluster | ✅ | ❌ | ✅ |
```
gbenhaim, line 120: what does it mean multicluster ? multicluster namespace provisioning? how traffic will be routed to the different clusters?

ADR/0040-worspace-design-proposal.md
```diff
@@ -0,0 +1,172 @@
...
+| Personal Workspaces | ✅ | ❌ | ❌ |
+| Workspace Idling | ✅ | ❌ | ❌ |
+| Multiple Workspaces | ✅ | ✅ | ✅ |
+| Vanilla K8S | ❌\* | ✅ | ✅ |
+
+\* planned for future
+
+#### Legend
+
+* ✅ implemented in the backend and available
```
gbenhaim, line 138: The local implementation is already available in https://github.com/konflux-workspaces/workspaces ?

ADR/0040-worspace-design-proposal.md
sadlerap: A user's workspace would be assigned to a single cluster, and the set of all workspaces for a konflux installation could be distributed across multiple distinct clusters.  We're not looking to have a workspace living on multiple clusters.

* * *

With the kubesaw backend, we would solve the traffic routing problem with registration-service.

ADR/0040-worspace-design-proposal.md
sadlerap: We've architected konflux-workspaces so that our storage layer is swappable, and it could be configured with a different backend (see `server/persistence`).  Right now, we've only implemented a backend that can talk to kubesaw, but there's no reason we can't implement more there besides developer throughput.

We'd also need to plumb through this configuration during server startup (such as read the backend choice from a configmap and set it up appropriately).  As always, we'd also need to implement tests for this.

ADR/0040-worspace-design-proposal.md
sadlerap: No.  This is just indicating what features each of the proposed backends could support.

ADR/0040-worspace-design-proposal.md
```diff
@@ -0,0 +1,172 @@
...
+
+Kubernetes namespaces where not designed for self-service provisioning in a multi-tenant.
+
+Access for namespace list and create operation is all namepaces or none. This affects multi-tenancy if self-service is desired.
+
+Access to get and delete can be granted via RBAC once the namespace is created. This will required admin access which also affect self-service.
+
+Allowing access to specific Roles or ClusterRoles via RoleBinding can be done by an Admin, also braking self-service. This is required to be able to execute Konflux actions inside the namespace.
+
+## Decision
```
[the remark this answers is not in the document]
dperaza4dustbit: > 1. Can you please provide details about how this will get implemented? It would be nice to understand how it will work for each operation.

Below you will find more details for the design proposal be since you asked I will answer here and we can create a list once everyone ask their questions.
> 2. The `workspace` resource is name spaced or is it a cluster resource?

Namepsace scoped

> 3. What is the schema for the workspace resource?

https://github.com/konflux-workspaces/workspaces/blob/main/doc/book/src/rest-api/crds.md
We though it was not needed to get that deep but if you want to explore that

> 4. Do we need to provide a limit to the amount of namespace a user can create? from where this requirement is coming from? why in your opinion it's an issue that Konflux should solve?

I think it needs to be limited, in any self-service multi-tenant services there are sensible limits. Maybe we don't want to solve but need to provide the ability to solve as admins will likely don't want their Clusters to be either attacked or abused.
  
> 5. Is it introducing just a controller or also a REST service?

Both

ADR/0040-worspace-design-proposal.md
```diff
@@ -0,0 +1,172 @@
...
+Introduce a workspace CRD and corresponding controller as a project of our community.
+
+We will provide all authenticated users access to create their own workspaces within a limit.
+They will also be able to update, list, get and delete only workspaces they own. 
+
+## Consequences
+
+Extending k8s API with the workspace concept will allow konflux user to manage their konflux workspaces in a secured, self-service manner while konflux admin can maintain multi-tenancy.
+
+Such extention of the API that wraps namepsace has precedence, for example OpenShift apis provide a Project kind the wraps a k8s namespace.
```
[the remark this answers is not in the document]
dperaza4dustbit: > 1. If I'm using Openshift would I need the proposed worksapce implementation or I can keep using Openshift projects that solve some of the issues mentioned above?
No, OpenShift is solving a different problem but it is an example of an abstraction to overcome some of the namespace limitation. 
> 2. Do you maybe know if there is a way to use openshift projects on non openshift cluster?
I would not go that way
> 3. Do you have any other examples?
This is one very popular example, not sure we need another.  I was just trying to set precedence, to show is not not a new thing we came up with.

ADR/0040-worspace-design-proposal.md
dperaza4dustbit: I agree, there are multiple solutions to a problem, this is not an exception

ADR/0040-worspace-design-proposal.md
dperaza4dustbit: > 1. Do we have any requirement today for multi-cluster extensions or we can wait until we will have evidence that it's needed?
We do not have to create a backend that support multi-cluster. But the ability to extend the controler with your own backend is what the value is here. You have the ability to keep your user API and scale more as needed and btw contribute your backend back to the community. 
> 2. When you are saying "we", to who you refer? If I'm a contributor of Konflux, how do I develop an extension? to where I submit the code and who needs to review it? how the workspace abstraction helps me to develop an extension?
The audience here is the Konflux community, so "we" is the audience for this ADR

dperaza4dustbit (commented):
Responding to some questions from reviewer

ADR/0040-worspace-design-proposal.md
```diff
@@ -0,0 +1,172 @@
...
+There is not support for multi-cluster for a single instance of Konflux. The Idea here is to require a new deployment of konflux of there is a need to scale beyond a single cluster. Therefore, there is a one to one relationship between konflux instance and k8s cluster.
+
+With this contraint in place workspace concept can be implemented differently
+We relay on the ability to create namespaces directly. Which requires access to all authenticated users to have premissions to create a namespace.
+
+The boostrapping of the namespace created is delegate to a GitOps workflow that will automatically create an application with the necesary resources, including RoleBinding and Quotas and will continue to synch this through the lifecycle of the project.
+
+To be able to see exisiting workspaces for a specific user, we developed a small abstraction called workspace manager, which is able to determine what namepsace have a rolebinding providing access to the current user.
+
+There is currently no way to delete a workspace self-service. User will have to make a request to a konflux admin to delete a specific workspace
```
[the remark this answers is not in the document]
gbenhaim: depends on the cluster-admin and the way that namespaces are being managed on this cluster.

ADR/0040-worspace-design-proposal.md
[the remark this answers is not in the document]
gbenhaim: | https://github.com/konflux-workspaces/workspaces/blob/main/doc/book/src/rest-api/crds.md
We though it was not needed to get that deep but if you want to explore that

how each field in the status section is calculated?

ADR/0040-worspace-design-proposal.md
gbenhaim: so traffic routing isn't something related to the workspaces abstraction? if so why it should be mentioned in this adr?

ADR/0040-worspace-design-proposal.md
```diff
@@ -0,0 +1,172 @@
...
+## Decision
+
+Introduce a workspace CRD and corresponding controller as a project of our community.
+
+We will provide all authenticated users access to create their own workspaces within a limit.
+They will also be able to update, list, get and delete only workspaces they own. 
+
+## Consequences
+
+Extending k8s API with the workspace concept will allow konflux user to manage their konflux workspaces in a secured, self-service manner while konflux admin can maintain multi-tenancy.
```
gbenhaim, line 56: what is the process for giving a user permission on a workspace?

ADR/0040-worspace-design-proposal.md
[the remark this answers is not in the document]
gbenhaim: an important difference when comparing to OCP projects, in OCP you can work with the namespaces directly and the project will get created for you. It doesn't force you to use projects.

ADR/0040-worspace-design-proposal.md
gbenhaim: @sadlerap which interfaces should be implemented in order to add a new provider?

ADR/0040-worspace-design-proposal.md
gbenhaim, line 107: | . It will require a workspace config object that describes the default resources to set during the bootstrapping phase.

what is workspace config? workspace CRD?

ADR/0040-worspace-design-proposal.md
sadlerap: While kubesaw would implement traffic routing for us, the GitOps backend might require a different solution.  I'm not sure if we could get registration-service working standalone in this use case, but if not, this routing problem is something that the workspaces abstraction is going to need to deal with.

Either way, it's something that workspaces will need to care about, even if it does punt a solution to another provider.

ADR/0040-worspace-design-proposal.md
sadlerap: It would either be a configmap or a CRD, and I think a CRD would be preferable so we can add and enforce structure in the config.

ADR/0040-worspace-design-proposal.md
sadlerap: We'd need to implement the interfaces represented in the [`ReadClient`](https://github.com/konflux-workspaces/workspaces/blob/main/server/persistence/readclient/readclient.go#L15-L20) and [`WriteClient`](https://github.com/konflux-workspaces/workspaces/blob/main/server/persistence/writeclient/writeclient.go#L18-L24) structs.  Of particular interest would be the [`DirectAccessChecker`](https://github.com/konflux-workspaces/workspaces/blob/main/server/persistence/clientinterface/readclient.go#L46-L48) interface, as this is how we check if a user has access to a workspace.

On further review, we'd also need to make changes in the `operator/internal/controller` package.  Currently, the `InternalWorkspace` CRD gets converted to a `Space` that manages the namespace controlling that workspace, along with granting users correct permissions via invoking `SpaceBindings`.  If we were to use a separate backend not based on kubesaw, we'd need to swap out these resources for their equivalent semantics.

ADR/0040-worspace-design-proposal.md
dperaza4dustbit: Well something has to take an action to provision namespace, assign the admin role and bootstrap it. The proposal here is to have a workspace controller react to a CR. Yes, that could be other implementations to do the same. I just think this is the simplest solution, suitable to run on kind for example.

ADR/0040-worspace-design-proposal.md
dperaza4dustbit: > 1. Do you see any reasons why we can't achieve it by using namespace directly + keyverno/argocd, or as mentioned above in openshift using projects?

There are many solution as you mentioned that involved using existing controllers that reacts to user input. I just think that this approach provides the simples solution, for the admin and for the end user. End user is particularly benefiting from this as it will only need to learn how to manage a single workspace CR. Documentation will be straight forward as well. 

> 2. Why do you think it's an issue for Konflux to solve?  for example, when I install (kubevirt)[https://kubevirt.io/], it doesn't create a namespace for me where I can create VMs in. Same goes for Tekton, I doesn't create a namespace for me where I can run piplines. TBH, I'm not aware of any extension that provides this functionality, why Konflux is special?

From the beginning the way I think about konflux is technology put together to implement a secured CI/CD service, not as a product. In services the on-boarding and off-boarding of users are a cornerstone piece of the puzzle, providing consistency that can enable other services to integrate with it. One advantage of us providing this solution in the community is that we will benefit on Red Hat deployments from the community ideas, and the community will benefit from our Red Hat reference implementation.

ADR/0040-worspace-design-proposal.md
```diff
@@ -0,0 +1,172 @@
+# 40. Workspace Abstraction
```
lcarva, line 1: I don't agree that this should be within the scope of the Konflux _project_. My concern is that this moves Konflux closer to the cluster management space which is not what Konflux is intended to do. It lowers the flexibility of the project. 

Instead, the Konflux project should be compatible with different mechanisms in which namespaces can be provisioned, including as simple as manually running `kubectl` to create a namespace.

KubeSaw definitely has its benefits and we should make it possible for a Konflux _deployment_ to be compatible with it.

ADR/0040-worspace-design-proposal.md
dperaza4dustbit: From my point of view we will support whatever authentication option k8s support: https://kubernetes.io/docs/reference/access-authn-authz/authentication/ whatever we have in place can work fine or we can think of a different authentication reference implementation as a separate topic. Having an abstraction for workspace does not affect authentication it does simplify authorization by auto generating the role binding for workspace creator.

ADR/0040-worspace-design-proposal.md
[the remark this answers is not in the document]
dperaza4dustbit: > | https://github.com/konflux-workspaces/workspaces/blob/main/doc/book/src/rest-api/crds.md We though it was not needed to get that deep but if you want to explore that
> 
> 1. how each field in the status section is calculated?

Two fields right now owner and space. Owner is the user making the request to create workspace, space is the namepsace that gets generated for them.

> 
> | Namepsace scoped
> 
> 2. how would you implement setting limits to the amount of workspaces a user can create? since all the workspace objects exist in the same namespace, ResourceQuota can't be used (they are tied to a namespace, not a user).

You can have a two prone approach Validating Web Hook that limits the number of workspaces an owner can have. But also the controller reconciler can check and error out. This provide additional protection. Lets say the validating web hook crashes for whatever reason, well the reconciler will catch it. Lets say the reconciler crashes, well it won't be able to create the corresponding namespace, which is what we are trying to prevent in the first place. Kyverno can simplify the validating webhook part, but won't create that addition protection that the reconciler provides

> 
> | Is it introducing just a controller or also a REST service? | Both
> 
> 3. Today in upstream konflux the user can do everything he/she need using `kubectl`. how will it work with the workspaces rest service, can I create workspaces using kubectl?

Yes I would not have it other way, the REST API is fully K8s compatible and user will be able to interact with like they do with any other resource via kubectl or via GitOps if they want

ADR/0040-worspace-design-proposal.md
dperaza4dustbit: Thanks for you opinion on this @lcarva can you expand on why you think this will be cluster management? I would say this will provide multi-tenant, self-service namespace management. Not aiming to provide cluster management at all.

ADR/0040-worspace-design-proposal.md
```diff
@@ -0,0 +1,172 @@
...
+We have a workspace CRD defined with a controller that interfaces with KubeSaw as the NSaaS implementation.
+
+Kubesaw provides multiple features but mainly namespace provisioning, multi-cluster placement algorithms, syncronization of tiers (Quotas, Roles, RoleBindings and ConfigMaps) and user lifecycle management.
+
+Workspace CRD serves as an abstraction to be able to make changes to a workspace or create new workspaces for a specific users. 
+
+This abstraction also enable Konflux to implement other features in the future like for example limmiting the amount of workspaces per user, allowing workspace delete self-service and adding workspace delete protection.
+
+
+### Upstream
```
MatousJobanek, line 24: I would recommend making one step back and looking at the requirements and expectation you have for upstream. What do you want to provide as part of the upstream project?
Is it only a secured build system?
Do you want to also provide namespace management tooling?
How about personas and their ability to manage namespaces and other resources in the cluster? Is it something you want to provide in Konflux as well or not?

My point is, do you want to focus on implementing the secured pipeline only, or do you want to try to solve anything related to multi-tenancy as well? Or would you rather delegate this to the setup that is used by the customer/consumer? 
The environments where Konflux will be used may have tens of ways of managing namespaces and access to them (KubeSaw, ArgoCD, Kyverno, kubectl, RHDH, OpenShift projects, custome in-house implementation, ...). I wouldn't try to limit it to only a small set of specific supported tools if the project should be really adopted in an already existing environments. Because keep in mind that there are running clusters that already have many namespaces created, how should be Konflux integrated in such an environment?
In addition to that, each company may have completely different set of policies for managing access to clusters. There can be small start-ups that doesn't care and everyone has cluster-admin permissions, but there are can be companies where devs can only view stuff, or don't have any access to the cluster at all.

And just to be clear, the before-mentioned text is about requirements for "upstream" only. 
The requirements for the internal running Konflux instance is completely different topic. Red Hat is just one adopter of Konflux project that has its own specific requirements on how the namespaces and access is managed there.

ADR/0040-worspace-design-proposal.md
```diff
@@ -0,0 +1,172 @@
...
+
+There is not support for multi-cluster for a single instance of Konflux. The Idea here is to require a new deployment of konflux of there is a need to scale beyond a single cluster. Therefore, there is a one to one relationship between konflux instance and k8s cluster.
+
+With this contraint in place workspace concept can be implemented differently
+We relay on the ability to create namespaces directly. Which requires access to all authenticated users to have premissions to create a namespace.
+
+The boostrapping of the namespace created is delegate to a GitOps workflow that will automatically create an application with the necesary resources, including RoleBinding and Quotas and will continue to synch this through the lifecycle of the project.
```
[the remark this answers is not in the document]
MatousJobanek: Out of curiosity, do you mention (or recommend) anywhere how the namespace should be provisioned? Or is it really purely on the owner of the cluster how this is done there?
In other words, does it work in the same way as any other operator (eg OpenShift AI) that rely on which namespace they have access to?

ADR/0040-worspace-design-proposal.md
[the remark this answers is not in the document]
MatousJobanek: So if I understood @gbenhaim correctly, Konflux doesn't care who creates or deletes the namespaces. It just works with what is in the cluster and what user has access to, is that correct?

ADR/0040-worspace-design-proposal.md
```diff
@@ -0,0 +1,172 @@
...
+Access to get and delete can be granted via RBAC once the namespace is created. This will required admin access which also affect self-service.
+
+Allowing access to specific Roles or ClusterRoles via RoleBinding can be done by an Admin, also braking self-service. This is required to be able to execute Konflux actions inside the namespace.
+
+## Decision
+
+Introduce a workspace CRD and corresponding controller as a project of our community.
+
+We will provide all authenticated users access to create their own workspaces within a limit.
+They will also be able to update, list, get and delete only workspaces they own.
```
MatousJobanek, line 52: There are platforms where users have only limited namespace-scoped permissions in one or more specific namespaces, or limited access in general (only view for example). Granting permissions to create workspaces (and thus namespaces) to all authenticated users means privilege escalation and opening a back-door.

ADR/0040-worspace-design-proposal.md
MatousJobanek: Is this something you really want to provide as part of the upstream project? Let's not talk about tools (KubeSaw, Kyverno, ArgoCD, ...), but about requirements for the feature set of the upstream project. 

It's definitely needed for the internal RH instance of Konflux, but this is completely different topic which should be discussed separately.

ADR/0040-worspace-design-proposal.md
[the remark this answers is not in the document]
gbenhaim: |  Or is it really purely on the owner of the cluster how this is done there?
In other words, does it work in the same way as any other operator (eg OpenShift AI) that relies on which namespace they have access to and what permissions they have there?

yes exactly.

ADR/0040-worspace-design-proposal.md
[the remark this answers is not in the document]
gbenhaim: @MatousJobanek correct.

ADR/0040-worspace-design-proposal.md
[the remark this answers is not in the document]
gbenhaim: | Yes I would not have it other way, the REST API is fully K8s compatible and user will be able to interact with like they do with any other resource via kubectl or via GitOps if they want

it would mean the user will have to connect to the workspace rest service when he want to work with the workspace api and then switch to the cluster's api when he want to work with other resources?

> You can have a two prone approach Validating Web Hook that limits the number of workspaces an owner can have. But also the controller reconciler can check and error out. This provide additional protection. Lets say the validating web hook crashes for whatever reason, well the reconciler will catch it. Lets say the reconciler crashes, well it won't be able to create the corresponding namespace, which is what we are trying to prevent in the first place. Kyverno can simplify the validating webhook part, but won't create that addition protection that the reconciler provides

1. Is there any reason a validating webhook, such the one described above, can't be created for namespace objects?

2. If the validating webhook crashes it will block the creation of the resources it admits, since the k8s api will try to forward the request to a server that won't respond.

3. built in kubernets quotas are implemented using admission webhooks (https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#resourcequota), so I think it would be safe enough to relay solely on admission webhook for limiting the amount of namespaces a user can create.

ADR/0040-worspace-design-proposal.md
gbenhaim: > Instead, the Konflux project should be compatible with different mechanisms in which namespaces can be provisioned, including as simple as manually running kubectl to create a namespace.

I totally agree with this requirement.

ADR/0040-worspace-design-proposal.md
MatousJobanek: Important thing to understand is that the multi-cluster functionality is important for the user & namespace management. It shouldn't have any impact on the operators or APIs provisioned in the clusters. 

Let's take an example of DevSandbox:
* there are 3 member clusters installed with a long list of operators. None of the operators is multi-cluster aware - they work within the scope of the cluster only
* plus there is one more member cluster (provisioned with some specific configuration) that has only OpenShift AI operator installed and is used only for a specific use-case.

Each installation of the operator in each member cluster doesn't know about each other. And when you you are provisioned in any of the member clusters, then you have no clue about the others and what API is available in the other member clusters.

Having all data located in one cluster makes the admin-work much easier as there is only one source of truth. 

And the proxy (and registration-service) makes the usage of the multi-cluster environment easier for the end-users (or 3rd party services) - there is only one endpoint they need to talk to and there is only one URL the users need to get into to be redirected to the correct member-cluster or to know which clusters it has access to. Or you can configure your internal UI so it talks to the proxy instead of each of the member clusters.

In similar way, you may have one or more clusters installed & configured with Konflux - the number of Konflux installations is equal to the number of member clusters. Konflux doesn't have to be multi-cluster or proxy aware, but the whole environment (user & namespace management) can still benefit from the multi-cluster setup.

ADR/0040-worspace-design-proposal.md
arewm, line 1: Can we change this ADR to 42? I am trying to deconflict the in-review ADRs. 

I think that we should still merge this ADR even if the status is not accepted.

PR #210, opened 2024-10-18
Update the API references

arewm:
Removing deprecate/removed APIs. Also removing the workspace API which was being generated but not linked in the contents.

Adding the image controller references to the contents.


ralphbean (approved):

dirgim (approved):
+1

PR #212, opened 2024-10-23
ADR 43: add the Artifact construction ADR

dirgim:
This ADR aims to cover the proposed functionality for creating additional Artifacts as part of the Konflux build/test/release process.

Signed-off-by: dirgim <kpavic@redhat.com>

### DIFF
--- a/ADR/0043-artifact-constructors.md
+++ b/ADR/0043-artifact-constructors.md
@@ -1,0 +1,104 @@
+# 43. Creating Artifacts from Snapshots for testing and release
+
+Date: 2024-10-21
+
+## Status
+
+Proposed
+
+## Authors
+
+- Krunoslav Pavic
+- Andrew McNamara
+
+## Context
+
+This ADR aims to cover the proposed functionality for creating additional Artifacts as part of the Konflux build/test/release process.
+These artifacts would vary in type and would cover use cases where either the single built image or entire sets of images
+from the user's application are needed to generate one or more artifacts.
+
+Some limitations with our current approach that can be mitigated with this ADR:
+* The ability to produce multiple artifacts with a single PipelineRun
+* Reduced need for in-PR nudging (one of the common use cases for this is the operator bundle builds as mentioned below)
+
+
+## Use Cases in Detail
+
+* A team produces operands and operators. When integration service creates a Snapshot, it uses a constructor that
+  actually builds a bundle. When releasing, the team releases that single bundle artifact. That constructor also has an extract
+  operation to list all related images that are referenced, and that extraction generates the list for release service to operate on
+* A team produces a set of images with ko. The build task generates/pushes all images with ko and then produces a "build artifact"
+  OCI manifest which includes an immutable reference to each of the ko-built artifacts. When integration service creates a Snapshot,
+  it uses a constructor that doesn't need to create a new artifact. When releasing, the team releases that single artifact.
+  The constructor also has an extract operation to list all the ko-built artifacts and that extraction generates the list for release service to operate on.
+* A team produces a set of images that are just a group of artifacts that should be released together.
+  The artifacts are built via separate PipelineRuns. When integration service creates a Snapshot, it uses a constructor
+  that saves the Snapshot CR as an OCI blob. When releasing, the team releases that single artifact.
+  The constructor also has an extract operation to list all artifacts that were in the Snapshot and that extraction
+  generates the list for release service to operate on.
+
+
+## Decision
+
+A new API will be introduced which instructs the Integration service to create a Tekton pipeline which will generate the
+artifact.
+Once built, the artifact reference will be added to the Snapshot's metadata so it can be discovered during testing or release process.
+Additionally, the integration service will wait for the artifact constructor to finish before starting the integration testing pipelines
+in order to enable users to run tests using the new artifacts.
+
+### The ArtifactConstructor API
+
+The proposal here is to introduce the new ArtifactConstructor API which will define Tekton pipelines that will generate individual artifacts.
+These pipelines would be executed after a Snapshot is generated and before the integration test pipelines start.
+
+When each artifact generation pipelineRun is completed, the reference to the artifact is extracted and added to the original Snapshot.
+After that, the integration testing pipelines castart, and since the artifacts are referenced by the Snapshot,
+it is possible to fetch them and run tests on them.
+
+```yaml
+apiVersion: appstudio.redhat.com/v2alpha1
+kind: ArtifactConstructor
+metadata:
+    name: example--artifact-constructor
+spec:
+    selectors:
+        - expression: ‘metadata.labels["konflux-ci.dev/application"] == "integration-service" && metadata.annotations["test.pac.appstudio.openshift.io/event-type"] == "pull_request"’
+          description: “Selects only PR Snapshots associated with integration-service”
+    params:
+      - name: source
+        value: https://github.com/konflux-ci/integration-service
+    resolverRef:
+      resolver: git
+      params:
+        - name: url
+          value: https://github.com/konflux-ci/integration-examples
+        - name: revision
+          value: main
+        - name: pathInRepo
+          value: pipelines/bundle_generator_pipeline.yaml
+    extractors:
+      - ref: 'status.results.filter(result,result.name == "IMAGE_URL")[0].value'
+        name: ‘metadata.labels["appstudio.openshift.io/component"]’
+
+```
+
+Don't take the fields literally. The details will likely change after the ADR has settled. It is
+here for illustration purposes.
+
+The `selector` field will be used to identify resources that should trigger the construction of new Artifacts. This will be done using CEL expressions.
+
+The `resolverRef` resource will be used to define the Tekton pipeline definition which can be used to run the artifact pipelineRun.
+
+The `extractor` resource will use CEL expressions to extract the individual field values from the artifact pipelineRun.
+The individual artifacts are expected to use a reference that can be used to fetch it, like an image pullSpec or similar URL.
+Since the `extractor` resource is a list, it can take multiple extractor entries if the artifact pipeline is expected to provide multiple artifacts.
+Each extracted resource will be added to the Snapshot's metadata.
+
+## Consequences
+
+- [integration-service] now executes additional Tekton pipelineRuns which are expected to generate artifacts which
+  can then be associated with the Snapshot.
+- [integration-service] waits for the artifact pipelineRuns to finish before starting the integration testing pipelines
+  for a given Snapshot.
+- this will have implications for the [release service] as it will further remove coupling around an Application and an Application Snapshot
+- Potential decrease of a reliance on an Application, which will have implications on [build-service] and the Konflux UI


ADR/0041-integration-service-upstreaming.md
```diff
@@ -0,0 +1,95 @@
...
+The proposal here is to introduce the new ArtifactConstructor API which will define Tekton pipelines that will generate individual artifacts.
+These pipelines would be executed after a Snapshot is generated and before the integration test pipelines start.
+
+When each artifact generation pipelineRun is completed, the reference to the artifact is extracted and added to the Snapshot.
+After that, the integration testing pipelines castart, and since the artifacts are referenced by the Snapshot,
+it is possible to fetch them and run tests on them.
+
+```yaml
+apiVersion: appstudio.redhat.com/v2alpha1
+kind: ArtifactConstructor
```
arewm, line 55: This will result in pipelines run in users' namespaces without the use of PAC. We would lose the ability to interact with PipelineRuns as users might normally be used to (i.e. re-triggering them). How could we ensure that we still provide users with that necessary functionality?

arewm (commented):
Thanks for writing this up to start the discussion (and thanks for writing down some of the thoughts that I brought!). I ran out of time to talk things over more, but I think that it is a good start.

ADR/0041-integration-service-upstreaming.md
```diff
@@ -0,0 +1,95 @@
...
+  that saves the Snapshot CR as an OCI blob. When releasing, the team releases that single artifact.
+  The constructor also has an extract operation to list all artifacts that were in the Snapshot and that extraction
+  generates the list for release service to operate on.
+
+
+## Decision
+
+A new API will be introduced which instructs the Integration service to create a Tekton pipeline which will generate the
+artifact.
+Once built, the artifact reference will be added to the Snapshot's metadata so it can be discovered during testing or release process.
```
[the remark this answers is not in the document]
dirgim: For the higher-level purpose of upstreaming and decoupling services, yes, but I think that the Snapshot (or a future replacement API) should still be used by the integration service for the following:
- As a basis for generating the artifact(s) in the artifact pipelineRuns
- For running integration testing pipelines

ADR/0041-integration-service-upstreaming.md
```diff
@@ -0,0 +1,95 @@
...
+Additionally, the integration service will wait for the pipelines to finish before starting the integration testing pipelines
+in order to enable users to run tests using the new artifacts.
```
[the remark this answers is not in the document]
dirgim: Yes, updated to clarify.

ADR/0041-integration-service-upstreaming.md
```diff
@@ -0,0 +1,95 @@
...
+      params:
+        - name: url
+          value: https://github.com/konflux-ci/integration-examples
+        - name: revision
+          value: main
+        - name: pathInRepo
+          value: pipelines/bundle_generator_pipeline.yaml
+    extractors:
+      - ref: 'status.results.filter(result,result.name == "IMAGE_URL")[0].value'
+        name: ‘metadata.labels["appstudio.openshift.io/component"]’
```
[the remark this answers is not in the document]
dirgim: That's a good question. I think we are aiming for a similar paradigm to occur in multiple APIs, especially for the `selector`, but I'm not sure if a common abstraction would be required.

ADR/0041-integration-service-upstreaming.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Decision
+
+A new API will be introduced which instructs the Integration service to create a Tekton pipeline which will generate the
+artifact.
+Once built, the artifact reference will be added to the Snapshot's metadata so it can be discovered during testing or release process.
+Additionally, the integration service will wait for the pipelines to finish before starting the integration testing pipelines
+in order to enable users to run tests using the new artifacts.
+
+### The ArtifactConstructor API
```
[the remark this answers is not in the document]
dirgim: I'm leaning towards it being a generic artifact constructor to allow for the different types of artifacts to be generated. As long as the pipelines can build something that can be referenced and fetched, I think that we should be able to support it this way.

ADR/0041-integration-service-upstreaming.md
```diff
@@ -0,0 +1,95 @@
...
+When each artifact generation pipelineRun is completed, the reference to the artifact is extracted and added to the Snapshot.
+After that, the integration testing pipelines castart, and since the artifacts are referenced by the Snapshot,
+it is possible to fetch them and run tests on them.
+
+```yaml
+apiVersion: appstudio.redhat.com/v2alpha1
+kind: ArtifactConstructor
+metadata:
+    name: example--artifact-constructor
+spec:
```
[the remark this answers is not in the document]
dirgim: Would a `PARAMS` field work here? That way the users could provide their own set of parameters that could also include the source, configuration, etc.

ADR/0041-integration-service-upstreaming.md
dirgim: The integration pipelines are also run without the use of PaC, and there is already a re-triggering functionality available for them, would that be enough to cover the use cases?

ADR/0041-integration-service-upstreaming.md
```diff
@@ -0,0 +1,95 @@
...
+          description: “Selects only PR Snapshots associated with integration-service”
+    resolverRef:
+      resolver: git
+      params:
+        - name: url
+          value: https://github.com/konflux-ci/integration-examples
+        - name: revision
+          value: main
+        - name: pathInRepo
+          value: pipelines/bundle_generator_pipeline.yaml
```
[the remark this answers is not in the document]
dirgim: I may have erred on the side of more generic solution here - I think that we could support both directly building artifacts such as operator bundle images (or any other images), and building OCI artifacts that contain further data/metadata that can be used as described.

ADR/0041-integration-service-upstreaming.md
```diff
@@ -0,0 +1,95 @@
...
+The `selector` field will be used to identify resources that should trigger the construction of new Artifacts. This will be done using CEL expressions.
+
+The `resolverRef` resource will be used to define the Tekton pipeline definition which can be used to run the artifact pipelineRun.
+
+The `extractor` resource will use CEL expressions to extract the individual field values from the artifact pipelineRun.
+The individual artifacts are expected to use a reference that can be used to fetch it, like an image pullSpec or similar URL.
+Since the `extractor` resource is a list, it can take multiple extractor entries if the artifact pipeline is expected to provide multiple artifacts.
+Each extracted resource will be added to the Snapshot's metadata.
+
+## Consequences
```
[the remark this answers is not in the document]
dirgim: Good points, I'll add them to the consequences and think on further implications.

ADR/0041-integration-service-upstreaming.md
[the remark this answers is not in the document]
arewm: Would these CRDs be managed by the same controller (i.e. integration service) or would there be separate controllers? Would this just be some pattern that could be reused within the same codebase?

ADR/0041-integration-service-upstreaming.md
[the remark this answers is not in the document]
arewm: Are there other use cases that you have in mind?

ADR/0041-integration-service-upstreaming.md
arewm, line 58: Another use case that we need to consider is if the artifact-generating repository reference were updated, we would need to be able to test the new images/build process.

I don't know what the best way to do that is. Would it make sense to still add this Component and then designate _some component's repo/configuration_ as the constructor argument? This feels a little wonky to me.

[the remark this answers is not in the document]
arewm: Something like that might work. I started a new conversation thread related to both this one and the one talking about PAC.

ADR/0041-integration-service-upstreaming.md
arewm: It might help in some use cases. I don't know how narrowly scoped those are.

Another use case that we need to consider is if the artifact-generating repository reference were updated, we would need to be able to test the new images/build process.

I don't know what the best way to do that is. Would it make sense to still add this Component and then designate some component's repo/configuration as the constructor argument? This feels a little wonky to me.

ADR/0041-integration-service-upstreaming.md
```diff
@@ -0,0 +1,95 @@
...
+## Context
+
+This ADR aims to cover the proposed functionality for creating additional Artifacts as part of the Konflux build/test/release process.
+These artifacts would vary in type and would cover use cases where either the single built image or entire sets of images
+from the user's application are needed to generate one or more artifacts.
+
+## Use Cases in Detail
+
+* A team produces operands and operators. When integration service creates a Snapshot, it uses a constructor that
+  actually builds a bundle. When releasing, the team releases that single bundle artifact. That constructor also has an extract
```
gbenhaim, line 23: | it uses a constructor that actually builds a bundle

what does it mean? which constructor?

ADR/0041-integration-service-upstreaming.md
```diff
@@ -0,0 +1,95 @@
...
+- Krunoslav Pavic
+- Andrew McNamara
+
+## Context
+
+This ADR aims to cover the proposed functionality for creating additional Artifacts as part of the Konflux build/test/release process.
+These artifacts would vary in type and would cover use cases where either the single built image or entire sets of images
+from the user's application are needed to generate one or more artifacts.
+
+## Use Cases in Detail
```
gbenhaim, line 20: this section describes the current functionality or the one we will seek for?

ADR/0041-integration-service-upstreaming.md
gbenhaim: nvm, I see an explanation about it below https://github.com/konflux-ci/architecture/pull/212/files#diff-d6bbded950ff5bcfe768192ca004473b0d89e927f5f8cd128922a6dd42b5a167R44

ADR/0041-integration-service-upstreaming.md
```diff
@@ -0,0 +1,95 @@
...
+Once built, the artifact reference will be added to the Snapshot's metadata so it can be discovered during testing or release process.
+Additionally, the integration service will wait for the pipelines to finish before starting the integration testing pipelines
+in order to enable users to run tests using the new artifacts.
+
+### The ArtifactConstructor API
+
+The proposal here is to introduce the new ArtifactConstructor API which will define Tekton pipelines that will generate individual artifacts.
+These pipelines would be executed after a Snapshot is generated and before the integration test pipelines start.
+
+When each artifact generation pipelineRun is completed, the reference to the artifact is extracted and added to the Snapshot.
```
gbenhaim, line 49: added to the same snapshot or a new snapshot will get created?

ADR/0041-integration-service-upstreaming.md
gbenhaim, line 23: How this will work with the nudging functionality? today when a new operand is available a pr for updating the CSV of the bundle with the reference to it is created, and it starts the build process for the bundle.

With this proposal, the bundle build will start by the integration service without requiring  a change to the bundle source in git?

ADR/0041-integration-service-upstreaming.md
dirgim: It's added to the same Snapshot, will clarify.

ADR/0041-integration-service-upstreaming.md
dirgim: It describes the use cases that we want to cater  to.

ADR/0041-integration-service-upstreaming.md
dirgim: That's the general idea, the bundle builds would not require nudging unless the user explicitly wanted to have the bundle contents stored/updated in their git repository - and even then the contents could be updated as part of the release process if needed.

arewm: Users would have to migrate to the flow proposed by the ADR. This does not remove the functionality that is present today with explicit operand/operator/bundle builds and nudging.

It does, however, reduce the demand for in-PR nudging as one of the primary drivers for that was to produce bundle images to enable testing changes.

ADR/0041-integration-service-upstreaming.md
```diff
@@ -0,0 +1,95 @@
+# XX. Creating Artifacts from Snapshots for testing and release
```
arewm, line 1: Can we change this to ADR 43? I am trying to deconflict the in-review ADRs.

dirgim: Ack, updated

ADR/0043-artifact-constructors.md
```diff
@@ -0,0 +1,104 @@
...
+A new API will be introduced which instructs the Integration service to create a Tekton pipeline which will generate the
+artifact.
+Once built, the artifact reference will be added to the Snapshot's metadata so it can be discovered during testing or release process.
+Additionally, the integration service will wait for the artifact constructor to finish before starting the integration testing pipelines
+in order to enable users to run tests using the new artifacts.
+
+### The ArtifactConstructor API
+
+The proposal here is to introduce the new ArtifactConstructor API which will define Tekton pipelines that will generate individual artifacts.
+These pipelines would be executed after a Snapshot is generated and before the integration test pipelines start.
```
mmorhun, line 52: I'm again totally confused. Above you say `Once built, the artifact reference will be added to the Snapshot's metadata` and now you say `API which will define Tekton pipelines that will generate individual artifacts. These pipelines would be executed after a Snapshot is generated`.

ADR/0043-artifact-constructors.md
```diff
@@ -0,0 +1,104 @@
...
+
+When each artifact generation pipelineRun is completed, the reference to the artifact is extracted and added to the original Snapshot.
```
mmorhun, line 54: By `the artifact` you mean "artifact of artifacts"?

ADR/0043-artifact-constructors.md
```diff
@@ -0,0 +1,104 @@
...
+        name: ‘metadata.labels["appstudio.openshift.io/component"]’
+
+```
+
+Don't take the fields literally. The details will likely change after the ADR has settled. It is
+here for illustration purposes.
+
+The `selector` field will be used to identify resources that should trigger the construction of new Artifacts. This will be done using CEL expressions.
+
+The `resolverRef` resource will be used to define the Tekton pipeline definition which can be used to run the artifact pipelineRun.
```
mmorhun, line 90: Do you mean PipelineRun that creates an artifact?

ADR/0043-artifact-constructors.md
```diff
@@ -0,0 +1,104 @@
...
+
+The `extractor` resource will use CEL expressions to extract the individual field values from the artifact pipelineRun.
```
mmorhun, line 92: Shouldn't the pipeline just define result(s) with the reference to the produced by the pipelinerun  artifact?

ADR/0043-artifact-constructors.md
```diff
@@ -0,0 +1,104 @@
...
+The individual artifacts are expected to use a reference that can be used to fetch it, like an image pullSpec or similar URL.
+Since the `extractor` resource is a list, it can take multiple extractor entries if the artifact pipeline is expected to provide multiple artifacts.
+Each extracted resource will be added to the Snapshot's metadata.
+
+## Consequences
+
+- [integration-service] now executes additional Tekton pipelineRuns which are expected to generate artifacts which
```
mmorhun, line 99: I think that individual artifacts (e.g. image or bundle or ...) should be generated by the build pipeline. It does make sense to group individual (let's call "atomic") artifacts produced by build pipeline into a group and the release it as a single unit. But the group of atomic artifacts create an artifact to release which seen somewhat the same as snapshot.

ADR/0043-artifact-constructors.md
```diff
@@ -0,0 +1,104 @@
...
+  can then be associated with the Snapshot.
+- [integration-service] waits for the artifact pipelineRuns to finish before starting the integration testing pipelines
+  for a given Snapshot.
+- this will have implications for the [release service] as it will further remove coupling around an Application and an Application Snapshot
+- Potential decrease of a reliance on an Application, which will have implications on [build-service] and the Konflux UI
```
mmorhun, line 104: Build Service does not rely on Application, main resource is a Component. From Build Service point of view, we can delete Application CR even today.

mmorhun (commented):
The ADR looks interesting, but some terminology is unclear.

ADR/0043-artifact-constructors.md
dirgim: Good point, I think that a better term would be the `Application/Component model`.

ADR/0043-artifact-constructors.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Authors
+
+- Krunoslav Pavic
+- Andrew McNamara
+
+## Context
+
+This ADR aims to cover the proposed functionality for creating additional Artifacts as part of the Konflux build/test/release process.
+These artifacts would vary in type and would cover use cases where either the single built image or entire sets of images
```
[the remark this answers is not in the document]
dirgim: 1. Yes, it can be rpms or binaries or any other kind of artifact that can be stored somewhere and released.
2. Exactly, an operator bundle is a good example, or an fbc fragment - these rely on information (mainly images) from multiple components in order to be generated.

ADR/0043-artifact-constructors.md
```diff
@@ -0,0 +1,104 @@
...
+from the user's application are needed to generate one or more artifacts.
```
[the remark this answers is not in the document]
dirgim: Yes, I think releasing more than artifact at the same time should be supported. We are not delving too deep in the Release process here, but I would like to hear from @davidmogar or @johnbieren about their opinions on this.

ADR/0043-artifact-constructors.md
```diff
@@ -0,0 +1,104 @@
...
+
+Some limitations with our current approach that can be mitigated with this ADR:
+* The ability to produce multiple artifacts with a single PipelineRun
+* Reduced need for in-PR nudging (one of the common use cases for this is the operator bundle builds as mentioned below)
```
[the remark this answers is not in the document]
dirgim: I think that nudging for operator bundle builds could potentially be replaced completely, but I don't think that nudging between images in general  (e.g. base image -> images that are based on it) would be replaced.

ADR/0043-artifact-constructors.md
```diff
@@ -0,0 +1,104 @@
...
+
+
+## Use Cases in Detail
+
+* A team produces operands and operators. When integration service creates a Snapshot, it uses a constructor that
+  actually builds a bundle. When releasing, the team releases that single bundle artifact. That constructor also has an extract
+  operation to list all related images that are referenced, and that extraction generates the list for release service to operate on
```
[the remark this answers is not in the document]
dirgim: Yes, I believe that it would be possible to set up the release logic to release the bundle alongside the related images.

ADR/0043-artifact-constructors.md
```diff
@@ -0,0 +1,104 @@
...
+* A team produces a set of images with ko. The build task generates/pushes all images with ko and then produces a "build artifact"
```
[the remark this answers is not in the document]
dirgim: Good point, maybe calling it an `artifact generation task` would be more clear.

ADR/0043-artifact-constructors.md
```diff
@@ -0,0 +1,104 @@
...
+from the user's application are needed to generate one or more artifacts.
+
+Some limitations with our current approach that can be mitigated with this ADR:
+* The ability to produce multiple artifacts with a single PipelineRun
+* Reduced need for in-PR nudging (one of the common use cases for this is the operator bundle builds as mentioned below)
+
+
+## Use Cases in Detail
+
+* A team produces operands and operators. When integration service creates a Snapshot, it uses a constructor that
```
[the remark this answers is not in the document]
dirgim: It is the ArtifactConstructor CR that's described in the lower text. Maybe these use cases should go below that section and then reference the CR directly that way?

ADR/0043-artifact-constructors.md
```diff
@@ -0,0 +1,104 @@
...
+  The artifacts are built via separate PipelineRuns. When integration service creates a Snapshot, it uses a constructor
+  that saves the Snapshot CR as an OCI blob. When releasing, the team releases that single artifact.
+  The constructor also has an extract operation to list all artifacts that were in the Snapshot and that extraction
+  generates the list for release service to operate on.
+
+
+## Decision
+
+A new API will be introduced which instructs the Integration service to create a Tekton pipeline which will generate the
+artifact.
```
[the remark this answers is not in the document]
dirgim: Good idea, will add that to the ADR.

ADR/0043-artifact-constructors.md
```diff
@@ -0,0 +1,104 @@
...
+Once built, the artifact reference will be added to the Snapshot's metadata so it can be discovered during testing or release process.
```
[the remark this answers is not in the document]
dirgim: The information within the Snapshot (list of images, git references etc.) are used to construct these further artifacts. A single artifact also doesn't need to use all components from the Snapshot, so I think it's valuable to retain the Snapshot as the single point of truth about the whole state of the user's list of components (or Application in current terms).

ADR/0043-artifact-constructors.md
dirgim: Yeah, we may need a diagram here for the proposed workflow:

1. The Snapshot gets created (in response to a build pipeline or manually etc.)
2. The Artifact generation pipelineRuns are started and generate artifacts based on the information in the Snapshot
3. References to the generated artifacts will be added to the Snapshot's metadata (annotations) to provide a central place where we can find all generated artifacts and testing information.

ADR/0043-artifact-constructors.md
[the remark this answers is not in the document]
johnbieren: Not sure I am really following, but isn't this what we already do? We let you say where to release each container image

PR #213, opened 2024-10-24
Added ADR describing support of SDPX SBOM format in konflux

midnightercz:


ADR/0039-spdx-support.md
```diff
@@ -0,0 +1,144 @@
...
+
+SPDX SBOM format enables additional features not available in cyclondedx like multiple purl attributes per component. SPDX is also a widely adopted standard for software bill of materials.
+This ADR describes how to enable use of SPDX SBOM format in Konflux.
+
+## Decision
+
+
+### SBOM lifecycle in build pipeline
+
+At the start SBOMs are generated by cachi2 and syft. At later phase of the build pipeline, SBOMs from base images are extracted and merged toghether with SBOMs generated for the currently build container. When switching to spdx, newly built container image can be used as base image another container image. Therefore any tooling/task which works with sboms has to be able to work with both formats.
```
scoheb, line 19: > When switching to spdx, newly built container image can be used as base image another container image

This needs to be reworded

ADR/0039-spdx-support.md
```diff
@@ -0,0 +1,144 @@
...
+Due to that fact, all tools and tasks should implement sbomType attribute which will be used to determine sbom format expected on the input and produced on the output. It will also allow tools to be tested for spdx before whole ppile is switched to spdx.
```
scoheb, line 20:
```suggestion
Due to that fact, all tools and tasks should implement the sbomType attribute which will be used to determine the sbom format expected on the input and produced on the output. It will also allow tools to be tested for spdx before whole pipeline is switched to spdx.
```

ADR/0039-spdx-support.md
```diff
@@ -0,0 +1,144 @@
...
+| component.purl = <PURL>      | package.externalRefs = [{referenceCategory:”PACKAGE-MANAGER”, |
+|                              |                          referenceType:purl,                  |
+|                              |                         referenceLocator: <PURL>              |
+|                              |                          }]                                   |
+|------------------------------|---------------------------------------------------------------|
+
+
+#### Component.properties
+CycloneDX components properties describe mapping of string:string properties for given component. SPDX component doesn’t have anything similar to cyclonedx properties. SPDX Package annotations are the only attribute where custom data can be stored and the only “customizable” field there is comment which is simple string. Due to that fact, cycloneDX property in format of {“name”: <string>, “value”: <string>} is encoded into json string. There can be also annotations produced by other tools. Therefore to be able to tell annotation comment is json encoded, annotator should ends with string “:jsonencoded”
```
scoheb, line 96:
```suggestion
CycloneDX components properties describe mapping of string:string properties for given component. SPDX component doesn’t have anything similar to cyclonedx properties. SPDX Package annotations are the only attribute where custom data can be stored and the only “customizable” field where there is comment which is a simple string. Due to that fact, cycloneDX property in format of {“name”: <string>, “value”: <string>} is encoded into json string. There can be also annotations produced by other tools. Therefore to be able to tell annotation comment is json encoded, annotator should ends with string “:jsonencoded”
```

ADR/0039-spdx-support.md
```diff
@@ -0,0 +1,144 @@
...
+| Formulations.components = [{}]  | Relationships = [{                                         |
+|                                 |  spdxElementId = <A-PACKAGE-ID>                            |
+|                                 |  relationshipType=BUILD_TOOL_OF                            |
+|                                 |  relatedSpdxElement=<ROOT-DOCUMENT-ID>                     |
+|                                 | }]                                                         |
+|---------------------------------|------------------------------------------------------------|
+
+
+#### Metadata.tools
+For CycloneDX tools sub attributes, we are mostly interested in, are vendor and name elements. Information about creation of SPDX document can be stored into creationInfo. CreationInfo.creators element is basically a list of strings. There’s vague specification about how it should be structured in the standard. Strings should be formatted in the following way: “<Attribute>: <Value>”. For example vendor should be stored as “Vendor: <vendor>”
```
scoheb, line 118:
```suggestion
The CycloneDX tools sub attributes that we are mostly interested in are the vendor and name elements. Information about the creation of the SPDX document can be stored into creationInfo. CreationInfo.creators element is basically a list of strings. There’s a vague specification about how it should be structured in the standard. Strings should be formatted in the following way: “<Attribute>: <Value>”. For example, vendor should be stored as “Vendor: <vendor>”
```

ADR/0039-spdx-support.md
```diff
@@ -0,0 +1,144 @@
...
+
+|------------------------------------------------|--------------------------------------------------|
+| Metadata.tools = [{“vendor”: “X”, “name”: “Y”] |CreationInfo.creators = [“Vendor: X”, “Tool: Y”]  |
+|------------------------------------------------|--------------------------------------------------|
+
+
+#### Merging SPDX
+##### Packages
+Packages of two SPDX documents can be merged together as a concatenation of two lists. In cycloneDX component elements can have only a single purl attribute, therefore component elements representing packages with the same name and version but with different purl have to be stored as multiple elements. CycloneDX package elements can bear multiple purls. Therefore multiple cycloneDX components can be squashed together into single SPDX package element with purls concatenated into a single list. Following rules are applied to packages merging process:
+Packages with the same package name and versionInfo are squashed into single package element
```
scoheb, line 128: bullet point?

ADR/0039-spdx-support.md
```diff
@@ -0,0 +1,144 @@
...
+Package with the same package name and versionInfo set to None (or empty) is squashed with package with the same name and non-empty versionInfo
```
scoheb, line 129: bullet point?

ADR/0039-spdx-support.md
```diff
@@ -0,0 +1,144 @@
...
+NOTE: packages cannot be merged together based on SPDXID attribute as there’s no specification in spdx standard how SPDXID should be calculated. Individual tools can calculate it differently while still passing condition to make it unique across the whole document.
```
scoheb, line 130:
```suggestion
NOTE: packages cannot be merged together based on SPDXID attribute as there’s no specification in the spdx standard on how SPDXID should be calculated. Individual tools can calculate it differently while still passing condition to make it unique across the whole document.
```

ADR/0039-spdx-support.md
```diff
@@ -0,0 +1,144 @@
...
+##### Relationships
+SPDX relationships represent graph/tree structure of relations of elements in the document. Root element is SPDX document itself. Individual packages are in relationship CONTAINS with the root document. In the case of packages which were used to build the container. Packages are in relationship BUILD_TOOL_OF with the root document.
```
scoheb, line 132:
```suggestion
SPDX relationships represent graph/tree structure of relations of elements in the document. The Root element is the SPDX document itself. Individual packages are in relationship CONTAINS with the root document. In the case of packages which were used to build the container, packages are in relationship BUILD_TOOL_OF with the root document.
```

ADR/0039-spdx-support.md
```diff
@@ -0,0 +1,144 @@
...
+NOTE: Packages can be wrapped in a “virtual package”. This package can have empty name and version and on attributes, or name can be set to directory or container name which was set as source for generating SBOM document. This element has insignificant information value and can be omitted.
```
scoheb, line 133: > and on attributes

should this be "and no attributes"?

midnightercz: yes, typo

ADR/0039-spdx-support.md
```diff
@@ -0,0 +1,146 @@
...
+#### Formulations
+CycloneDX formulations describe how the container was manufactured. In SPDX, Relationship elements can be used for the same purpose. All elements in SPDX have SPDXID attribute which is an element identifier unique in the whole SBOM document. Relationship element describes relation between two elements using their SPDXID and relationship type. Relationship type BUILD_TOOL_OF can be used to express the relationship of packages which were used to build the container.
+
+| SPDX Attribute                  | CycloneDX Attribute                                        |
+|---------------------------------|------------------------------------------------------------|
+| Formulations.components = [{}]  | Relationships = [{                                         |
+|                                 |  spdxElementId = <A-PACKAGE-ID>                            |
+|                                 |  relationshipType=BUILD_TOOL_OF                            |
+|                                 |  relatedSpdxElement=<ROOT-DOCUMENT-ID>                     |
+|                                 | }]                                                         |
```
chmeliik, line 116: `<A-PACKAGE-ID> BUILD_TOOL_OF <ROOT-DOCUMENT-ID>` has weird implications. The package is not a build tool for the SPDX document itself (unless the package is cachi2 or syft). It's a build tool for the container we're building.

But I think our SBOMs currently don't include a Component for the container, so this is probably the best we can do for now

ADR/0039-spdx-support.md
```diff
@@ -0,0 +1,146 @@
...
+
+
+#### Metadata.tools
+The CycloneDX metadata.tools sub attributes that we are mostly interested in are the vendor and name elements. Information about the creation of the SPDX document can be stored into creationInfo. CreationInfo.creators element is basically a list of strings. There’s a vague specification about how it should be structured in the standard. Strings should be formatted in the following way: “<Attribute>: <Value>”. For example vendor should be stored as “Vendor: <vendor>”
+
+| SPDX Attribute                                 | CycloneDX Attribute                               |
+|------------------------------------------------|---------------------------------------------------|
+| Metadata.tools = [{“vendor”: “X”, “name”: “Y”] | CreationInfo.creators = [“Vendor: X”, “Tool: Y”]  |
```
chmeliik, line 124: > There’s a vague specification about how it should be structured in the standard. Strings should be formatted in the following way: “<Attribute>: <Value>”

Do you have a reference for this?

> CreationInfo.creators = [“Vendor: X”, “Tool: Y”]

So the Vendor is not associated with the Tool? They're just two elements in the creators list?

ADR/0039-spdx-support.md
```diff
@@ -0,0 +1,146 @@
...
+
+
+#### Merging SPDX
+##### Packages
+Packages of two SPDX documents can be merged together as a concatenation of two lists. In cycloneDX component elements can have only a single purl attribute, therefore component elements representing packages with the same name and version but with different purl have to be stored as multiple elements. CycloneDX package elements can bear multiple purls. Therefore multiple cycloneDX components can be squashed together into single SPDX package element with purls concatenated into a single list. Following rules are applied to packages merging process:
+- Packages with the same package name and versionInfo are squashed into single package element
+- Package with the same package name and versionInfo set to None (or empty) is squashed with package with the same name and non-empty versionInfo
```
chmeliik, line 131: I think merging components as described here is not correct. Having the same name and version is not sufficient indication that two components are the same component.

Consider two "same" components except one is an npm package and the other a python package
Or two identical-looking python packages, except one is from PyPI and the other is a random project with the same name and version hosted only on github.

IMO it's almost never correct to merge two components if their purls differ

ADR/0039-spdx-support.md
```diff
@@ -0,0 +1,146 @@
...
+NOTE: packages cannot be merged together based on SPDXID attribute as there’s no specification in the spdx standard on how SPDXID should be calculated. Individual tools can calculate it differently while still passing condition to make it unique across the whole document.
+##### Relationships
+SPDX relationships represent graph/tree structure of relations of elements in the document. The Root element is the SPDX document itself. Individual packages are in relationship CONTAINS with the root document. In the case of packages which were used to build the container, packages are in relationship BUILD_TOOL_OF with the root document.
```
chmeliik, line 134: > Individual packages are in relationship CONTAINS with the root document.

Why is that important?

ADR/0039-spdx-support.md
```diff
@@ -0,0 +1,146 @@
...
+NOTE: Packages can be wrapped in a “virtual package”. This package can have empty name and version and no attributes, or name can be set to directory or container name which was set as source for generating SBOM document. This element has insignificant information value and can be omitted.
```
chmeliik, line 135: That's syft-specific, right?

ADR/0039-spdx-support.md
```diff
@@ -0,0 +1,146 @@
...
+
+
+## Consequences
+All tooling used in pipeline needs to support SPDX SBOM format
+
+
+## References
+CycloneDX specification https://cyclonedx.org/specification/overview/o
+SPDX specification https://spdx.github.io/spdx-spec/v2.3/
+SPDX json schema https://github.com/spdx/spdx-spec/blob/development/v2.3/schemas/spdx-schema.json#L724
```
xsuchy, line 146: Nit pick.
The current version of SPDX spec is 
https://spdx.github.io/spdx-spec/v3.0.1/
https://spdx.github.io/spdx-spec/v3.0.1/serializations/

midnightercz: yes, but current version syft is using is 2.3

ADR/0039-spdx-support.md
midnightercz: Can you put here real world example? I'm only thinking of lets say "RPM" which can be both rpm package and python package. But eventually they are referring to the same component. For package which is installed github and pip, if it's same package it still does make sense to put it in one component. Still, we're merging sboms of one container produced by different tools or sboms of base image and the container.
Real world example of that situation is there's lets say python package X reported with no version by one tool and with some version reported by another tool. Note that even the name doesn't necessarily have to be same as one tool can report PyYAML and another pyyaml for example. I think the goal here is to produce sboms which doesn't have 4 different pyaml SPDX packages reported.
Note that it would help if cachi2 could generate components/packages with identifier marking the package as python, rpm, npm and so on. That would help in the merging process.
Merging just based on purl is nonsense as you in will work with released images having SBOM packages with multiple purls with with release-specific attributes like "&repo=xxx". So no purl in build phase will match with those.

ADR/0039-spdx-support.md
midnightercz: I don't understand your question. It describes structure of the document so I guess that's why it's important.

ADR/0039-spdx-support.md
midnightercz: I haven't reviewed all spdx sbom generators in the world, so I can't really answer this question.

ADR/0039-spdx-support.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Decision
+
+
+### SBOM lifecycle in build pipeline
+
+At the start SBOMs are generated by cachi2 and syft. At later phase of the build pipeline, SBOMs from base images are extracted and merged toghether with SBOMs generated for the currently build container. When switching to spdx, newly built container image can be used as base image another container image. Therefore any tooling/task which works with sboms has to be able to work with both formats.
+As a result, all tools and tasks should implement the sbomType attribute to specify the expected SBOM format for input and output. This will also allow tools to be tested with SPDX before the entire pipeline transitions to this format.
+
+Initially, SBOMs are generated by Cachi2 and Syft. In a later phase of the build pipeline, SBOMs from base images are extracted and merged with those generated for the container currently being built. When transitioning to SPDX, the newly built container image can serve as a base image for another container. Therefore, any tools or tasks that work with SBOMs must be compatible with both formats.
```
twaugh, line 22: (This repeats line 19)

ADR/0039-spdx-support.md
```diff
@@ -0,0 +1,146 @@
...
+        - List<Component>
+
+SPDX (2.3) is structured document in json format with following structure(not full specification):
+- Document Root
+    - name
+    - SPDXID
+    - creationInfo
+        - Creators
+            - List<String>
+    - components
```
twaugh, line 54:
```suggestion
    - packages
```

ADR/0039-spdx-support.md
```diff
@@ -0,0 +1,146 @@
...
+|                              |                         referenceLocator: <PURL>              |
+|                              |                          }]                                   |
+
+
+#### Component.properties
+CycloneDX components properties describe mapping of string:string properties for given component. SPDX component doesn’t have anything similar to cyclonedx properties. SPDX Package annotations are the only attribute where custom data can be stored and the only “customizable” field where there is comment which is a simple string. Due to that fact, cycloneDX property in format of {“name”: <string>, “value”: <string>} is encoded into json string. There can be also annotations produced by other tools. Therefore to be able to tell annotation comment is json encoded, annotator should ends with string “:jsonencoded”
+
+| SPDX Attribute                  | CycloneDX Attribute                                        |
+|---------------------------------|------------------------------------------------------------|
+| package.properties = [          | component.annotations = [                                  |
```
twaugh, line 102: Is this the right way round?

ADR/0039-spdx-support.md
twaugh: https://spdx.github.io/spdx-spec/v2.3/document-creation-information/#68-creator-field

ADR/0039-spdx-support.md
```diff
@@ -0,0 +1,146 @@
...
+
+| SPDX Attribute                                 | CycloneDX Attribute                               |
+|------------------------------------------------|---------------------------------------------------|
+| Metadata.tools = [{“vendor”: “X”, “name”: “Y”] | CreationInfo.creators = [“Vendor: X”, “Tool: Y”]  |
+
+
+#### Merging SPDX
+##### Packages
+Packages of two SPDX documents can be merged together as a concatenation of two lists. In cycloneDX component elements can have only a single purl attribute, therefore component elements representing packages with the same name and version but with different purl have to be stored as multiple elements. CycloneDX package elements can bear multiple purls. Therefore multiple cycloneDX components can be squashed together into single SPDX package element with purls concatenated into a single list. Following rules are applied to packages merging process:
```
twaugh, line 129:
```suggestion
Packages of two SPDX documents can be merged together as a concatenation of two lists. In cycloneDX component elements can have only a single purl attribute, therefore component elements representing packages with the same name and version but with different purl have to be stored as multiple elements. SPDX package elements can bear multiple purls. Therefore multiple cycloneDX components can be squashed together into single SPDX package element with purls concatenated into a single list. Following rules are applied to packages merging process:
```

ADR/0039-spdx-support.md
twaugh: I think it might be useful to show an example here. In SPDX, you might easily have something like this:
```
"packages": [
  {
    packageA,
    "externalRefs": [
      {
        "referenceCategory": "PACKAGE-MANAGER",
        "referenceType": "purl",
        "referenceLocator": purlA
      },
      {
        "referenceCategory": "PACKAGE-MANAGER",
        "referenceType": "purl",
        "referenceLocator": purlB
      }
    ]
  },
  ...
],
```

That is, a single package has multiple purls. Think of a container image which gets shipped to multiple repositories, such as ubi9. Or, really, any container image which bears multiple tags. Here's a hand-crafted example: https://github.com/RedHatProductSecurity/security-data-guidelines/blob/main/sbom/examples/container_image/release/ubi9-micro-container-9.4-6.1716471860.spdx.json

What this section is saying is that, to express this in CycloneDX 1.4, the only feasible option is to encode it as multiple components, identical except for their purl. (FWIW CycloneDX 1.6 provides the additional option of providing multiple forms of identity evidence: https://cyclonedx.org/docs/1.6/json/#components_items_evidence_identity_oneOf_i0_items_confidence)

So the equivalent CycloneDX 1.4 would need to look like this:
```
"components": [
  {
    "bom-ref": packageA.bom_ref
    packageA...
    "purl": purlA
  },
  {
    "bom-ref": packageA.bom_ref + "-2",
    packageA...
    "purl": purlB
  },
  ...
]
```

So, the trick is to recognise when these multiple CycloneDX components should be merged into a single SPDX package. One way to do this which might be less controversial is only to merge when the checksums are identical?

But I wonder -- would it cause a problem if they were encoded as multiple SPDX packages?

midnightercz: Cachi2 doesn't populate checksum field for sboms. For further processing I think multiple spdx package could  present additional difficulty. Like if you want to check if container contains specific package with a vulnerability you would have to iterate over several package. Also without any reduction of packages having containers like A->B->C (A is base image of B and so on) could lead to at least 6 spdx packages representing one real package (1 package cachi2 + 1 syft x number of purl variations with repo,arch,etc.)

ADR/0039-spdx-support.md
midnightercz: It has been added, just the snipped in this conversation is not updated

ADR/0039-spdx-support.md
chmeliik: > Note that it would help if cachi2 could generate components/packages with identifier marking the package as python, rpm, npm and so on.

The type is part of the purl: `pkg:{type}/{name}@{version}`

Cachi2 merges by the full purl, because in cachi2 different purls = different components (or at least components where you can't be certain they're the same). But for merging in other contexts, like release SBOMs, using only a subset of the purl attributes would be reasonable.

Is your goal here to come up with a generic merging solution? It may be more feasible to tailor it to specific use cases. Merging SBOMs in general and doing it well is, in my opinion, almost impossible

ADR/0039-spdx-support.md
chmeliik: CycloneDX 1.4 probably can't, but we use 1.5

CycloneDX 1.5:
```jsonc
{
  "metadata": {
    "component": {
      // main component (e.g. the output container image)
    }
  },
  "components": [
    // the dependencies included in the container image
  ],
  "formulation": [
    {
      "components": [
        // build-time-only dependencies of the container image
      ]
    }
  ]
}
```

Conversion to SPDX, in my mind:

```
SPDXRef-DOCUMENT DESCRIBES .metadata.component
.metadata.component CONTAINS .components[]
.formulation[].components[] BUILD-TOOL-OF .metadata.component
```

chmeliik: In the current state, Konflux SBOMs don't have a useful .metadata.component. Which IMO shouldn't stand in the way of this ADR, we should document the desired state and file a story to make it so

ADR/0039-spdx-support.md
chmeliik: Ack, @midnightercz could you call that out? Perhaps restructure the doc to separate out the handling of Syft specifically?

> I haven't reviewed all spdx sbom generators in the world, so I can't really answer this question.

Most of your assumptions are based on Syft and that's fine, let's just try to make it clear

chmeliik: Or rather, let's separate "this is what Syft does" and "this is how our SBOMs should be", because what Syft does is often not a good example to follow

ADR/0039-spdx-support.md
midnightercz: So can we agree on merging based on type name and version of purl?

chmeliik: For merging cachi2 SBOMs with other cachi2 SBOMs - no, it should continue merging by the full purl.

For merging cachi2 SBOMs with Syft SBOMs, yes that sounds reasonable. Although I believe the current merge script has some more specialized syft-specific logic already.

For generic merging, yes that sounds reasonable

ADR/0039-spdx-support.md
midnightercz: so you suggest to create another virtual package to describe build-tool-of components? Btw, if you want to have that detailed level of relationships then I think right relationship would be:
SPDXRef-DOCUMENT  -> DESCRIBES:  SPDXRef-Image
SPDXRef-Image -> CONTAINS: PackageA
SPDXRef-DOCUMENT  -> DESCRIBES: SPDXRef-BaseImages
SPDXRef-BaseImages -> CONTAINS:  SPDXRef-BaseImage-Package-1
SPDXRef-BaseImages -> BUILD_TOOL_OF:  SPDXRef-Image

but that's quite complicated relations and in the end only relavant thing is

SPDXRef-BaseImage-Package-1 -> BUILD_TOOL_OF: SPDXRef-Image

And then there's no point of creating SPDXRef-Image as it would be basically equivalent of the virtual element which syft creates already

chmeliik: > And then there's no point of creating SPDXRef-Image

There very much is! But it should not be a virtual element, it should be the output image. In fact I believe having a proper SPDXRef-Image is critical for another feature that @allda and team are working on https://github.com/konflux-ci/build-tasks-dockerfiles/pull/165#issuecomment-2416479615

chmeliik: > SPDXRef-DOCUMENT -> DESCRIBES: SPDXRef-Image
> SPDXRef-Image -> CONTAINS: PackageA
> SPDXRef-BaseImages -> BUILD_TOOL_OF: SPDXRef-Image

These make sense to be. The other BaseImages relations not so much - the SBOM doesn't describe the base images, those have their own SBOMs

ADR/0039-spdx-support.md
```diff
@@ -0,0 +1,151 @@
...
+
+| CyloneDX Attribute                             | SPDX Attribute                                    |
+|------------------------------------------------|---------------------------------------------------|
+| Metadata.tools = [{“vendor”: “X”, “name”: “Y”] | CreationInfo.creators = [“Vendor: X”, “Tool: Y”]  |
+
+
+#### Merging SPDX
+##### Packages
+Packages of two SPDX documents can be merged together as a concatenation of two lists. In cycloneDX component elements can have only a single purl attribute, therefore component elements representing packages with the same name and version but with different purl have to be stored as multiple elements. SPDX package elements can bear multiple purls. Therefore multiple cycloneDX components can be squashed together into single SPDX package element with purls concatenated into a single list. Following rules are applied to packages merging process:
```
a-ovchinnikov, line 127: Asking to educate myself: what does it mean to have packages with identical names and versions, but different purls?

ADR/0039-spdx-support.md
```diff
@@ -0,0 +1,151 @@
...
+- Packages with the same purl package name and version and type are squashed into single package element
+NOTE: packages cannot be merged together based on SPDXID attribute as there’s no specification in the spdx standard on how SPDXID should be calculated. Individual tools can calculate it differently while still passing condition to make it unique across the whole document.
+
+##### Relationships
+SPDX relationships represent graph/tree structure of relations of elements in the document. The Root element is the SPDX document itself (with SPDXID SPDXRef-Document). Individual packages are in relationship CONTAINS with the root document. In the case of packages which were used to build the container, packages are in relationship BUILD_TOOL_OF with the root document.
```
a-ovchinnikov, line 132: > packages are in relationship ...

Does this mean that a SBOM describing a container would contain not just the information about packages within it, but also about everything used to build it?

ADR/0039-spdx-support.md
```diff
@@ -0,0 +1,151 @@
...
+NOTE: Packages can be wrapped in a “virtual package”. Details are described in syft specific sbom details section.
+Relations of two documents needs to be merged together into single graph in a way which keeps the graph structure of the original graph of the main document (into which other document will be merged to). Once packages are merged together, relationships of the second document can be cleared off relations which refer to packages not included in the merged package list. SpdxElementId and relatedSpdxElement point to root document id of the second document should be replaced with root document id of the main document. If there’s “virtual package” in the second document, ids in relationships referring to it should be replaced with “virtual package” of the main document or main document id directly (if there’s no “virtual package”)
```
a-ovchinnikov, line 134: > relationships of the second document can be cleared off relations which refer to packages not included in the merged package

`can` or `must`? Also which packages will be excluded?

ADR/0039-spdx-support.md
xsuchy: It depends. If the relationship is DEPENDS_ON then it match normal `Requires` from RPM spec. If the relationship is BUILD_DEPENDENCY_OF then it matches 'BuildRequires'  and then yes - it would contain information about packages used to build it. This may be required for SBOMs of type `Build`. For more information about relationship types see https://spdx.github.io/spdx-spec/v2.3/relationships-between-SPDX-elements/ For SBOM types see https://www.cisa.gov/sites/default/files/2023-04/sbom-types-document-508c.pdf

ADR/0039-spdx-support.md
xsuchy: You can have one document with `pkg:maven/org.apache.commons/commons-lang3@3.12.0` and a second document with `pkg:github/apache/commons-lang@3.12.0`. Ie. one package comes from maven repository and the other from GH repository. 
For RHEL this is IMO theoretical option. But it may be likely for containers or layered products.

chmeliik: Or you could have `pkg:npm/requests@1.0.0` and `pkg:pypi/requests@1.0.0`. Or any other common library name in any two ecosystems

chmeliik: Or even a package with the same name and version in the _same_ ecosystem, which is actually a different package. This is surprisingly common in the Go world, especially in the context of OpenShift where many modules are forks of Kubernetes module. See for example https://github.com/openshift/cluster-api-provider-aws/blob/master/go.mod (note that the module name matches the name of the original module, not of the fork).

Depending on how the SBOM-generation tool handles this case, the purl may either be completely indistinguishable from the original, or may include the fork url as an attribute, or may use the name of the fork instead of the original name

ADR/0039-spdx-support.md
twaugh: @midnightercz I think virtual packages should be completely avoided unless they solve a real problem. (The kind of problem they _might_ solve: describing the build root for an RPM.)

I'm concerned about "base image" here: I think we're talking about builder images, not parent images.

And the view @chmeliik has here agrees with what Rogue and I discussed last week: we should include references (by purl) to the builder images, but we should not include references to their content.

chmeliik: > I'm concerned about "base image" here: I think we're talking about builder images, not parent images.

Asi in, "parent image" = the last FROM instruction, "builder image" = any FROM instruction other than the last?

Currently, Konfux reports both of those in the `.formulations[]`, distinguished by a [custom property](https://github.com/konflux-ci/build-tasks-dockerfiles/blob/c34cbcd4274879871d0e968a78f9704b643ba7b3/sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py#L52-L55)

twaugh: > Asi in, "parent image" = the last FROM instruction, "builder image" = any FROM instruction other than the last?

Exactly.

midnightercz: > These make sense to me. The other BaseImages relations not so much - the SBOM doesn't describe the base images, those have their own SBOMs

So the you basically want:
```
SPDXRef-DOCUMENT -> DESCRIBES: SPDXRef-Image
SPDXRef-DOCUMENT -> CONTAINS: SPDXRef-Image
SPDXRef-Image -> CONTAINS: PackageA
SPDXRef-BuilderImage-Package-1-> BUILD_TOOL_OF: SPDXRef-Image
```
So for purl, I assume it will be something like:
`pkg:docker/ubi9/ubi@sha256:abcdef?repository_url=registry.redhat.io`

ADR/0039-spdx-support.md
midnightercz: yeah, "must" better describes the process. To what packages will be excluded, basically "virtual packages" of the secondary documents and all packages in secondary documents which are "squashed" into existing packages in main documents or duplicate packages

ADR/0039-spdx-support.md
```diff
@@ -0,0 +1,151 @@
...
+
+SPDX SBOM format enables additional features not available in cyclondedx like multiple purl attributes per component. SPDX is also a widely adopted standard for software bill of materials.
+This ADR describes how to enable use of SPDX SBOM format in Konflux.
+
+## Decision
+
+
+### SBOM lifecycle in build pipeline
+
+At the start SBOMs are generated by cachi2 and syft. At later phase of the build pipeline, SBOMs from base images are extracted and merged toghether with SBOMs generated for the currently build container. When switching to spdx, newly built container image can be used as base image another container image. Therefore any tooling/task which works with sboms has to be able to work with both formats.
```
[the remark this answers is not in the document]
midnightercz: Actually it's not accurate. We discussed this and sboms from parent images are not processed by konflux pipeline. I need to update this section

ADR/0039-spdx-support.md
twaugh: That looks correct except for
> SPDXRef-DOCUMENT -> CONTAINS: SPDXRef-Image

chmeliik: And the purl will be pk:oci rather than pkg:docker, same as the builder/base images (but the exact format doesn't matter that much for this ADR(. And FYI, Ales is already working on the purl https://github.com/konflux-ci/build-tasks-dockerfiles/pull/181

ADR/0039-spdx-support.md
```diff
@@ -0,0 +1,144 @@
...
+| components            | packages            |
+| component.name        | package.name        |
+| component.versionInfo | package.version     |
+|-----------------------|---------------------|
+
+
+#### Component.purl
+CycloneDX (version 1.4) supports only a single purl attribute per component. SPDX doesn’t have a direct attribute, but instead every package includes an externalRefs array which describes all external references for the package. There are defined reference categories and types. For PURL, category PACKAGE-MANAGER and type purl is used. The purl itself will be stored as referenceLocator
+
+|------------------------------|---------------------------------------------------------------|
```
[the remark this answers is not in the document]
twaugh: The fix would be to use `<br>` for line breaks I think.

twaugh:
It looks like this adds _two_ identical ADRs with different numbers?

ADR/0044-spdx-support.md
```diff
@@ -0,0 +1,162 @@
...
+
+| CycloneDX Attribute             | SPDX Attribute                                             |
+|---------------------------------|------------------------------------------------------------|
+| Formulations.components = [{}]  | Relationships = [                                          |
+|                                 |     {                                                      |
+|                                 |          spdxElementId = `<ROOT-DOCUMENT-ID>`,             |
+|                                 |          relationshipType=DESCRIBES,                       |
+|                                 |          relatedSpdxElement=`<CONTAINER-IMAGE-ID>`,        |
+|                                 |     },                                                     |
```
twaugh, line 119: I don't understand why we'd need to add an additional `DESCRIBES` relationship for the builder image. It's connected to the `DESCRIBE`d image transitively by the `BUILD_TOOL_OF` relationship already isn't it? Here's a simple example (in tag form) of a similar situation: only the "thing" has a `DESCRIBES` relationship, not the other things which were used to build it or which it was build from: https://github.com/spdx/spdx-examples/blob/master/software/example1/spdx2.2/example1.spdx

midnightercz:
> It looks like this adds _two_ identical ADRs with different numbers?

I forgot to push mv changes. Now it should be in sync

ADR/0044-spdx-support.md
midnightercz: DESCRIBES relations is for container image of image which is being processed, not for builder image.

twaugh (approved):

chmeliik (commented):
I have more concerns over the content, but I haven't had time to do a thorough re-review. I'll hopefully get to it tomorrow

ADR/0044-spdx-support.md
```diff
@@ -0,0 +1,162 @@
...
+
+| CyloneDX Attribute                             | SPDX Attribute                                    |
+|------------------------------------------------|---------------------------------------------------|
+| Metadata.tools = [{“vendor”: “X”, “name”: “Y”] | CreationInfo.creators = [“Vendor: X”, “Tool: Y”]  |
+
+#### Merging SPDX
+##### Packages
+Packages of two SPDX documents can be merged together as a concatenation of two lists. In cycloneDX component elements can have only a single purl attribute, therefore component elements representing packages with the same name and version but with different purl have to be stored as multiple elements. SPDX package elements can bear multiple purls. Therefore multiple cycloneDX components can be squashed together into single SPDX package element with purls concatenated into a single list. Following rules are applied to packages merging process:
+- Packages with the same purl's package name and version and type are squashed into single package element
```
chmeliik, line 139: Similar concern to https://github.com/konflux-ci/architecture/pull/213#discussion_r1843490460 - this shouldn't be the authoritative descision on how to merge packages. Not all tools should do it this way, cachi2 certainly shouldn't

chmeliik: Perhaps we can just re-word this to make it clear that this is a sort of default suggestions for how to do the merging if you don't have more specific requirements

ADR/0044-spdx-support.md
midnightercz: Cachi2 doesn't use metadata.component and for cdx format, builder deps are populated into formulations. I think metadata.components shouldn't even be mentioned here as we don't work with it explicitly.

ADR/0044-spdx-support.md
```diff
@@ -0,0 +1,162 @@
...
+This ADR describes how to enable use of SPDX SBOM format in Konflux.
+
+## Decision
+
+### SBOM lifecycle in build pipeline
+
+At the start SBOMs are generated by [cachi2](#references) and [syft](#references). These two SBOM files are merged together into single SBOM document. At later phase of the build pipeline, builder images of the currently build image are added into SBOM as build dependency of the image. To switch to SPDX format, all tools producing and processing SBOMS in the pipeline has to be able to work with SPDX format. SBOMS of builder images are not processed by the pipeline, therefore builder images SBOMs doesn't have to be in SPDX format. This leads to fact that when tools generating the SBOMs are switched to SPDX format, all tools processing SBOMS can expect SPDX format only. There's no need for any tool to be able to work with mixed inputs of SPDX and CycloneDX formats
+As a result, all tools and tasks should implement the sbomType attribute to specify the expected SBOM format for input and output. This will allow tools to be tested with SPDX before the entire pipeline transitions to this format.
```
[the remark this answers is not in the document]
midnightercz: Ok, it makes some sense for tools. But I assume we want to have that attribute for tekton tasks which generates sboms right?

ADR/0044-spdx-support.md
```diff
@@ -0,0 +1,162 @@
...
+                - `List<Annotation>`
+                    - annotationDate
+                    - annotationType
+                    - annotator
+                    - Comment
+            - relationships
+                - `List<Relationship>`
+                    - spdxElementId
+                    - relationshipType
+                    - relatedSpdxElement
```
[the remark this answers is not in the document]
midnightercz: Yeah, it should be. I messed up indentation when I was copying the text from gdoc

ADR/0044-spdx-support.md
```diff
@@ -0,0 +1,162 @@
...
+
+#### Metadata.tools
+The CycloneDX metadata.tools sub attributes that we are mostly interested in are the vendor and name elements. Information about the creation of the SPDX document can be stored into creationInfo. CreationInfo.creators element is basically a list of strings. There’s a vague specification ([here](https://spdx.github.io/spdx-spec/v2.3/document-creation-information/#68-creator-field)]) about how it should be structured in the standard. Strings should be formatted in the following way: `<Attribute>: <Value>`. For example vendor should be stored as `Vendor: <vendor>`
+
+| CyloneDX Attribute                             | SPDX Attribute                                    |
+|------------------------------------------------|---------------------------------------------------|
+| Metadata.tools = [{“vendor”: “X”, “name”: “Y”] | CreationInfo.creators = [“Vendor: X”, “Tool: Y”]  |
```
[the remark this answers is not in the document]
midnightercz: Yeah, I guess it can be misleading. I'll change it to "tool" only

ADR/0044-spdx-support.md
```diff
@@ -0,0 +1,162 @@
...
+
+##### Relationships
+SPDX relationships represent graph/tree structure of relations of elements in the document. The Root element is the SPDX document itself (with SPDXID SPDXRef-Document).
+SPDX Root document typically contains a package representing source used for generating the SBOM. This can be container image, directory, etc. Root document is in relationship DESCRIBES with this source package. Other packages are in specific relationships with the source package. See also [syft specific sbom details](#syft-specific-sbom-details)
+
+Relations of two documents needs to be merged together into single graph in a way which keeps the graph structure of the original graph of the main document (into which other document will be merged to). Once packages are merged together, relationships of the second document must be cleared off relations which refer to packages not included in the merged package list. SpdxElementId and relatedSpdxElement point to root document id of the second document should be replaced with root document id of the main document. Source package element id in the second documents needs to be replaced with source package element id of the main document.
```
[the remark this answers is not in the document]
midnightercz: Sure

ADR/0044-spdx-support.md
midnightercz: yeah, good point

ADR/0044-spdx-support.md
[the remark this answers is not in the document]
chmeliik: Yeah, the "output type" param definitely makes sense

ADR/0044-spdx-support.md
```diff
@@ -0,0 +1,162 @@
...
+
+#### Component.purl
+CycloneDX (version 1.4) supports only a single purl attribute per component. SPDX doesn’t have a direct attribute, but instead every package includes an externalRefs array which describes all external references for the package. There are defined reference categories and types. For PURL, category PACKAGE-MANAGER and type purl is used. The purl itself will be stored as referenceLocator
+
+| CycloneDX Attribute          | SPDX Attribute                                                |
+|------------------------------|---------------------------------------------------------------|
+| component.purl = `<PURL>`    | package.externalRefs = [{referenceCategory:”PACKAGE-MANAGER”, |
+|                              |                          referenceType:purl,                  |
+|                              |                          referenceLocator: `<PURL>`           |
+|                              |                          }]                                   |
```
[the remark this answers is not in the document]
midnightercz: I believe this was fixed. All tables look correct to me

ADR/0044-spdx-support.md
chmeliik: Now that we have a section for `metadata.component`, could this go there?

ADR/0044-spdx-support.md
```diff
@@ -0,0 +1,279 @@
...
+
+| CyloneDX Attribute                             | SPDX Attribute                                    |
+|------------------------------------------------|---------------------------------------------------|
+| Metadata.tools = [{“vendor”: “X”, “name”: “Y”] | CreationInfo.creators = [“Tool: Y”]  |
+
+#### Metadata.component
+Metadata component describes component which is the component which whole SBOM is related to. For example If SBOM describes internal components and dependencies of a container image, this component should represent the container image itself. In SPDX, a package which is equivalent to this component is root package (see [root-packages](#root-packages) to have idea how syft handles this package). This package is in relationship `SPDXRef-ROOT` `DESCRIBES` `SPDXRef-RootPackage`. If there's no cycloneDX metadata.component attribute, SPDX SBOM already includes this "root package" and shouldn't be changed. If there is metadata.component attribute, existing root package should be replaced with the new one representing the SBOM document. All SPDX IDs in relationships refering to the old root package should be replaced with SPDX ID of the new root package. Example:
```
chmeliik, line 139: > SPDX SBOM already includes this "root package" and shouldn't be changed

I'm not sure I understand the context here. Are we talking about merging a CycloneDX doc into an existing SPDX doc? I thought this was about converting CycloneDX to SPDX

ADR/0044-spdx-support.md
```diff
@@ -0,0 +1,279 @@
...
+We run syft/cachi2 on source directory of a project which should be build into container. Generated SBOM contains
+```json
+{
+  "SPDXID": "SPDXRef-DOCUMENT",
+  ...
```
chmeliik, line 144: nitpick: the JSON syntax errors are distracting (especially in the un-rendered source, which is entirely red now). You could avoid them by using `jsonc` for the syntax and using comments
```suggestion
```jsonc
{
  "SPDXID": "SPDXRef-DOCUMENT",
  // ...
```

ADR/0044-spdx-support.md
```diff
@@ -0,0 +1,279 @@
...
+                "algorithm": "SHA256",
+                "checksumValue": "9ac75c1a392429b4a087971cdf9190ec42a854a169b6835bc9e25eecaf851258"
+            }
+        ],
+        ...
+        "externalRefs": [
+            {
+                "referenceCategory": "PACKAGE-MANAGER",
+                "referenceType": "purl",
+                "referenceLocator": "pkg:oci/container-registry.com/my-org/my-image@sha256:9ac75c1a392429b4a087971cdf9190ec42a854a169b6835bc9e25eecaf851258"
```
chmeliik, line 205: In the interest of aligning with https://github.com/konflux-ci/build-tasks-dockerfiles/pull/181, I think these should be 
```suggestion
      {
        "name": "my-image",
        "SPDXID": "SPDXRef-image",
        "versionInfo": "latest",
        ...
        "checksums": [
            {
                "algorithm": "SHA-256",
                "checksumValue": "9ac75c1a392429b4a087971cdf9190ec42a854a169b6835bc9e25eecaf851258"
            }
        ],
        ...
        "externalRefs": [
            {
                "referenceCategory": "PACKAGE-MANAGER",
                "referenceType": "purl",
                "referenceLocator": "pkg:oci/my-image@sha256:9ac75c1a392429b4a087971cdf9190ec42a854a169b6835bc9e25eecaf851258?repository_url=container-registry.com/my-org/my-image"
```

ADR/0044-spdx-support.md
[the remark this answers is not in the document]
chmeliik: The loss of indentation hurts readability, especially for this one

image

How about changing them from Markdown tables to monospace blocks? Or just wrapping the tables in `` ``` `` so that they look the same as in the source code

ADR/0044-spdx-support.md
midnightercz: But this is not related to metadata.component. Metadata.component is translated as new root package of spdx. Formulations.components are translated as packages with BUILD_TOOL_OF relationships to the root package

ADR/0044-spdx-support.md
midnightercz: It's just example, it doesn't really have to be 1:1 with real world output, but ok

ADR/0044-spdx-support.md
```diff
@@ -0,0 +1,162 @@
...
+### CycloneDX -> SPDX conversion
+
+CycloneDX (1.4) is structured document in json format with following structure (not full specification)
+
+- Document Root
+    - Metadata
+        - Tools
+            - `List<Tool>`
+                - vendor
+                - name
```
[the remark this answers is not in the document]
midnightercz: fixed

ADR/0044-spdx-support.md
chmeliik: This hunk is for `ROOT-DOCUMENT-ID DESCRIBES CONTAINER-IMAGE-ID`. That's the `metadata.component` no?

chmeliik (commented):
@allda, you and I had a private conversation in which Aleš mentioned that our SPDX docs were invalid according to the online validator. Could you update this according to the required changes? The rest LGTM

chmeliik:
Oh and this thread still needs a resolution https://github.com/konflux-ci/architecture/pull/213#discussion_r1849890331

ADR/0044-spdx-support.md
midnightercz: Ah, I see. But that's mentioned there only for completeness

ADR/0044-spdx-support.md
```diff
@@ -0,0 +1,302 @@
...
+```
+
+#### Component.properties
+CycloneDX components properties describe mapping of string:string properties for given component. SPDX component doesn’t have anything similar to cyclonedx properties. SPDX Package annotations are the only attribute where custom data can be stored and the only “customizable” field where there is comment which is a simple string. Due to that fact, cycloneDX property in format of {“name”: <string>, “value”: <string>} is encoded into json string. There can be also annotations produced by other tools. Therefore to be able to tell annotation comment is json encoded, annotator should ends with string “:jsonencoded”. To indicate annotator was a tool, prefix “Tool:" has to be included in the field.
+
+```
+| CycloneDX Attribute                       | SPDX Attribute                                 |
+|-------------------------------------------|------------------------------------------------|
+| components.properties = [                 | package.annotations = [                        |
+|   {“name”: …, “value”: …}                 |   {..., annotator: "`Tool:<tool>`:jsonencoded” |
```
chmeliik, line 111:
```suggestion
|   {“name”: …, “value”: …}                 |   {..., annotator: "`Tool: <tool>`:jsonencoded” |
```

ADR/0044-spdx-support.md
```diff
@@ -0,0 +1,302 @@
...
+#### SPDX specific attributes
+
+There are SPDX attributes which are required to be present in the document, however there's
+no cycloneDX equivalent for them. These attributes are:
+
+##### Document.documentNamespace
+
+documentNamespace is URI which provides way how locate the document or reference it other documents. When creating SPDX document locally via syft or cachi2, this attribute has no meaning
+as it's not clear yet how the document will be published. Therefore it's set to `NOASSERTION`.
+However when the document is copied to a container. It would make sense to provide a link to the document container
```
chmeliik, line 252: @Allda thoughts? Currently, the image-index SBOM uses a made up documentNamespace https://github.com/konflux-ci/build-tasks-dockerfiles/blob/377a181fc0ae4bbcbccfef9a5fc677c6a7b322b8/sbom-utility-scripts/scripts/index-image-sbom-script/index_image_sbom_script.py#L154

ADR/0044-spdx-support.md
```diff
@@ -0,0 +1,302 @@
...
+
+##### Package.downloadLocation
+downloadLocation is URI which provides way how to download the package. When creating SPDX document locally via syft, this information is not available. Also syft doesn't provide anything to this field. Therefore it's set to `NOASSERTION`.
```
chmeliik, line 255: More general rather than just focusing on Syft please :pray: 
```suggestion
downloadLocation is URI which provides way how to download the package. This is not always available, and it's not clear if it's useful. Therefore it's set to `NOASSERTION`.
```

ADR/0044-spdx-support.md
chmeliik: So what do you want to do with this? Remove it or move to correct section?

ADR/0044-spdx-support.md
midnightercz: Why would you want to have space there?

chmeliik: I thought the missing space was a typo, based on the other mention of `Tool: <tool>` in this doc and https://spdx.github.io/spdx-spec/v2.3/document-creation-information/#68-creator-field

ADR/0044-spdx-support.md
twaugh: The `documentNamespace` is not intended to be interpreted as a URL. It is simply a unique name for the document, so that `externalDocumentRefs` can refer to other documents in the same collection. See §6.5.2, "SPDX document namespace field - Intent":
> NOTE: The URI does not have to be accessible. It is only intended to provide a unique ID. In many cases, the URI will point to a Web accessible document, but this should not be assumed to be the case.
NOTE: The URI does not have to be accessible. It is only intended to provide a unique ID. In many cases, the URI will point to a Web accessible document, but this should not be assumed to be the case.

https://spdx.github.io/spdx-spec/v2.3/document-creation-information/#652-intent

chmeliik: Ack. We probably shouldn't use `NOASSERTION` then, and the way `index_image_sbom_script.py` does it seems fine

ADR/0044-spdx-support.md
midnightercz: It looks like both variants are valid (according to validation tool) . I personally prefer one without space as when converting back to cyclonedx you can simply remove the prefix, but in the spec there is a space. So maybe lets put there a space just to be spec compatible.

ADR/0044-spdx-support.md
midnightercz: I can remove it if you think it's rather confusing than informative

chmeliik (approved):
LGTM

ADR/0044-spdx-support.md
```diff
@@ -0,0 +1,296 @@
+# 44. SPDX SBOM support
+
+Date: 2024-09-24
+
+## Status
+
+Proposed
```
arewm, line 7: Is the status still proposed or is it accepted now?

scoheb (approved):

ralphbean (approved):

PR #216, opened 2024-11-06
ADR 45: Guidelines for developing tasks

arewm:
Presenting broad guidelines for where to contribute different types of tasks as well as patterns that should be generally followed in the tasks in the locations.

On the 2024-11-05 Konflux community architecture call, we discussed whether there were any guidelines for where to contribute tasks to. The result of that call was that we need to create an ADR outlining the recommendations so that we can also enable more targeted conversation.

The recording for the call can be found on YouTube: https://www.youtube.com/watch?v=uJbNdceDrTg

### DIFF
--- a/ADR/0045-guidelines-for-developing-tasks.md
+++ b/ADR/0045-guidelines-for-developing-tasks.md
@@ -1,0 +1,54 @@
+# 45. Guidelines for developing tasks
+
+* Date: 2024-11-06
+
+## Status
+
+Proposed
+
+## Context
+
+One of the architectural goals for Konflux is extensibility. The core system provides opinionated [build pipelines]
+and [release pipelines], but lets users extend those and/or create their own. Beyond this goal, however, there has
+not been any clear guidance about where contributions should be added or the types of functionality that are typical
+for the stages of software development.
+
+## Decision
+
+* If any process is required to modify artifacts themselves, it should be done in [build pipelines]. This will ensure
+  that Konflux users can appropriately test and verify these artifacts before triggering releases with them.
+* Any tasks which need to be included in the artifacts' provenance (enabling verification with EC policies) must be
+  included in the [build pipelines].
+* Whenever possible, a task that just changes the format of an artifact (for example creating a tar from a container image)
+  should be performed in the same pipeline that produced the artifact. These modified artifacts should be attached using
+  the referrer's API and the blob digest reported via results to enable verification that the artifacts were produced
+  from a trusted task.
+* Artifact-modifying steps should generally be avoided in [release pipelines]. When interacting with produced artifacts,
+  the release pipeline should be limited to extracting artifacts from storage (i.e. an OCI registry), copying/pushing
+  artifacts to target locations, and updating metadata associated with the artifacts.
+* Reusable tasks or pipelines that are intended to be used in IntegrationTestScenarios should be contributed to
+  [pipeline samples].
+
+## Consequences
+
+* All artifacts produced in Konflux should have provenance attestations. These attestations will include information
+  about which tests have run.
+* Whenever possible, the provenance attestations should be made available with the released artifacts. It may be
+  necessary to update this metadata during the release process (for example, signing provenance with a different key
+  or updating the provenance to indicate that a build was hermetic). If it is not possible to publish the provenance
+  as-is, a summary attestation can be used instead.
+* It should be possible to test any artifact that is going to be released.
+
+## Footnotes
+
+* It is not currently possible to produce trusted build-time tasks without pushing them to the [build definitions]
+  repository. Once the process for trusting tasks from various locations has been established, a new ADR should be
+  created to update the recommendations presented here.
+* Once it is possible to generate attestations for integration tests, the requirement to add all gating tasks to the
+  [build definitions] can be relaxed. This will, of course, require that the tasks and their results can be
+  appropriately trusted.
+
+[build pipelines]: https://github.com/konflux-ci/build-definitions/
+[build definitions]: https://github.com/konflux-ci/build-definitions/
+[release pipelines]: https://github.com/konflux-ci/release-service-catalog
+[pipeline samples]: https://github.com/konflux-ci/pipeline-samples


ADR/0045-guidelines-for-developing-tasks.md
```diff
@@ -0,0 +1,54 @@
...
+* It is not currently possible to produce trusted build-time tasks without pushing them to the [build definitions]
+  repository. Once the process for trusting tasks from various locations has been established, a new ADR should be
+  created to update the recommendations presented here.
+* Once it is possible to generate attestations for integration tests, the requirement to add all gating tasks to the
+  [build definitions] can be relaxed. This will, of course, require that the tasks and their results can be
+  appropriately trusted.
+
+[build pipelines]: https://github.com/redhat-appstudio/build-definitions/
+[build definitions]: https://github.com/redhat-appstudio/build-definitions/
+[release pipelines]: https://github.com/redhat-appstudio/release-service-catalog
```
johnbieren, line 53: `s/redhat-appstudio/konflux-ci/`

ADR/0045-guidelines-for-developing-tasks.md
```diff
@@ -0,0 +1,54 @@
...
+## Decision
+
+* If any process is required to modify artifacts themselves, it should be done in [build pipelines]. This will ensure
+  that Konflux users can appropriately test and verify these artifacts before triggering releases with them.
+* Any tasks which need to be included in the artifacts' provenance (enabling verification with EC policies) must be
+  included in the [build pipelines].
+* Whenever possible, a task that just changes the format of an artifact (for example creating a tar from a container image)
+  should be performed in the same pipeline that produced the artifact. These modified artifacts should be attached using
+  the referrer's API and the blob digest reported via results to enable verification that the artifacts were produced
+  from a trusted task.
```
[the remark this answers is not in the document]
arewm: It enables users to verify that the format change is done properly before those artifacts are released. 

I am not sure what you mean by post-release here. These are modifications that are needed before releases can progress.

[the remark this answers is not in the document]
chmeliik: > It enables users to verify that the format change is done properly before those artifacts are released.

Maybe I'd need a better example than the one in this doc. When I `skopeo copy docker://registry.access.redhat.com/ubi9/ubi-micro:9.4 oci:/tmp/ubi`, I don't need to verify anything.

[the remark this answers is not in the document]
arewm: If you have an integration test scenario that does not use the ubi-migro image in registry but instead uses a tarball, you would need that tarball to verify the artifact.

The need to verify isn't always required by a human but by some process. It should be possible to have all artifacts-to-be-released verified _as is_. If testing/verifying, a user or process shouldn't have to replicate the modifications done during a release pipeline within integration pipelines.

davidmogar (approved):

ADR/0045-guidelines-for-developing-tasks.md
```diff
@@ -0,0 +1,54 @@
...
+## Context
+
+One of the architectural goals for Konflux is extensibility. The core system provides opinionated [build pipelines]
+and [release pipelines], but lets users extend those and/or create their own. Beyond this goal, however, there has
+not been any clear guidance about where contributions should be added or the types of functionality that are typical
+for the stages of software development.
+
+## Decision
+
+* If any process is required to modify artifacts themselves, it should be done in [build pipelines]. This will ensure
```
zregvart, line 18: I'm not aware any use cases that modify the artifacts from the build pipeline. Perhaps a footnote/appendix with examples of such modifications could be useful. Some OLM use cases I'm not aware of?

This could be even broader/stricter, e.g. "any production of the artifact must be performed on the build pipelines with the intent that the produced artifact is shipped as-is".

ADR/0045-guidelines-for-developing-tasks.md
```diff
@@ -0,0 +1,54 @@
...
+  that Konflux users can appropriately test and verify these artifacts before triggering releases with them.
+* Any tasks which need to be included in the artifacts' provenance (enabling verification with EC policies) must be
```
zregvart, line 20: :+1:

ADR/0045-guidelines-for-developing-tasks.md
```diff
@@ -0,0 +1,54 @@
...
+* Whenever possible, a task that just changes the format of an artifact (for example creating a tar from a container image)
+  should be performed in the same pipeline that produced the artifact. These modified artifacts should be attached using
+  the referrer's API and the blob digest reported via results to enable verification that the artifacts were produced
+  from a trusted task.
+* Artifact-modifying steps should generally be avoided in [release pipelines]. When interacting with produced artifacts,
+  the release pipeline should be limited to extracting artifacts from storage (i.e. an OCI registry), copying/pushing
+  artifacts to target locations, and updating metadata associated with the artifacts.
+* Reusable tasks or pipelines that are intended to be used in IntegrationTestScenarios should be contributed to
+  [pipeline samples].
+
```
zregvart, line 31: Do we want to add a bullet point to where/how Tasks are contributed to the build (or even release) pipelines here?

ADR/0045-guidelines-for-developing-tasks.md
```diff
@@ -0,0 +1,54 @@
...
+## Consequences
+
+* All artifacts produced in Konflux should have provenance attestations. These attestations will include information
+  about which tests have run.
```
zregvart, line 35: The word "test" could be a bit ambiguous, perhaps 
```suggestion
  about which artifact or source checks have run.
```

ADR/0045-guidelines-for-developing-tasks.md
```diff
@@ -0,0 +1,54 @@
...
+* Whenever possible, the provenance attestations should be made available with the released artifacts. It may be
```
zregvart, line 36: Not sure how this relates to the topic of developing tasks.

ADR/0045-guidelines-for-developing-tasks.md
```diff
@@ -0,0 +1,54 @@
...
+  necessary to update this metadata during the release process (for example, signing provenance with a different key
+  or updating the provenance to indicate that a build was hermetic). If it is not possible to publish the provenance
+  as-is, a summary attestation can be used instead.
+* It should be possible to test any artifact that is going to be released.
```
zregvart, line 40: Perhaps not related to the topic of developing tasks?

ADR/0045-guidelines-for-developing-tasks.md
```diff
@@ -0,0 +1,54 @@
...
+Proposed
+
+## Context
+
+One of the architectural goals for Konflux is extensibility. The core system provides opinionated [build pipelines]
+and [release pipelines], but lets users extend those and/or create their own. Beyond this goal, however, there has
+not been any clear guidance about where contributions should be added or the types of functionality that are typical
+for the stages of software development.
+
+## Decision
```
zregvart, line 16: Would mentioning [Task](https://enterprisecontract.dev/docs/ec-policies/task_policy.html) [policies](https://enterprisecontract.dev/docs/ec-policies/build_task_policy.html) be of value here? As in: "Tasks contributed to Konflux need to adhere to the Build and Task policies"

ADR/0045-guidelines-for-developing-tasks.md
johnbieren: https://github.com/konflux-ci/release-service-catalog/pull/656 modifies the sbom, which is an artifact from the build pipeline

ADR/0045-guidelines-for-developing-tasks.md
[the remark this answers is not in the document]
chmeliik: > The need to verify isn't always required by a human but by some process. It should be possible to have all artifacts-to-be-released verified as is. If testing/verifying, a user or process shouldn't have to replicate the modifications done during a release pipeline within integration pipelines.

Agreed, but still think that the example is really bad. The use case is purely hypothetical, nobody is ever going to create an integration test that requires an image to be in the weird format before testing the image. And in this example the release pipeline *doesn't do any modifications.* If we're calling `skopeo copy` a modification, then the release pipeline always modifies images on release

ADR/0045-guidelines-for-developing-tasks.md
```diff
@@ -0,0 +1,54 @@
...
+  repository. Once the process for trusting tasks from various locations has been established, a new ADR should be
+  created to update the recommendations presented here.
+* Once it is possible to generate attestations for integration tests, the requirement to add all gating tasks to the
+  [build definitions] can be relaxed. This will, of course, require that the tasks and their results can be
+  appropriately trusted.
+
+[build pipelines]: https://github.com/konflux-ci/build-definitions/
+[build definitions]: https://github.com/konflux-ci/build-definitions/
+[release pipelines]: https://github.com/konflux-ci/release-service-catalog
+[pipeline samples]: https://github.com/konflux-ci/pipeline-samples
```
flacatus, line 54: https://github.com/konflux-ci/tekton-integration-catalog lets add the integration catalog here.

ADR/0045-guidelines-for-developing-tasks.md
Tojaj: @arewm as my team is working on RPM builds in Konflux, let me describe how RPMs work.

RPMs are often unsigned right after a build process and they are signed later (usually after the package pass testing).

The RPM signature is stored in the RPM header so signed package has different file checksum than the unsigned version.

In a build system like Koji (used by Fedora, CentOS Stream or RH internally), signing happens on a signing server, and then the signed RPM package is imported back and stored with the original artifact. The original unsigned copy remains, but the signed copy is now also available for consumers.

As I understand the wording here, if we built an RPM in Konflux, test it and later decide to sign it, we would need to do the signing as a build task? 

I don't think it would be a problem, I just wonder if there are some implications we would need to know about between using a build pipeline vs using a regular tekton pipeline later in the process?

For the context, we plan to have a 3-level hierarchy of Konflux workspaces 1) Multiple dev workspaces where builds of RPMs happen 2) Integration workspace where all the RPMs are integrated and tested together as a coherent product 3) Release engineering workspace where release to production happens. So with that in mind, I would imagine we would like to do signing in the integration workspace (the middle one).

arewm: How would the integration workspace be viewed from the dev workspaces?

This ADR doesn't preclude users from creating artifact-modifying tasks/pipelines and injecting them in various places. As long as those tasks/pipelines are _themselves_ compliant with eventual policies configured. The ADR is more scoped for how we will look to integrate common operations into the tasks and pipelines that we build.

Tojaj: > How would the integration workspace be viewed from the dev workspaces?

It will be a regular workspace. Developers won't have access to it, or at least a direct one. The integration workspace will implement a business logic regarding composing the product and managing its releases and that logic will provide ways for developers to see about what's going on.

E.g. the business logic in the integration workspace will be maintaining data and metadata of the product (in a gitops way), so developers would be able to see how their components (packages) integrate with the rest of the product - e.g. to see in what product repositories they belong to and so on.

The bussines logic in the integration workspace may also provide an interface for developers to configure / tweak the business logic. Again via GitOps - via MRs. Devs would be able to submit MRs to, for example, change to which RPM repo their package should belong to, but ultimately there will be a team responsible for such productization work that will have to review and decide if to merge that MR or not.

PR #217, opened 2024-11-15
ADR 46: Build a common Task Runner image

chmeliik:
Related to https://issues.redhat.com/browse/STONEBLD-2982. Building the RHTAP task runner image was the final impulse, but I've felt the need for a unified runner image for some time now.

### DIFF
--- a/ADR/0046-common-task-runner-image.md
+++ b/ADR/0046-common-task-runner-image.md
@@ -1,0 +1,141 @@
+# 46. Build a common Task Runner image
+
+Date: 2024-11-15
+
+## Status
+
+Proposed
+
+## Context
+
+Tekton Tasks often depend on specific CLI tools.
+The tools come from a restricted set of container images
+(see `allowed_step_image_registry_prefixes` in the [policy data][rule-data]).
+
+### The current mess of Task containers
+
+If no image containing a required CLI tool already exists, the current approach
+is to build a new image specifically for that one tool. Or to add the tool to
+one or more of the existing images, if Task authors find that more convenient.
+Examples:
+
+* [yq-container]
+* [oras-container]
+  (also includes `yq` and a copy of one script from [build-trusted-artifacts])
+* [git-clone container][git-clone]
+  (for the `git-init` tool, also includes `find`)
+* [buildah-container]
+  (also includes a wild variety of tools such as
+  `dockerfile-json`, `rsync`, `kubectl`, `jq`, `subscription-manager` and others)
+
+Then, we have some use-case-oriented containers which are somewhat intertwined
+with the tool-oriented containers (or at least share software, sometimes installed
+using different approaches).
+
+* [build-trusted-artifacts]
+  (a set of Trusted Artifacts scripts, also includes `oras` and `jq`)
+* [source-container-build]
+  (script for building source containers, also includes `skopeo` and `jq`)
+
+And last, some Tasks use the [appstudio-utils] image, which contains a variety
+of tools installed straight from GitHub releases. Many of which are also available
+in the tool-oriented containers (and installed via more legitimate means).
+
+The current situation increases confusion, maintenance burden (both for the container
+maintainers and for [build-definitions] maintainers) and, in case of `appstudio-utils`,
+breaks good secure supply chain practices.
+
+### Splitting Tasks into steps
+
+The set of CLI tools you need may already be containerized, but in two or more separate
+containers. In that case, rather than adding the tools you need to one of the containers,
+the better solution could be to take advantage of Tekton Tasks' `steps` feature (each
+step can use a different container image).
+
+*Could* be, but isn't. In practice, what this achieves is:
+
+* Increased complexity of the Task code, since it typically requires splitting the
+  code in unnatural ways and sharing some data between Task steps. Inexperienced
+  Tekton users may not even think of this approach or know how to achieve it.
+* Increased compute resource requirements for the Task. The total resource requirements
+  for a Task are not the *maximum* of its steps' resource requirements, they are
+  the *sum* (see [Compute Resources in Tekton][compute-resources-in-tekton]).
+* Reduced size limit of the results that the task can return (unless the Tekton
+  installation enables [Results from sidecar logs][results-from-sidecar-logs]).
+
+### Konflux users and custom Tasks
+
+The Enterprise Contract team has developed the Trusted Artifacts concept to enable
+Konflux users to add custom Tasks to the pipelines without compromising the
+trustworthiness of the build.
+
+But Konflux users face the same difficulties described above (made worse by the
+fact that they don't tend to have much Tekton experience). The initial hurdle of
+finding/building the right container image for what they want to do may be too high.
+
+## Decision
+
+Build and maintain a common "Task Runner" image.
+
+The image must:
+
+* Include all the tools commonly needed by Konflux build tasks.
+* Build and release via Konflux, hermetically if possible.
+* Document the list of installed tools and their versions, similar to how GitHub
+  documents the [software installed in their runner images][github-runner-software].
+  * The list of tools is a public interface, both Konflux devs and Konflux users
+    can depend on it.
+* Use proper semver versioning. The deletion of a tool, or a change in the major
+  version of a tool, is a breaking change and must result in a major version change
+  for the Task Runner image.
+
+Gradually deprecate all the current tool-oriented container images and replace
+their usage with the common Task Runner image.
+
+The Task Runner image does not replace the more specialized use-case-oriented images,
+but they can use it as a base image if desirable.
+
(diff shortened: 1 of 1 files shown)


ADR/0046-common-task-runner-image.md
```diff
@@ -0,0 +1,138 @@
...
+By no longer relying on a Tekton-specific feature (steps), most Tasks become nothing
+more than a bash script wrapped in some YAML. It enables a saner approach to authoring
+Tasks. Write a bash script that works on your machine, wrap it in a bunch of YAML,
+verify that it works, ship it.
+
+May be a step towards reconciliation with the not-quite-sibling project, RHTAP. RHTAP
+supports not only Tekton, but also other CI systems. RHTAP takes the approach outlined
+above, where every task is one bash script. To enable this, RHTAP already builds
+a [runner image][rhtap-runner-image] but cannot use it for Tekton (RHTAP inherits
+Tekton tasks from Konflux).
```
arewm, line 123: I don't think that this should be relevant for the ADR. It is something specific to Red Hat itself and not Konflux.

ADR/0046-common-task-runner-image.md
```diff
@@ -0,0 +1,138 @@
...
+    can depend on it.
+* Use proper semver versioning. The deletion of a tool, or a change in the major
+  version of a tool, is a breaking change and must result in a major version change
+  for the Task Runner image.
+
+Gradually deprecate all the current tool-oriented container images and replace
+their usage with the common Task Runner image.
+
+The Task Runner image does not replace the more specialized use-case-oriented images,
+but they can use it as a base image if desirable.
```
arewm, line 96: Should we have any recommendation on when it makes sense to create these specialized images? Or how narrowly focused they should be? 

I proposed adding an additional endpoint to the build-trusted-artifacts image, but the contract team deemed that it would be better to put it somewhere else.

@zregvart @lcarva , fyi. I feel like this ADR is the opposite direction of what you were proposing.

chmeliik: The recommendation could be this: if you're thinking of including a tool in the Task Runner image and it does not follow semver and/or isn't easily installable as a standalone tool, don't include it.

As an example, that would mean these images should be kept standalone
* build-trusted-artifacts (is not a tool, is a set of bash scripts)
* the SBOM and source-container scripts in https://github.com/konflux-ci/build-tasks-dockerfiles (they have no versioning)
* cachi2 (has specific requirements for its execution environment)

ADR/0046-common-task-runner-image.md
chmeliik: dropped

ADR/0046-common-task-runner-image.md
```diff
@@ -0,0 +1,131 @@
...
+* Reduced size limit of the results that the task can return (unless the Tekton
+  installation enables [Results from sidecar logs][results-from-sidecar-logs]).
+
+### Konflux users and custom Tasks
+
+The Enterprise Contract team has developed the Trusted Artifacts concept to enable
+Konflux users to add custom Tasks to the pipelines without compromising the
+trustworthiness of the build.
+
+But Konflux users face the same difficulties described above (made worse by the
```
lcarva, line 72: Can you clarify who Konflux users are? I think of this as folks using Konflux (add application, component, etc), but from context could it be that you're referring to Konflux developers/maintainers?

ADR/0046-common-task-runner-image.md
```diff
@@ -0,0 +1,131 @@
...
+fact that they don't tend to have much Tekton experience). The initial hurdle of
+finding/building the right container image for what they want to do may be too high.
+
+## Decision
```
lcarva, line 76: Probably outside of the scope of this ADR, but I wonder if using this image in a Task would have an impact on RHTAP.

ADR/0046-common-task-runner-image.md
lcarva: > I proposed adding an additional endpoint to the build-trusted-artifacts image, but the contract team deemed that it would be better to put it somewhere else.

Yeah, because of the nature of the build-trusted-artifacts image/repo. The proposal here is to create a multi-tool repo/image, which by definition has a much wider scope.

> The recommendation could be this: if you're thinking of including a tool in the Task Runner image and it does not follow semver and/or isn't easily installable as a standalone tool, don't include it.

IMO the current state of those tools reflect the current state of getting code to run in the Konflux tasks. If we have a well-defined list of requirement for what it takes to be added to the runner image, then I expect these tools to be motivated to abide to them, especially if they are reasonable.

I expect a tool like oras to be compiled from source when building the runner image. Much like it's current done in https://github.com/konflux-ci/oras-container. The source is referenced via a git submodule. Would this be a reasonable approach for all of the tools?

ADR/0046-common-task-runner-image.md
```diff
@@ -0,0 +1,131 @@
...
+number of rebuilds needed due to CVEs stays the same but is not scattered across
+tool-container repos anymore.
+
+Tasks get easier to write because all the tools you need are available in the same
+image. For both Konflux devs and Konflux users.
+
+Tasks have lower resource requirements because it's no longer necessary to split
+them into steps.
+
+The Task Runner image is larger than any of the individual images used by the Tasks
```
lcarva, line 110: Also (maybe a stretch) there is a higher chance that the image will already be cached in the node.

ADR/0046-common-task-runner-image.md
```diff
@@ -0,0 +1,131 @@
...
+at present. But it's much smaller than all the individual images combined. And
+because Pipelines pull each image at most once, this is a win.
+
+By no longer relying on a Tekton-specific feature (steps), most Tasks become nothing
+more than a bash script wrapped in some YAML. It enables a saner approach to authoring
+Tasks. Write a bash script that works on your machine, wrap it in a bunch of YAML,
```
lcarva, line 116: Although I agree with this, it seems to be contrary to this [conversation](https://github.com/konflux-ci/architecture/pull/217/files#r1846022425). Are you saying that it's ok to have a mixture of random scripts in the runner image?

ADR/0046-common-task-runner-image.md
chmeliik: > The source is referenced via a git submodule. Would this be a reasonable approach for all of the tools?

With git submodules, it's kind of hard to to follow versions (or even to know what version you are building). Renovate just bumps them to the latest commit in main.

For tools written in Go, I was thinking of experimenting with a [`tools.go`-based](https://go.dev/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module) approach. We wouldn't include them as submodules, just declare them as dependencies. That gets you proper versioning for free. (That probably doesn't belong in this ADR though, as long as the image knows exactly what version of each tool it includes, it's fine)

> If we have a well-defined list of requirement for what it takes to be added to the runner image, then I expect these tools to be motivated to abide to them, especially if they are reasonable.

:100:

ADR/0046-common-task-runner-image.md
chmeliik: It's the former. I've tried to differentiate between "Konflux devs/Task authors" and "Konflux users". Would "Konflux end users" be better?

ADR/0046-common-task-runner-image.md
chmeliik: Exceptions would still need to exist.

`use-trusted-artifact` / `create-trusted-artifact` steps would still be needed (until TA becomes a standalone tool and the TA-generation tool takes that into account, if ever)

The `build-source-image` task would still use the image tailor-built for that one task, until the script becomes a standalone tool (if ever).

I'll rephrase this to make it clearer

ADR/0046-common-task-runner-image.md
chmeliik: I realized that "Pipelines pull each image at most once" is not true, Tasks don't run on the same node. I still think the common image is a win overall though. Rephrased

ADR/0046-common-task-runner-image.md
chmeliik: Softened the wording, added an example where we would still want to use steps

ADR/0046-common-task-runner-image.md
chmeliik: Added the suggested requirements for tool inclusion

ADR/0046-common-task-runner-image.md
lcarva: Re-reading this, I think it's fine.

arewm (approved):

chmeliik:
@arewm if this is ready for merge, could you merge this please?

And what are the next steps? I suppose this needs a Konflux Feature?

lcarva (approved):

lcarva:
Let's merge.

PR #218, opened 2024-11-21
feat(KONFLUX-5434): ADR 47: Creating pending pipeline runs

jhutar:


ADR/0039-creating-pending-pipelineruns.md
```diff
@@ -0,0 +1,66 @@
...
+
+Accepted
+
+## Context
+
+There is a frequent use case amongst Konflux users when triggering large amount of builds: sometimes users do not care how quickly the pipeline run will finish, but they want it to finish eventually.
+Now what happens is that part of the pipeline runs started in this big bulk is blocked by unavailable resources (e.g. namespace memory limit quotas or MPC pool size).
+Because of how Tekton works, no matter what these pipeline runs are considered running (even if their pods are waiting for resources) and so eventually timeout setting strikes.
+
+So say you started 300 pipeline runs at Fry evening and you expect to see 300 new images at Mon morning.
```
scoheb, line 15:
```suggestion
So say you started 300 pipeline runs on Friday evening and you expect to see 300 new images by Monday morning.
```

ADR/0039-creating-pending-pipelineruns.md
```diff
@@ -0,0 +1,66 @@
...
+You have resource quota for 20 concurrent builds, so 20 builds get started, when they finish 20 more and so on.
+But because all 300 pipeline runs started (from Tekton point of view) when they were created, after few iterations all the remaining pipeline runs will be cancelled (as timeouted).
```
scoheb, line 17:
```suggestion
But because all 300 pipeline runs started (from Tekton point of view) when they were created, after few iterations all the remaining pipeline runs will be cancelled (due to a timeout).
```

ADR/0039-creating-pending-pipelineruns.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Status
+
+Accepted
+
+## Context
+
+There is a frequent use case amongst Konflux users when triggering large amount of builds: sometimes users do not care how quickly the pipeline run will finish, but they want it to finish eventually.
+Now what happens is that part of the pipeline runs started in this big bulk is blocked by unavailable resources (e.g. namespace memory limit quotas or MPC pool size).
+Because of how Tekton works, no matter what these pipeline runs are considered running (even if their pods are waiting for resources) and so eventually timeout setting strikes.
```
arewm, line 13: Trying to present the problem more clearly. Feel free to take any or none of the suggestion (some of this suggestion blurs into the next paragraph too).
```suggestion
When pipelines start, the time is immediately being taken away from the configured timeout (even if their pods are waiting for resources). While this behavior is likely sufficient when fewer concurrent pipelines are running, timeouts are more likely to be exceeded when bulk builds are triggered (i.e. a Renovate job updating common dependencies). While the speed at which a pipeline completes is important, a delayed and completed pipeline will likely finish quicker than an immediately and timeout-killed pipeline. Futhermore, having multiple pipelines running at the same time may starve other concurrent pipelines from acquiring the necessary resources resulting in fewer completed pipelines.
```

ADR/0039-creating-pending-pipelineruns.md
arewm, line 17: As I mentioned above, the case could be that none of the pipelines completed because they were all starving each other of the needed resources.

ADR/0039-creating-pending-pipelineruns.md
```diff
@@ -0,0 +1,66 @@
...
+
+So say you started 300 pipeline runs at Fry evening and you expect to see 300 new images at Mon morning.
+You have resource quota for 20 concurrent builds, so 20 builds get started, when they finish 20 more and so on.
+But because all 300 pipeline runs started (from Tekton point of view) when they were created, after few iterations all the remaining pipeline runs will be cancelled (as timeouted).
+
+Another use case for what we are proposing is protecting cluster (e.g. etcd) from spikes of running lots of concurrent pipeline runs.
```
arewm, line 19:
```suggestion
When many pipelines are triggered across multiple namespaces, there is also an increased load on etcd which may degrade the cluster's overall performance.
```

ADR/0039-creating-pending-pipelineruns.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Decision
+
+We will use idea of creating pipeline runs in pending state and having a standalone controller that would implement any custom logic on when to start each pending pipeline:
+
+1. We add `.spec.status: "PipelineRunPending"` to pipeline run YAMLs.
```
arewm, line 25: The numbering is off in this section.

ADR/0039-creating-pending-pipelineruns.md
```diff
@@ -0,0 +1,66 @@
...
+
+So say you started 300 pipeline runs at Fry evening and you expect to see 300 new images at Mon morning.
+You have resource quota for 20 concurrent builds, so 20 builds get started, when they finish 20 more and so on.
+But because all 300 pipeline runs started (from Tekton point of view) when they were created, after few iterations all the remaining pipeline runs will be cancelled (as timeouted).
+
+Another use case for what we are proposing is protecting cluster (e.g. etcd) from spikes of running lots of concurrent pipeline runs.
+
+## Decision
+
+We will use idea of creating pipeline runs in pending state and having a standalone controller that would implement any custom logic on when to start each pending pipeline:
```
arewm, line 23: Is this proposed as a Konflux core or addon component?

Many of the considerations refer to the multi-platform controller, but that is not a core component.

ADR/0039-creating-pending-pipelineruns.md
```diff
@@ -0,0 +1,66 @@
...
+
+1. We add `.spec.status: "PipelineRunPending"` to pipeline run YAMLs.
+1. We will implement controller, that will contain code that will decide when to start (by just nulling `.spec.status` field - then Pipelines Controller takes over) what pipeline run. Some possibilities come to mind:
+   1. Controller will keep an eye on current number of concurrently running pipeline runs in the cluster (using Prometheus metric `avg(tekton_pipelines_controller_running_pipelineruns)`), and if that is below given threshold, it starts the pipeline run(s).
+   1. If pipeline run have `prsstarter.example.com/mpc_platforms: linux/arm64,linux/ppc64le,linux/s390x` annotation, controller checks if MPC have free capacity (using max for the pool and Prometheus metric `multi_platform_controller_running_tasks`) for mentioned platforms and if so, will start it.
```
arewm, line 28: Is this an annotation that is used today or is it a new proposed one? Can it be generalized as the multi-platform-controller isn't a core component?

The specific annotation values are implementations of individual Konflux deployments.

ADR/0039-creating-pending-pipelineruns.md
```diff
@@ -0,0 +1,66 @@
...
+2024-11-20 14:05:00,529 prs_starter.my_reconciler Thread-2 (my_reconciler) INFO The my-comp-2-on-push-zz6sw/jhutar-tenant (429faf3a-620d-4ccd-ac89-c7ba4db11c90) was started, not watching it anymore
+2024-11-20 14:05:00,529 prs_starter.my_reconciler Thread-2 (my_reconciler) DEBUG Reconciling my-comp-3-on-push-hj5bw/jhutar-tenant (ba328d32-e912-4a2c-98e8-506937577912)
+2024-11-20 14:05:00,529 prs_starter.my_reconciler Thread-2 (my_reconciler) DEBUG Missing capacity for my-comp-3-on-push-hj5bw/jhutar-tenant (ba328d32-e912-4a2c-98e8-506937577912) on linux/s390x
+```
+
+This is what was happening:
+
+* at 2024-11-20 14:04:44,646, we checked MPC capacity and s390x was fully used, so reconciling pipeline runs (that are in pending state) `my-comp-2-on-push-...` and `my-comp-3-on-push-...` did not started them
+* at 2024-11-20 14:04:57,309, we noticed MPC freed the capacity for s390x
+* at 2024-11-20 14:04:59,660 pipeline run `my-comp-2-on-push-...` was started, no luck for `my-comp-3-on-push-...` though
```
arewm, line 61: These types of capacity decisions are not always blocked on single events/quotas. When considering the multi-platform controller, there are multiple different platforms that may be requested. 

* How can we reconcile the queues for all of these platforms?
* It seems like we would need to add some sort of queuing mechanism for the pending PipelineRuns. How can we ensure fairness with all of the pipelines which may be waiting for different metrics/resources to indicate capacity?
* How can we ensure that a pipeline is not unnecessarily blocked by an unneeded resource just because it is waiting behind some other pipeline in a queue?

ADR/0039-creating-pending-pipelineruns.md
```diff
@@ -0,0 +1,66 @@
...
+
+Currently this is POC code only and besides lots of other things, we know about these issues that would need to happen:
```
arewm, line 63: You refer to code but don't reference it. Code isn't necessary for an ADR as we are instead trying to understand the motivation behind a problem and decide on the defining properties of functionality.

ADR/0039-creating-pending-pipelineruns.md
```diff
@@ -0,0 +1,66 @@
...
+
+Another use case for what we are proposing is protecting cluster (e.g. etcd) from spikes of running lots of concurrent pipeline runs.
+
+## Decision
+
+We will use idea of creating pipeline runs in pending state and having a standalone controller that would implement any custom logic on when to start each pending pipeline:
+
+1. We add `.spec.status: "PipelineRunPending"` to pipeline run YAMLs.
+1. We will implement controller, that will contain code that will decide when to start (by just nulling `.spec.status` field - then Pipelines Controller takes over) what pipeline run. Some possibilities come to mind:
+   1. Controller will keep an eye on current number of concurrently running pipeline runs in the cluster (using Prometheus metric `avg(tekton_pipelines_controller_running_pipelineruns)`), and if that is below given threshold, it starts the pipeline run(s).
```
arewm, line 27: How might this be reconciled between namespace quotas and cluster-wide performance?

ADR/0039-creating-pending-pipelineruns.md
```diff
@@ -0,0 +1,66 @@
+# 39. Creating pending pipeline runs
```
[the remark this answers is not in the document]
jhutar: Thank you, fixed.

ADR/0039-creating-pending-pipelineruns.md
jhutar: Hello! I have not experienced that, but it is definitely possible, especially in combination with some MPC bugs.

This is mine mental model of what is happening:

Say you have this pipeline:

1. init
2. clone
3. prefetch
4. [build-s390x, build-arm64]

The pipeline usually runs for 1 hour and have timeout set to 2 hours. MPC have capacity of 2 s390x and 2 arm64, all free.

Now you trigger 5 builds.

1. First 2 builds start, all 5 builds are considered started by Tekton, last 3 are waiting for VMs
2. Hour later, first 2 builds finish, second 2 builds start and last one is still waiting
3. Another hour later second 2 builds finish (just in time before timeout) and last, 5th, build fails because of timeout

So we end up with 4 finished builds and one failed because of timeout.

Anyway, I'm happy to change the wording as I think we both agree that the issue is there.

ADR/0039-creating-pending-pipelineruns.md
jhutar: Oh, I think it depends on what logic we will use. If we decide we want to protect etcd by capping number of concurrent pipelineruns, it can be core.

If we decide we want to protect multi-arch users from timeouting, it should be addon I guess.

Should I add this to the doc?

ADR/0039-creating-pending-pipelineruns.md
jhutar: It is just a random string I used in tests in Stage we did.

I was thinking about this as sort of plugable system:

* `prsstarter.example.com/mpc_platforms: linux/arm64,linux/ppc64le,linux/s390x` - means you want your pipeline run to be gated based on MPC pool availability
* `prsstarter.example.com/memory_limit_quota: 10GiB` - means you want your pipeline to be started only if your namespace have at least 10GiB free in memory limit quota
* `prsstarter.example.com/other_plugin: other_plugin_configuration` - means you want your pipeline to be gated by some other plugin

Should I just avoid specific annotations or stress it is just an example? How can I improve it in the doc?

ADR/0039-creating-pending-pipelineruns.md
jhutar: > How can we reconcile the queues for all of these platforms?

Here I was checking all platforms from that `prsstarter.example.com/mpc_platforms: linux/arm64,linux/ppc64le,linux/s390x` annotation, but `linux/s390x` was the only one that was not available. Code looks like this:

```
def can_pr_be_started(pr_required_platforms, mpc_available_slots)
    for platform in pr_required_platforms:
        if mpc_available_slots[platform] <= 0:
            return False
    return True
```

> It seems like we would need to add some sort of queuing mechanism for the pending PipelineRuns. How can we ensure fairness with all of the pipelines which may be waiting for different metrics/resources to indicate capacity?

You are right, we need to be reconciling (i.e. possibly starting) pipelineruns in order how they were created.

We can also introduce come sort of priority annotation that would also allow the pipelineruns to mess with the order (e.g. you are building security update and you do not want to wait for all ongoing rebuild of all RHEL 9.8 packages build).

> How can we ensure that a pipeline is not unnecessarily blocked by an unneeded resource just because it is waiting behind some other pipeline in a queue?

I was thinking about it in a way that each pipeline gated only by gating plugins it configures in it's annotations, so if I want to be gated by MPC, I use `prsstarter.example.com/mpc_platforms`, if I want to be gated by something else, I use `prsstarter.example.com/some_other_plugin`

And we can also have some global, implicit gateways like that etcd protection one that would be enabled for all pending pipelines by default.

Of course this all would only apply to pipeline runs created in this Pending state, so normal pipelines would just go ahead with kinda priority, as they would not be restricted by this gating.

ADR/0039-creating-pending-pipelineruns.md
jhutar: This is the code I have so far: https://gitlab.cee.redhat.com/jhutar/mystoneinst/-/blob/main/prs-controller.py?ref_type=heads

I just wanted to note that there are some blockers and I can also mention some problems I hit when creating and testing the code where I only have workarounds for them, not solutions.

Or I can just remove this section?

ADR/0039-creating-pending-pipelineruns.md
jhutar: I would expect that a pending pipeline run need to pass all the gates (AND) to be actually started. Say that MPC and etcd-protection gate applies to it: it need to pass both of them to be started.

But I'm unsure if this answers your question, sorry :)

ADR/0039-creating-pending-pipelineruns.md
jhutar: Should be OK now.

ADR/0039-creating-pending-pipelineruns.md
vdemeester: >  It seems like we would need to add some sort of queuing mechanism for the pending PipelineRuns. How can we ensure fairness with all of the pipelines which may be waiting for different metrics/resources to indicate capacity?

We could do a simple "FIFO" based on the creation dates.

vdemeester (commented):
Do we list "shortcomings" (or possible problems) in ADR ? Asking because as of today: it does hijack some pac behavior, making, for example, the github checks being updated only at the end.

arewm:
> Do we list "shortcomings" (or possible problems) in ADR ? Asking because as of today: it does hijack some pac behavior, making, for example, the github checks being updated only at the end.

Yes, I think that it is important to call these out.

ADR/0039-creating-pending-pipelineruns.md
arewm: This is certainly an option, but it would result in a single-arch build staying in pending until multi-arch builds at the queue HEAD have sufficient capacity scheduled? 

Or would there be separate queues for drastically different types of requests?

jhutar:
> > Do we list "shortcomings" (or possible problems) in ADR ? Asking because as of today: it does hijack some pac behavior, making, for example, the github checks being updated only at the end.
> 
> Yes, I think that it is important to call these out.

Aaah, thank you, added!

gbenhaim:
@jhutar thank you for writing this ADR. Aftr I read it and dedicated some time to think about the issue, here are my though:

I think it makes more sense to control the execution of a taskruns since it is being mapped to a pod, and the pod can have node selectors, etc that will affect the node it's going to be scheduled on.
I think that there is no need to block the execution of the entire pipeline if there are some resources it can use, for example multi arch pipeline, why block it entirely if only the nodes for one arch are not available.

In addition, if the taskrun will have the status field for controlling its execution, the management of task runs can be integrated with kueue https://github.com/kubernetes-sigs/kueue

jhutar:
> @jhutar thank you for writing this ADR. Aftr I read it and dedicated some time to think about the issue, here are my though:
> 
> I think it makes more sense to control the execution of a taskruns since it is being mapped to a pod, and the pod can have node selectors, etc that will affect the node it's going to be scheduled on. I think that there is no need to block the execution of the entire pipeline if there are some resources it can use, for example multi arch pipeline, why block it entirely if only the nodes for one arch are not available.
> 
> In addition, if the taskrun will have the status field for controlling its execution, the management of task runs can be integrated with kueue https://github.com/kubernetes-sigs/kueue

Hello @gbenhaim !

Just gating TaskRuns will not help with "timeouting PipelineRuns issue" which was a trigger for this - once PLR is started, it's timeout starts ticking. E.g. MPC job that passes amd64, arm64 and ppc64 builds that is waiting for s390x builder for 2 hours will timeout and all 3 built architectures are wasted.

Anyway I do not know kueue, so I can not really tell if it can help with this.

ggallen:
Isn't this something that should be handled in Tekton?  It seems like we are writing a scheduler.

jhutar:
> Isn't this something that should be handled in Tekton? It seems like we are writing a scheduler.

Unfortunately Tekton does not provide that. Goal here is to create kinda queuing mechanism.

ggallen:
> > Isn't this something that should be handled in Tekton? It seems like we are writing a scheduler.
> 
> Unfortunately Tekton does not provide that. Goal here is to create kinda queuing mechanism.

No, but Kubernetes does.  And I believe if the problem with Tekton timeouts was fixed, then you could just use the normal Kubenetes scheduler.

vdemeester:
> Isn't this something that should be handled in Tekton? It seems like we are writing a scheduler.

The reason there is `spec.status` and `Pending` are for Tekton to *not* have to know anything about "when" to start it. Aka it is there for users to be able to implement the behavior they would like. Essentially tektoncd/pipeline delegates the "when should I start" this PipelineRun (or TaskRun) to something else (You could see it as Deployment `spec.replicas` — Deployment object itself doesn't have a knowledge when to scale up or down its replicas, something else does).

>  And I believe if the problem with Tekton timeouts was fixed, then you could just use the normal Kubenetes scheduler.

What is "the problem with Tekton timeouts" here exactly ? If a TaskRun runs executes, it *has to have* a timeout ; same for a PipelineRun. In the MPC example, it is a TaskRun in the middle of the PipelineRun, so the PipelineRun is running, time is ticking.

And again, it's just normal Kubernetes scheduler. Here, the idea is a Go/No-go type of controller, it does not schedule anything, it's just an off->on switch based on some behavior. This can be used with a very simple behavior *or* with a very complex one, but it's up to the service/tool/people who know the behavior to *code* the behavior.

ggallen:
> What is "the problem with Tekton timeouts" here exactly ? If a TaskRun runs executes, it _has to have_ a timeout ; same for a PipelineRun. In the MPC example, it is a TaskRun in the middle of the PipelineRun, so the PipelineRun is running, time is ticking.

It's the basis for this entire ADR, and in the first sentence:

> When pipelines start, the time is immediately being taken away from the configured timeout (even if their pods are waiting for resources).

The timeout used by Tekton starts when the pipeline is submitted, not when it actually starts.

We've talked about this multiple times over the past couple of years. I proposed a solution to the timeout but am not sure where it went, but I think it probably just died.

From what I read this is the root of the problem. If the timeout didn't start immediately, then none of these problems occur.  The jobs would all be submitted, kubernetes would schedule them as resources permit, and things would just work.

To work around that, the proposal seems to implement a new scheduler. Why? This is not a problem I think we should be trying to solve.  Getting these things right take time, like years, and lots of experience. You can already see the questions here about trying to handle cases where resources aren't available. It's only going to get more complicated. Adding more complexity to an already complex system doesn't seem like a good idea.

To me this is a classic example of re-inventing the wheel.

jhutar:
Hello @ggallen. I do not think "PipelineRun timeout" is broken and that we are reinventing something although I'm not very knowleadgable on what Kubernetes scheduler can do. Also we do not have years to make it right.

IMO use-cases here are:

1) Only allow, say, 500 PipelineRuns to run in parallel, keep rest of them in a queue - this is scenario where we are protecting ETCD based on what we know it is able to handle without dying
2) Only start MPC PipelineRun if MPC pools required by PLR have at least 1 free seat - this is to avoid cases when PLR timeouts because some its task is waiting for resources that are not available for a long time, probably because other PLRs are using it

As of now, Gal is also proposing another solution that is quite similar to this one.

This is definitely not the only solution here.

You mentioned you had some proposal - could you link it here? Not sure if ADR is needed, I'm just +-out-sider here :-)

vdemeester:
> The timeout used by Tekton starts when the pipeline is submitted, not when it actually starts.
> 
> We've talked about this multiple times over the past couple of years. I proposed a solution to the timeout but am not sure where it went, but I think it probably just died.
> 
> From what I read this is the root of the problem. If the timeout didn't start immediately, then none of these problems occur. The jobs would all be submitted, kubernetes would schedule them as resources permit, and things would just work.

But this is not the root of the problem. In the case of MPC for example, the "time it waits" is in the middle of the PipelineRun. Essentially, the Pipeline starts just fine, does its thing, and at some point, a Task waits for something to be available (and by the Task, I mean the process running inside the Task). In *that case*, the timeouts are just doing their job. Tekton *doesn't* know what the task does (does it wait, does it do something) and shouldn't care even (just like the pod controller doesn't care why the Pod crashes for example).

The proposal is relatively simple and based on an existing feature of tektoncd/pipeline to delegate "when" should a PipelineRun start. From the [*very small* TEP](https://github.com/tektoncd/community/blob/main/teps/0015-pending-pipeline.md#summary) :

> […]
> 
> **The primary motivation for this is to support operators who need to control the
start up of `PipelineRuns` in clusters that are under heavy load.** By allowing
`PipelineRuns` to be created without actually starting them, **one could implement
a custom controller to implement any `PipelineRun` scheduling policy that is
transparent to the `PipelineRun` controller and `PipelineRun` creator (e.g.,
Tekton triggers)**.
> 
> […]

One example could have been "if you need to pro-provision something prior to execution" (*so that it doesn't timeout while executing, waiting for provisionning*), this could be used. It's not about writing a generic or even highly configurable scheduler. It's about implementing a very small, very focused behavior for a very specific case (or set of cases but it shouldn't be too big).

> To me this is a classic example of re-inventing the wheel.

Sometimes you need different wheels, so *reinventing* is not necessarily that bad (especially as here, it's not really reinventing but…). If we go that way, one could do a CI with just `Pod` and `Job` types, Tekton is just reinventing the wheel, isn't it 😝 😜  ?

ADR/0047-creating-pending-pipelineruns.md
```diff
@@ -0,0 +1,71 @@
...
+
+So say you started 300 pipeline runs on Friday evening and you expect to see 300 new images by Monday morning.
+You have resource quota for 20 concurrent builds, so 20 builds get started, when they finish 20 more and so on.
+But because all 300 pipeline runs started (from Tekton point of view) when they were created, after few iterations all the remaining pipeline runs will be cancelled (due to a timeout).
+
+When many pipelines are triggered across multiple namespaces, there is also an increased load on etcd which may degrade the cluster's overall performance.
+
+## Decision
+
+We will use idea of creating pipeline runs in pending state and having a standalone controller that would implement any custom logic on when to start each pending pipeline:
```
arewm, line 21: @jhutar , do you want to update this to use https://kueue.sigs.k8s.io/docs/overview/ instead of a custom controller?

ggallen:
> @ggallen But this is not the root of the problem. In the case of MPC for example, the "time it waits" is in the middle of the PipelineRun.

It's unclear to me how this ADR addresses this issue.  Does it mean to hold the pipeline until resources are available for each task in the pipeline?  Even if it does that, is it good enough?  Because by the time the pipeline gets to the task in question those resource may not now be available. Unless the proposal here is to actually some reserve all the resources required for all the pipeline runs.

Which also brings me to this question - how does any of this work with resource quotas?  Will it respect namespace quotas? For example, in the context section I see this:

> You have resource quota for 20 concurrent builds

But in the decision section I see this:

> Controller will keep an eye on current number of concurrently running pipeline runs in the cluster (using Prometheus metric `avg(tekton_pipelines_controller_running_pipelineruns)`), and if that is below given threshold, it starts the pipeline run(s).

I think what could happen here is that I submit my jobs and non get run because of the cluster threshold. Meanwhile, I look at my namespace and quota and see nothing running and can't figure out why.  I think that would be confusing to users.

Also, why use prometheus?  Why not just use a cluster quota and check that?

Here's an off-the-cuff idea.  Could this controller just watch pipeline runs, and if it finds one that's waiting for resources could it go and change the status to `Pending`?  Then it can periodically re-set the status and start the job.  It's essentially a retry, but it still leaves all the scheduling issues to Kubernetes.

@vdemeester, would Tekton work with that?

vdemeester:
> Does it mean to hold the pipeline until resources are available for each task in the pipeline? Even if it does that, is it good enough? Because by the time the pipeline gets to the task in question those resource may not now be available.

Well, that's *another story*, but the assumption would be that the resource is then dedicated to the PipelineRun somehow.

> Here's an off-the-cuff idea. Could this controller just watch pipeline runs, and if it finds one that's waiting for resources could it go and change the status to Pending? Then it can periodically re-set the status and start the job. It's essentially a retry, but it still leaves all the scheduling issues to Kubernetes.
> 
> @vdemeester, would Tekton work with that?

I don't think a PipelineRun can be set to `Pending` once it is started – maybe something to explore but it might have weird implications. We could think of having each `TaskRun` from a PipelineRun start as Pending (with an mutating admission controller ?) but it would not fix the timeout as, the PipelineRun timeout would still be counting down.

jhutar:
Hello!

> > @ggallen But this is not the root of the problem. In the case of MPC for example, the "time it waits" is in the middle of the PipelineRun.
> 
> It's unclear to me how this ADR addresses this issue. Does it mean to hold the pipeline until resources are available for each task in the pipeline?

The ADR aims (in the MPC use-case) to only start the PipelineRun once MPC have resources for all or PLR builds. That way build should not be stuck.

> Even if it does that, is it good enough?

No idea here :-/, but IMHO we will not know if we do not try.

> Because by the time the pipeline gets to the task in question those resource may not now be available. Unless the proposal here is to actually some reserve all the resources required for all the pipeline runs.

Part of the proposal (and of the demo code IIRC) was to "reserve" these MPC resources for some time after the PLR was started. Simple and stupid (but quick) solution would be to reserve resources for say 5min, more complex solution would be to track PLR execution ond reserve it until these tasks finish or fail.

> Which also brings me to this question - how does any of this work with resource quotas? Will it respect namespace quotas? For example, in the context section I see this:
> 
> > You have resource quota for 20 concurrent builds
> 
> But in the decision section I see this:
> 
> > Controller will keep an eye on current number of concurrently running pipeline runs in the cluster (using Prometheus metric `avg(tekton_pipelines_controller_running_pipelineruns)`), and if that is below given threshold, it starts the pipeline run(s).
> 
> I think what could happen here is that I submit my jobs and none get run because of the cluster threshold.

You are right. It might be possible MPC resources would be available, but namespace quota would be blocking the PLR execution. Given we would have a control over gating criteria, this can be added into equation as needed.

> Meanwhile, I look at my namespace and quota and see nothing running and can't figure out why. I think that would be confusing to users.

This proposal will not make current situation around "mine PLR is not started because of quota" worse as it is now (what you just described). Given the gating criteria can me extended, it can actually make it clearer if we standardize on some PLR annotation where the controller can be pushing the explanation.

> Also, why use prometheus? Why not just use a cluster quota and check that?

IMHO MPC machine pool status is not available there (we need to know something like "for arm64 MPC pool 18 out of 20 is currently occupied". If it is possible in some easier way than calling Prometheus, then sure, let's use it!

ggallen:
The discussion that is going on here has not alleviated any of my concerns. In fact it's made them worse.

What is being proposed here is much more than "creating pending pipeline runs", it's the implementation of a complete scheduling system that also allows for the reservation of resources.

I really don't think we want to go there.

ADR/0047-creating-pending-pipelineruns.md
arewm: I talked with @gbenhaim today briefly. A controller may still be used even if integrating with kueue.

PR #223, opened 2025-02-06
Update stale references to spec.containerImage

Josh-Everett:
Updated references to Spec.ContainerImage within integration service docs to reflect change to use the Status.LastPromotedImage field instead.


jsztuka (approved):
lgtm

ralphbean (approved):

PR #229, opened 2025-04-07
ADR 47: Support build-time security scans in integration service

14rcole:


jsztuka (approved):
lgtm

dirgim (approved):
LGTM

ADR/0047-movable-build-tests.md
```diff
@@ -0,0 +1,34 @@
+# 47. Attestable Build-Time Tests in Integration Service
```
arewm, line 1: 47 is taken by #218.

ADR/0047-movable-build-tests.md
```diff
@@ -0,0 +1,34 @@
...
+they be able to run security scans during their integration tests.
+
+## Decision
+
+We will add support for adding attestations to images from their integration
+pipelines.
+
+### Attestations
+
+It is already possible to generate attestations from an integration pipeline. Users can add `IMAGE_URL` and `IMAGE_DIGEST` result fields to the pipeline definition. When the pipeline runs and completes, Tekton Chains will sign an attestation and push it to the matching image/digest in Quay.  We will need to document this process.
```
arewm, line 23: I think that we need to address the fact that this attestation will be a [SLSA provenance predicate](https://github.com/in-toto/attestation/blob/main/spec/predicates/provenance.md). This content isn't _necessarily_ provenance, but it hasn't _really_ been provenance all along. We will continue to use the provenance predicate as generated by Chains. If, in the future, Chains starts supporting additional predicates, we would plan to change.

We do not want tasks to generate their own attestations because we still want to allow policies to require certain tasks to run and for these tasks to come from trusted references. We therefore need to have the attestation generated for the entire PipelineRun.

This is still consistent with the pattern that we have been following as these results have always been in a SLSA provenance predicate.

ADR/0047-movable-build-tests.md
```diff
@@ -0,0 +1,34 @@
...
+## Status
+
+Proposed
+
+## Context
+
+Some of the security scans run during the build pipeline take a long time to
+run. This forces users to wait longer to be notified of the results of their
+builds.  In order to receive more immediate feedback, users have requested that
+they be able to run security scans during their integration tests.
```
arewm, line 14: While not necessarily incorrect, I feel like we can be more clear here.

```suggestion
Konflux makes it easy to run many security checks immediately after builds. While
the tests are shifted left (good), the checks are serialized to run immediately after the
builds which increase the feedback time (bad). Konflux has the ability to trigger additional
PipelineRuns immediately after the completion of artifact builds via IntegrationTestScenarios.

This architecture additionally improves the user experience as it allows these checks to be rerun
without having to rebuild the artifacts themselves, further decreasing feedback time upon failed
post-build TaskRuns.
```

ADR/0047-movable-build-tests.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Decision
+
+We will add support for adding attestations to images from their integration
+pipelines.
```
arewm, line 19: Moving some of the content from earlier to the decision.

```suggestion
In order to get scan-related feedback immediately after builds complete, we will add
support for creating enforceable test results from Integration pipelines. By leveraging
IntegrationTestScenarios, Konflux users can separate their build tasks from their testing
tasks. The test pipelines can be rerun independently of the build pipelines but all pipelines
will have attestations generated for them to enable Conforma evaluations.
```

ADR/0047-movable-build-tests.md
```diff
@@ -0,0 +1,34 @@
...
+
+### Attestations
+
+It is already possible to generate attestations from an integration pipeline. Users can add `IMAGE_URL` and `IMAGE_DIGEST` result fields to the pipeline definition. When the pipeline runs and completes, Tekton Chains will sign an attestation and push it to the matching image/digest in Quay.  We will need to document this process.
+
+### Serialization
+
+The integration service will also add support for serialization of tests so
+that the Conforma pipeline can run after any tests that generate attestations.
+The exact mechanism for this has yet to be determined.
```
ralphbean, line 29: In the time that passed since writing this, is the mechanism more clear now? Can we write anything about what it is yet? Or, still TBD?

ADR/0047-movable-build-tests.md
ralphbean: I think Ryan's original is more clear. It states the context that brings us to the decision, without yet stating the decision or its consequences.

ADR/0047-movable-build-tests.md
```diff
@@ -0,0 +1,34 @@
...
+## Context
+
+Some of the security scans run during the build pipeline take a long time to
+run. This forces users to wait longer to be notified of the results of their
+builds.  In order to receive more immediate feedback, users have requested that
+they be able to run security scans during their integration tests.
+
+## Decision
+
+We will add support for adding attestations to images from their integration
```
lcarva, line 18: One thing to keep in mind is that we need to ensure conforma will behave as expected when there are two (different) SLSA Provenance attestations attached to the same image. Conforma ought to handle this, but previous experiments have shown that it does not handle this gracefully. I believe one of the issues was that it wanted to find every required build task in each of the attestations.

ADR/0047-movable-build-tests.md
arewm: I agree after re-reading. I don't like what I proposed anymore. :)

ADR/0047-movable-build-tests.md
14rcole: Yes, this will be updated when I push the current commit.

ADR/0047-movable-build-tests.md
arewm: I think it would be good to record at least some of the "why" since reuse of the predicate is likely to be seen as an anti-pattern.

ADR/0047-movable-build-tests.md
14rcole: > We also need to ensure that the integration test attestation does not "overwrite" the build pipeline attestation by using the same tag

I'm confused as to how this might happen.  Can you give me an example? Also, do you mean the image tags (e.g. `latest`, `v1`, etc) or a tag specifically related to the attestations?

mmorhun (approved):

ADR/0048-movable-build-tests.md
```diff
@@ -0,0 +1,68 @@
...
+required tasks will be part of this group unless there is a reason that they
+cannot or should not be run in integration pipelines.
+
+### Serialization
+
+The integration service will also add support for serialization of tests so
+that the Conforma pipeline can run after any tests that generate attestations.
+Integration tests will run as a directed acyclic graph.
+
+#### Integration Test Scenario DAG rules
```
adambkaplan, line 57: Another vote for contributing to Tekton [Pipelines in Pipelines](https://tekton.dev/docs/pipelines/pipelines-in-pipelines/), an alpha feature that is missing most of its implementation.

ADR/0047-movable-build-tests.md
arewm: After talking with the KGC today, we want to see the change reflected in the ADR. As we are trying to position Konflux in the community as an example of SLSA use and supply chain security, we should be more careful with how we use provenance. Generating provenance for an image that isn't produced in that PipelineRun is much less ideal than each task producing its own attestations and Chains generating provenance for those attestations. This approach is also consistent with our current trust model that leverages provenance as a critical step in the verification process.

14rcole: I've updated the ADR to reflect your comments, let me know if it looks good

ADR/0048-movable-build-tests.md
```diff
@@ -0,0 +1,68 @@
...
+pipelines but all pipelines will have attestations generated for them to enable
+Conforma evaluations.
+
+### Attestations
+
+Each task will be responsible for generating an attestation in Cosign. `vuln`
+attestations will be used for security scans rather than `provenance`
+attestations. Tasks will output `ARTIFACT_URI` and `ARTIFACT_DIGEST` which will
+point to the location of the attestation in Quay.  These results will tell
+Chains to sign the attestation.
```
ralphbean, line 29: Just want to be clear that chains won't sign the attestation directly generated by the task.

```suggestion
point to the location of the attestation in Quay.  These results will tell
Chains to generate and sign a SLSA provenance attestation for the
task-generated attestation.
```

ADR/0048-movable-build-tests.md
```diff
@@ -0,0 +1,68 @@
...
+## Decision
+
+We will add support for adding attestations to images from their integration
+pipelines. Test pipelines can be run and re-run independently of build
+pipelines but all pipelines will have attestations generated for them to enable
+Conforma evaluations.
+
+### Attestations
+
+Each task will be responsible for generating an attestation in Cosign. `vuln`
```
arewm, line 25: Are you trying to say https://github.com/in-toto/attestation/blob/main/spec/predicates/vuln.md or https://github.com/in-toto/attestation/blob/main/spec/predicates/vulns_02.md?

ADR/0048-movable-build-tests.md
```diff
@@ -0,0 +1,68 @@
...
+attestations will be used for security scans rather than `provenance`
+attestations. Tasks will output `ARTIFACT_URI` and `ARTIFACT_DIGEST` which will
```
arewm, line 27: Can we just use `*ARTIFACT_OUTPUTS`? https://tekton.dev/docs/chains/slsa-provenance/#artifact_outputs

Also, we can have specific types of result names if we want, i.e. `VULN_ARTIFACT_OUTPUTS` or `TEST_OUTPUT_ARTIFACTS`.

ADR/0048-movable-build-tests.md
arewm: We might also want other types of attestations, i.e. https://github.com/in-toto/attestation/blob/main/spec/predicates/test-result.md

ADR/0048-movable-build-tests.md
```diff
@@ -0,0 +1,68 @@
...
+- A failed or cancelled test will result in all child tests being cancelled
+- If a test scenario is re-run and passes then child scenarios will also be
+triggered
+
+## Consequences
+
+By leveraging IntegrationTestScenarios, Konflux users can separate their build
+tasks from their testing tasks. This will enable faster turnaround time between
+triggering a build and getting the result.  In some cases the build-test
+process will take less time overall.
```
ralphbean, line 68:
```suggestion
process will take less time overall.

It is worth noting that right now, almost all of the scan tasks run in parallel in the build pipeline.
If we move some of those to be IntegrationTestScenarios but not all of them, then the end to
end will take even longer than it does today. We will have taken one set of parallel things and
turned it into two sets of parallel things, run in sequence.
```

ralphbean:
@konflux-ci/kgc is pretty positive about this change. We're going to bump it out one more week to give @14rcole an opportunity to respond to the latest comments and questions.

ADR/0048-movable-build-tests.md
14rcole: 0.2, I'll update the document with a link.

ADR/0048-movable-build-tests.md
14rcole: We can, but I'm not sure what it gets us if Chains can watch TaskRuns since each TaskRun would only generate one artifact anyway. If you think its better to have `*_ARTIFACT_OUTPUTS` outputted by the PipelineRun that's a different story.

ADR/0048-movable-build-tests.md
14rcole: I'll change that, but doesn't Chains generate a [vulnerability attestation](https://github.com/in-toto/attestation/blob/main/spec/predicates/vulns_02.md)? Or does that attestation type only apply to the task-generated attestation and the Chains-generated attestation is always a SLSA provenance attestation?

ADR/0048-movable-build-tests.md
arewm: If we use `*_ARTIFACT_OUTPUTS` then it is a clearer path if a task wants to produce multiple attestations.

PR #239, opened 2025-07-02
ADR-0049: VSA Attestations for Release Policies

adambkaplan:
Proposal to produce SLSA Verification Summary Attestations (VSAs) as output from Enterprise Contract IntegrationTestScenarios, and to use these in release pipeline Conforma policy checks.

### DIFF
--- a/ADR/0049-vsa-support.md
+++ b/ADR/0049-vsa-support.md
@@ -1,0 +1,57 @@
+# ADR-0049: Verification Summary Attestations for Release Policies
+
+Date: 2025-07-02
+
+## Status
+
+Proposed
+
+## Context
+
+We are facing several challenges with our current approach to Enterprise Contract (EC) verification
+and SDL checks:
+
+1. **Release-time EC verification performance**: Running EC at release time requires checking a
+   large number of attestations, which takes too long. Scaling to handle large applications like
+   OpenShift is proving challenging.
+2. **Pipeline arrangement limitations**: Currently we have to run all SDL checks in the build
+   pipeline, which limits our options on how to arrange pipelines and prevents us from potentially
+   moving some checks into integration tests.
+3. **Redundant verification of unchanged artifacts**: Pre-merge EC checks for large snapshots often
+   include one freshly built image plus a number of images that haven't changed, yet we redo the EC
+   check on all images.
+4. **Scalability concerns**: As our applications grow in complexity (like OpenShift snapshots), the
+   current approach becomes increasingly inefficient.
+5. **Release externally-built artifacts**: Our current policies for release pipelines assume
+   artifacts were built on Konflux. This inhibits ingestion and release of artifacts from other
+   external (but trusted) systems.
+
+## Decision
+
+We will adopt Verification Summary Attestations ([VSAs](https://slsa.dev/spec/v1.1/verification_summary))
+for recording SDL policy check results. These VSAs will be used as the input to release pipeline
+Conforma checks.
+
+## Consequences
+
+- Conforma policy [packages](https://conforma.dev/docs/policy/authoring.html#_package_annotations)
+  provide metadata that associates a set of rules with a VSA [verification level](https://slsa.dev/spec/v1.1/verification_summary#fields).
+  - Rules that validate SLSA build levels MUST indicate the appropriate [SlsaResult](https://slsa.dev/spec/v1.1/verification_summary#emslsaresult-stringem)
+    value.
+  - Rules that are not related to SLSA (or have not been incorporated into SLSA yet) MUST use a
+    value that does not have the `SLSA_*` prefix. For example, rulesets that verify an artifact
+    provides source in accordance with GPL v3 could use `GNU_GPL_3` as its "verification level."
+- Conforma's CLI produces output in VSA format:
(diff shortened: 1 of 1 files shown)


ADR/0049-vsa-support.md
```diff
@@ -0,0 +1,69 @@
...
+2. **Pipeline arrangement limitations**: Currently we have to run all SDL checks in the build
+   pipeline, which limits our options on how to arrange pipelines and prevents us from potentially
+   moving some checks into integration tests.
+3. **Redundant verification of unchanged artifacts**: Pre-merge EC checks for large snapshots often
+   include one freshly built image plus a number of images that haven't changed, yet we redo the EC
+   check on all images.
+4. **Scalability concerns**: As our applications grow in complexity (like OpenShift snapshots), the
+   current approach becomes increasingly inefficient.
+5. **Release externally-built artifacts**: Our current policies for release pipelines assume
+   artifacts were built on Konflux. This inhibits ingestion and release of artifacts from other
```
gbenhaim, line 26: 1. What is to use case for releasing artifacts built by other systems? I believe additional work would be needed to support this.

2. What is the meaning of "artifact built by konflux"? One can use the konflux tasks (without running a full konflux instance) and the policies should accept the build.

ADR/0049-vsa-support.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Context
+
+We are facing several challenges with our current approach to Enterprise Contract (EC) verification
+and SDL checks:
+
+1. **Release-time EC verification performance**: Running EC at release time requires checking a
+   large number of attestations, which takes too long. Scaling to handle large applications like
+   OpenShift is proving challenging.
+2. **Pipeline arrangement limitations**: Currently we have to run all SDL checks in the build
```
gbenhaim, line 17: I didn't see in https://github.com/konflux-ci/architecture/pull/229/files a requirement to use the suggested format. Is it a requirement for moving SDL checks to the integration tests?

ralphbean:
Noting here that the specifics of this decision might change as a result of https://github.com/in-toto/attestation/pull/470.

ADR/0049-vsa-support.md
adambkaplan: No. I raised a question wrt to what exactly is being attested in ADR 48, and how the scan results can be obtained. https://github.com/konflux-ci/architecture/pull/229#discussion_r2353555357

ADR/0049-vsa-support.md
```diff
@@ -0,0 +1,69 @@
+# ADR-0049: Verification Summary Attestations for Release Policies
+
+Date: 2025-07-02
+
+## Status
+
+Proposed
+
```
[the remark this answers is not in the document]
adambkaplan: Updated!

ADR/0049-vsa-support.md
```diff
@@ -0,0 +1,81 @@
...
+## Context
+
+We are facing several challenges with our current approach to Enterprise Contract (EC) verification
+and SDL checks:
```
ifireball, line 12: SDL?

ADR/0049-vsa-support.md
```diff
@@ -0,0 +1,81 @@
...
+5. **Release externally-built artifacts**: Our current policies for release pipelines assume
+   artifacts were built using a sanctioned Konflux Tekton pipeline. This inhibits ingestion and
+   release of artifacts from external (but trusted) systems that do not execute builds with Tekton.
+   Ex: and artifact produced from a trusted Jenkins instance, or Java artifacts built using
+   [Project Newcastle](https://github.com/project-ncl).
+
+## Decision
+
+We will adopt Verification Summary Attestations ([VSAs](https://slsa.dev/spec/v1.1/verification_summary))
+for recording SDL policy check results. These VSAs will be used as the input to release pipeline
```
ifireball, line 35: Please explain a bit how and why VSAs will help, and introduce the term so people can uderstand the ADR without going to the link.

ADR/0049-vsa-support.md
adambkaplan: Software development lifecycle. Fixed.

ADR/0049-vsa-support.md
```diff
@@ -0,0 +1,89 @@
...
+- Allow Conforma to verify attestations by checking for the presence/absence of a policy check in
+  the summary attestation. This should perform faster than a deeper inspection of the attestation
+  and related artifacts.
+- Allow Conforma to verify attestations in a general manner from a non-Konflux (or Tekton) system.
+
+## Decision
+
+We will adopt SLSA Verification Summary Attestations ([VSAs](https://slsa.dev/spec/v1.1/verification_summary))
+for recording SDL policy check results. These VSAs will be used as the input to release pipeline
+Conforma checks.
```
ifireball, line 44: What will generate the VSAs? What provides provenance guarantees for them?

adambkaplan: It is up to the implementer to generate the VSA, and IMO it makes sense for Konflux to require that the VSA be signed in order for it to be trusted.

Within Konflux itself - Conforma can produce a VSA as its output (currently "alpha" feature). We can package this in an OCI artifact, push to a registry, and have Chains provide the signature + attestation.

PR #240, opened 2025-07-09
Add ADR-0047: GitOps Onboarding Redesign

brianwcook:
This ADR proposes replacing Konflux's current dual onboarding flows (UI + GitOps) with a unified Git-centric approach via VS Code plugin.

## Key Changes
- Single onboarding flow through VS Code plugin with local validation
- Multi-repo GitOps model for team autonomy  
- Fleet awareness for enterprise multi-instance deployments
- Declarative secrets management (Kubernetes + Vault backends)
- UI preserved for monitoring only

## Impact
- Eliminates maintenance overhead of dual flows
- Improves developer experience with familiar IDE tooling
- Enables enterprise-scale fleet management

This is a breaking change - UI onboarding will be deprecated with no fallback.


ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,220 @@
...
+
+### GitOps Onboarding
+- Requires manually authoring YAML across multiple files
+- Validation happens during GitLab CI runs, creating tension between Konflux and GitLab CI workflows
+- Secrets must still be created via UI, breaking the GitOps flow
+- Onboarding is slow and often requires multiple failed PRs to converge due to trial-and-error approach
+
+### Current Pain Points
+1. **Fragmented experience**: Developers must context-switch between UI and GitOps approaches
+2. **Validation delays**: Schema validation only occurs during CI runs, not locally
```
ralphbean, line 26: This isn't true of the existing RH gitops repo. Run `tox` (or write a VSCode plugin to run `tox`).

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,220 @@
...
+
+**Improving the existing UI** was considered but rejected because:
+- Despite continuous UI improvements, many users still choose GitOps onboarding for its inherent disaster recovery and change management benefits
+- UI enhancements don't address the fundamental issue of maintaining two separate onboarding flows
+- Users who prefer GitOps principles cannot benefit from UI improvements, creating a permanent split in the user experience
+
+**Chosen approach** replaces both current flows with a unified Git-centric solution that preserves GitOps benefits while providing the usability improvements users expect.
+
+### Core Shift
+- **Replace UI configuration** with a VS Code plugin for onboarding and editing
```
ralphbean, line 46: Is this meant to be _in place of_ a CI flow for validation?

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,220 @@
...
+- **Use local schema validation**, linting, and previews to reduce onboarding errors
+- **Use Git as the single source of truth** for applications, components and secrets
```
ralphbean, line 48: What about integration test scenarios and releaseplans?

Perhaps just say "tenant resources".

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,220 @@
...
+- Linting and validation against Konflux schemas
+- Integration with local secret store session (Vault or Kubernetes)
+- Git-aware diff view and PR-ready commit generation
+- Optional: preview pipeline graph / dependency resolution
+
+**Secrets UX:**
+- Form-based secret creation (with validation, TTLs, metadata)
+- Generated secret definitions point to secret store paths via ESO
+- Secrets created in the secret store configured for that instance of Konflux
+- Initially supported stores: Vault and Kubernetes Secrets
```
ralphbean, line 85: What is meant by "Kubernetes Secrets" here? Do you mean to use ESO without a SecretStore?

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,220 @@
...
+- Optional: audit log stored in `.konflux/` for local tracking
+
+**GitOps Flow:**
+1. Developer opens repo in VS Code
+2. Launches Konflux plugin
+3. Uses wizard to configure new component/build
+4. Adds secrets as needed via plugin form
+5. Plugin generates YAML and commits to Git
+6. Developer pushes PR
```
ralphbean, line 94: To where?

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,220 @@
...
+7. CI runs lightweight schema validation (should pass if plugin was used)
```
ralphbean, line 95: Ah, good - CI isn't going away.

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,220 @@
...
+
+### Multi-Repo GitOps Model
+
+**Tenant Onboarding via Repo Registration:**
+- Konflux will support multiple GitOps repositories, each representing an independent tenant
+- Each registered GitOps repo is treated as a distinct tenant with its own Kubernetes namespace
+- Teams register GitOps repo via VS Code plugin, CLI (`konflux register`), or UI
+- Konflux provisions a new Kubernetes namespace and runs initial validation
```
ralphbean, line 103: Missing detail here. When users issues `konflux register`, what API do they hit?

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,220 @@
...
+
+### Secrets Management Integration
+
+**Core Capability:**
+- VS Code plugin enables declarative secrets creation directly from the IDE
+- Secrets are managed through Git configuration with secure backend integration
+- Generated secret definitions use External Secrets Operator (ESO) to reference backend stores
+
+**Supported Backends:**
+- **Kubernetes Secrets**: Native Kubernetes secret storage
```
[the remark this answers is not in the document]
ralphbean: Is "kubernetes secrets" a backend store for ESO?

ESO has "providers". Here's a list: https://external-secrets.io/latest/provider/aws-secrets-manager/

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,220 @@
...
+
+**Security Principles:**
+- Plugin only creates secrets, does not read or list existing secrets
+- All secret operations are scoped and audited
+- Optional: CLI fallback (`konflux secrets create`)
+
+### Fleet Awareness
+
+**Multi-Instance Discovery:**
+When a user's organization operates multiple Konflux instances, the VS Code plugin should provide fleet-wide awareness and intelligent instance selection. Fleet metadata is simple metadata with a description of the clusters capabilities and limitations (text), plus an indicator if the cluster has capacity for new tenants namespace. Each Konflux instance should have a complete copy of the fleet metadata.
```
ralphbean, line 126: This seems like it is its own idea. Break this out into a separate ADR to flesh it out.

What's the metadata look like? How does a cluster know if its full? How does each instance get a copy of the fleet metadata? How do instances join the fleet? What defines the fleet? If two instances have different copies of the fleet metadata, how does that get reconciled? Can you get the fleet metadata without otherwise being authorized on a cluster? If not how do you get authorized in the first place?

ralphbean: Besides - this feels like a _generic_ problem. What other fleet management technologies should we look at before building anything new here?

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,220 @@
...
+- Same validator embedded in VS Code plugin
+- Optional: Plugin lints against live cluster (e.g., available build agents)
+
+### UI Changes
+
+**Preserve:**
+- Dashboard
+- Build and release logs
+- Live environment status
+- Observability / metrics
```
ralphbean, line 162: Are there metrics in the UI?

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,220 @@
...
+- Validates schema, references, build logic statically
+- Same validator embedded in VS Code plugin
+- Optional: Plugin lints against live cluster (e.g., available build agents)
+
+### UI Changes
+
+**Preserve:**
+- Dashboard
+- Build and release logs
+- Live environment status
```
ralphbean, line 161: We deprecated the Environment API a long time ago.

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,220 @@
...
+
+**GitOps Flow:**
+1. Developer opens repo in VS Code
+2. Launches Konflux plugin
+3. Uses wizard to configure new component/build
+4. Adds secrets as needed via plugin form
+5. Plugin generates YAML and commits to Git
+6. Developer pushes PR
+7. CI runs lightweight schema validation (should pass if plugin was used)
+
```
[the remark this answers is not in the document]
cit1zen: I believe that for this proposal to be effective, there has to be a service that reconciles these changes running withing the Konflux.

Saying that Konflux is a GItOps-only/first system and then not providing any build-in GitOps solution would lead to poor and bizarre experience.

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,220 @@
...
+### Multi-Repo GitOps Model
+
+**Tenant Onboarding via Repo Registration:**
+- Konflux will support multiple GitOps repositories, each representing an independent tenant
```
cit1zen, line 100: What if your app consists of multiple components in multiple repositories?

ADR/0047-gitops-onboarding-redesign.md
cit1zen: I second that fleet management should be its own ADR. I would even go as far and question if we need any idea of a fleet on the cluster level.

ADR/0047-gitops-onboarding-redesign.md
brianwcook: I have done that, and it resulted in merge conflict hell - I think its been improved since, but having a monorepo makes merge conflicts much more likely than distributed individual repos

ADR/0047-gitops-onboarding-redesign.md
brianwcook: no, but its meant to ensure that commits will pass CI in all but the rarest of circumstances

ADR/0047-gitops-onboarding-redesign.md
brianwcook: It probably also makes sense to move that to the gitops based approach.

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,220 @@
...
+3. **Error-prone process**: Manual YAML authoring leads to frequent failures
+4. **Secrets inconsistency**: Secrets cannot be managed declaratively through Git
+5. **Monorepo complexity**: Current GitOps monorepo for tenant configuration creates CODEOWNERS management overhead
+6. **UI improvement limitations**: Despite continuous improvements to the UI, many users still choose GitOps onboarding for its inherent disaster recovery and change management benefits, leaving them unable to benefit from UI enhancements
+
+## Decision
+
+Redesign the onboarding experience to converge UI and GitOps into a single coherent, Git-centric flow through the following architectural changes:
+
+### Alternative Approaches Considered
```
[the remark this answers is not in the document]
brianwcook: done

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,220 @@
...
+
+| Component | Role |
+|-----------|------|
+| VS Code Plugin | Main interface for onboarding and configuration |
+| Git Repos | Source of truth for Konflux config |
+| Vault | Preferred secrets backend |
+| Kubernetes Secrets | Alternative backend for secrets |
+| External Secrets Operator (ESO) | Delivery of secrets to Kubernetes |
+| Konflux UI | Monitoring and runtime insight only |
+| Fleet Registry | Service discovery and instance metadata for multi-instance deployments |
```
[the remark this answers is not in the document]
brianwcook: agree, I removed it

ADR/0047-gitops-onboarding-redesign.md
brianwcook: to their own gitops repo that is registered with soe Konflux instance.

ADR/0047-gitops-onboarding-redesign.md
brianwcook: we can give them the tools for CI but we can't enforce it as in my design, the user owns their own repo.

ADR/0047-gitops-onboarding-redesign.md
brianwcook: I hope there will be. Things like DORA metrics, current queue depth, average queue times, etc

ADR/0047-gitops-onboarding-redesign.md
brianwcook: removed

ADR/0047-gitops-onboarding-redesign.md
brianwcook: that isn't a problem as long as you are building them in a single namespace. if you need to configure N namespaces, you need to bring N gitops repos.

brianwcook:
@cit1zen @ralphbean : Thanks for your feedback. I have slimmed up the scope of this ADR to only the most important concepts, addressed many of your comments and tried to improve overall clarity. When you have time, take another look.

ADR/0047-gitops-onboarding-redesign.md
ralphbean: Update the pain points to reflect that. That's not what the bullet says.

ADR/0047-gitops-onboarding-redesign.md
ralphbean: Where/when does that come in? This is the first time it shows up in the numbered list.

ADR/0047-gitops-onboarding-redesign.md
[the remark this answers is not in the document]
ralphbean: Architecturally, is that argo instance located on the shard, on the cluster? (like we do in the fedora instance?)

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,229 @@
...
+**Improving the existing UI** was considered but rejected because:
+- Despite continuous UI improvements, many users still choose GitOps onboarding for its inherent disaster recovery and change management benefits
+- UI enhancements don't address the fundamental issue of maintaining two separate onboarding flows
+- Users who prefer GitOps principles cannot benefit from UI improvements, creating a permanent split in the user experience
+
+**Chosen approach** replaces both current flows with a unified Git-centric solution that preserves GitOps benefits while providing the usability improvements users expect.
+
+### Core Shift
+- **Replace UI configuration** with a VS Code plugin for onboarding and editing
+- **Use local schema validation**, linting, and previews to reduce onboarding errors
```
ralphbean, line 46: This ADR needs a security considerations sections:

* Nothing can ensure that any tenant gitops repo is not full of malicious manifests. Argo configuration needs to be strictly controlled to limit what can be applied to the cluster, and where, from which gitops repo. Brainstorm some specifically bad scenarios in the consequences section - get creative - and discuss how they're mitigated or not.
* Can users see the argo UI? If they can, can they see other tenants' resources? In the monorepo, this is moot because all the resources are there anyways. In the multi-repo model, now it's not clear they should have view rights cross-tenant in the control plane system.
* Can users register a private gitops repo with argo? How does that flow work?

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,229 @@
...
+
+### Multi-Repo GitOps Model & Tenant Registration Service
+
+**Tenant Registration Service:**
+Konflux will introduce a new **Tenant Registration Service** API that manages the registration and lifecycle of GitOps repositories as first-class tenants.
+
+**Tenant Model:**
+- **Tenant Definition**: A tenant is equivalent to a Kubernetes namespace
+- **1:1 Mapping**: Each GitOps repository maps to exactly one Kubernetes namespace
+- **Namespace Declaration**: The namespace name is declared within the GitOps repo itself and is immutable after registration
```
lcarva, line 99: Consider the namespace not being part of the GitOps repo. Since the name is immutable, it should reside somewhere where the user doesn't have access to alter it. We could make it so it's managed by some sort of immutable CR, but that seems overkill.

The registration API will already take as input the repo and the namespace name. Its implementation should claim that namespace so it can only be done once, rejecting if it is already claimed. Behind the scenes, that should create, among other things, something like an ArgoCD [Application](https://argo-cd.readthedocs.io/en/stable/operator-manual/declarative-setup/#applications). (And probably grant read access to the user for that Application).

If someone wants to register the same GitOps repo but for a different namespace, who cares? All the created resources are scoped to each namespace so there's no collision/conflict. Is that a naive approach?

ADR/0047-gitops-onboarding-redesign.md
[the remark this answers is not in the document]
lcarva: I think that will depend on how much overhead argo brings to syncing and monitoring each gitops repo.

Does it being, or not being, a core service change things? I'd imagine that if it's a core service, it should reside within each cluster.

ADR/0047-gitops-onboarding-redesign.md
brianwcook: yeah, this idea has evolved a little bit. I agree that the name being in the repo when it is immutable is wrong. In the mocks, it is part of the registration process. That seems right, and leads me to believe the namespace name should be a parameter to the registration API and _not_ in the gitops repo itself. I will make the update.

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,220 @@
...
+
+### Component Architecture
+
+| Component | Role |
+|-----------|------|
+| VS Code Plugin | Main interface for onboarding and configuration |
+| Git Repos | Source of truth for Konflux config |
+| Vault | Preferred secrets backend |
+| Kubernetes Secrets | Alternative backend for secrets |
+| External Secrets Operator (ESO) | Delivery of secrets to Kubernetes |
```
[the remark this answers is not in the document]
brianwcook: No, if you use kubernetes secrets, no gitops for you. The biggest downside of this IMO is that you lose the disaster revcovery properties of gitops - you can easily restore the gitops repos, but many things will likely be broken without the secrets. We should heavily encourage use of ESO but it is too heavy handed to _require_ it I think.

ADR/0047-gitops-onboarding-redesign.md
brianwcook: also i dropped everything related to secrets from the ADR - it seems like it is its own topic

ADR/0047-gitops-onboarding-redesign.md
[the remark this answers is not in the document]
brianwcook: if gitops is not a core service, then we really aren't introducing a gitops based approach. We're just deprecating the UI and telling people they have to configure Kubernetes themselves. Some might use gitops, some might use kubectl apply. If what we want is for people to use Gitops, then why would we not provide a Gitops instance?

gbenhaim:
I still need to go over the fine details of this ADR, but overall LGTM and I really like the approach.

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,220 @@
...
+5. **Monorepo complexity**: Current GitOps monorepo for tenant configuration creates CODEOWNERS management overhead
+6. **UI improvement limitations**: Despite continuous improvements to the UI, many users still choose GitOps onboarding for its inherent disaster recovery and change management benefits, leaving them unable to benefit from UI enhancements
+
+## Decision
+
+Redesign the onboarding experience to converge UI and GitOps into a single coherent, Git-centric flow through the following architectural changes:
+
+### Alternative Approaches Considered
+
+**Improving the existing UI** was considered but rejected because:
```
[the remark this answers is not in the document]
cit1zen: Yes, that was something what I was imagining. Not sure how it would look like, but you would do changes in some configurator and then hit commit and it would create or update a PR to your GitOps repository.

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,230 @@
...
+- Despite continuous UI improvements, many users still choose GitOps onboarding for its inherent disaster recovery and change management benefits
+- UI enhancements don't address the fundamental issue of maintaining two separate onboarding flows
+- Users who prefer GitOps principles cannot benefit from UI improvements, creating a permanent split in the user experience
+
+**Chosen approach** replaces both current flows with a unified Git-centric solution that preserves GitOps benefits while providing the usability improvements users expect.
+
+### Core Shift
+- **Replace UI configuration** with a VS Code plugin for onboarding and editing
+- **Use local schema validation**, linting, and previews to reduce onboarding errors
+- **Use Git as the single source of truth** for applications, components, integration test scenarios, releasePlans and RBAC
```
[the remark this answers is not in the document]
gbenhaim: We should use ArgoCD projects for limiting what tenants can deploy to their namespace (like we do today).

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,279 @@
...
+- **Without both services**: Konflux remains fully functional but requires manual GitOps repository setup and synchronization. This configuration provides:
+  - Smaller deployment footprint suitable for lower-spec machines
+  - Simplified setup for testing and development environments  
+  - Reduced operational complexity for teams preferring manual control
+  - Degraded user experience with manual setup steps
+  - Reduced disaster recovery capabilities due to lack of automated drift detection
+
+**Tenant Model:**
+- **Tenant Definition**: A tenant is equivalent to a Kubernetes namespace
+- **1:1 Mapping**: Each GitOps repository maps to exactly one Kubernetes namespace
```
gbenhaim, line 114: Why this limitation is needed?

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,279 @@
...
+5. Developer is prompted by UI to install / open repo in VS Code
+3. Launches Konflux plugin and uses it to configure new components, tests, releasePlans
+6. Plugin generates YAML objects in repo
+7. Developer commits and pushes code (optionally via PR, with CI,reviews etc.)
+9. ArgoCD automatically syncs merged changes to the namespace
+
+### Multi-Repo GitOps Model & Optional Services
+
+**GitOps Registration Service (Optional):**
+Konflux will optionally provide a **GitOps Registration Service** API that automates the registration and lifecycle of GitOps repositories as first-class tenants. This service is recommended for production deployments but not required for basic Konflux functionality.
```
gbenhaim, line 98: What is the meaning of registration without running argo as part of konflux?

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,279 @@
...
+**Tenant Model:**
+- **Tenant Definition**: A tenant is equivalent to a Kubernetes namespace
+- **1:1 Mapping**: Each GitOps repository maps to exactly one Kubernetes namespace
+- **Namespace Declaration**: The namespace name is a parameter to the registration service. Once the repo is registered, the namespace it is immutable.
+- **Registration Immutability**: Re-registering the same repo is rejected to prevent namespace conflicts
+
+**Resource Scope:**
+- **Multi-Resource Support**: One GitOps repo can contain multiple applications, components, integrationTestScenarios, and releasePlans
+- **Single Namespace Constraint**: All resources from a GitOps repo deploy to its designated namespace only
+- **No Cross-Namespace Interaction**: Changes affecting N namespaces require changes in N separate GitOps repos
```
gbenhaim, line 121: I don't think we need this limitation. I can imagine that some teams would want to manage multiple namespaces from the same gitops repo.

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,279 @@
...
+
+**Registration Process (when using optional services):**
+1. Team registers GitOps repo via VS Code plugin, CLI (`konflux register`), or UI
+2. at registration time the desired namespace name is provided to (UI | API)
+3. GitOps Registration Service validates the repo and the requested namespace
+4. Service provisions the new requested Kubernetes namespace
+6. Service configures ArgoCD to monitor the repo for continuous deployment
```
gbenhaim, line 128: I'm not sure an additional service is needed to configure argo. Why not creating argo resources directly?

ADR/0047-gitops-onboarding-redesign.md
brianwcook: both the registration server _and_ argo are optional, but the assumption is if you choose to run registration service, you would also run Argo.
The point is you don't need either one to run a minimal, functioning konflux

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,279 @@
...
+  - Simplified setup for testing and development environments  
+  - Reduced operational complexity for teams preferring manual control
+  - Degraded user experience with manual setup steps
+  - Reduced disaster recovery capabilities due to lack of automated drift detection
+
+**Tenant Model:**
+- **Tenant Definition**: A tenant is equivalent to a Kubernetes namespace
+- **1:1 Mapping**: Each GitOps repository maps to exactly one Kubernetes namespace
+- **Namespace Declaration**: The namespace name is a parameter to the registration service. Once the repo is registered, the namespace it is immutable.
+- **Registration Immutability**: Re-registering the same repo is rejected to prevent namespace conflicts
```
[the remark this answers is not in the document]
brianwcook: If i try to register my repo to as a new namespace "foo" but foo already exists, it will return 409 error.
If I try to register my repo to an _existing namespace_ (different API endpoint) it checks to make sure I am an admin in the target namespace.

ADR/0047-gitops-onboarding-redesign.md
brianwcook: I like it, its clean and easy to understand / enforce and we have long said that namespaces are the logical team boundaries for Konflux. It fits our model.

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,279 @@
...
+- Forms and wizards to scaffold and edit YAML objects
+- Built-in schema-aware YAML editor
+- Linting and validation against Konflux schemas
+- Git-aware diff view and PR-ready commit generation
+- Optional: preview pipeline graph / dependency resolution
+
+
+**GitOps Flow:**
+
+1. Developer creates an empty git repo with Github, Gitlab, etc.
```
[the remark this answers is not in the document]
brianwcook: I suppose both of these are possible but we should enhance the default PAC on-cel expressions to not trigger pipeline runs on changes to argocd manifests.

the reason I think this is sub-optimal is that the only time it really makes sense is if the user is using a single monorepo with the namespace. otherwise, the argocd manifests are un-naturally colocataed with a single components repo.

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,279 @@
...
+
+**Resource Scope:**
+- **Multi-Resource Support**: One GitOps repo can contain multiple applications, components, integrationTestScenarios, and releasePlans
+- **Single Namespace Constraint**: All resources from a GitOps repo deploy to its designated namespace only
+- **No Cross-Namespace Interaction**: Changes affecting N namespaces require changes in N separate GitOps repos
+
+**Registration Process (when using optional services):**
+1. Team registers GitOps repo via VS Code plugin, CLI (`konflux register`), or UI
+2. at registration time the desired namespace name is provided to (UI | API)
+3. GitOps Registration Service validates the repo and the requested namespace
```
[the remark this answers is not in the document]
brianwcook: i've provided for a way to turn off new registrations already when a cluster is at capacity

ADR/0047-gitops-onboarding-redesign.md
brianwcook: Who's going to create it?

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,279 @@
...
+    class Repo repoStyle
+    class NS nsStyle
+    class GRS optionalServiceStyle
+    class ArgoCD optionalServiceStyle
+```
+
+**Key Relationships:**
+- **Optional Registration**: GitOps Registration Service (when available) is used once per repo to create namespace and configure ArgoCD
+- **Optional Continuous Sync**: ArgoCD (when available) continuously propagates changes from repo to namespace
+- **Manual Alternative**: Without optional services, teams manually manage namespace creation and GitOps synchronization
```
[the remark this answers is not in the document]
brianwcook: I don't think we need humans in the way of creating namespace on clusters with adequate capacity. At _most_ we need a simple approval process, but I do not like the current process.

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,279 @@
...
+
+### UI Changes
+
+**Preserve:**
+- Dashboard
+- Build and release logs
+- Operational actions: triggering builds/tests, starting manual releases, pipeline management
+
+**Remove/Demote:**
+- UI forms for configuring components, applications, and other declarative resources
```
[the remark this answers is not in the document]
brianwcook: agree, I would like to keep that exact experience on the gitops registration page

ADR/0047-gitops-onboarding-redesign.md
ifireball: Is the proposed registration service basically doing what an ArgoCD repo with an app-of-apps in it would do? 

In other words, wouldn't we want to have GitOps for the registration as well?

Perhaps we want to automate away the registration by defining some heuristics to allow for automated detection of GitOps repos?

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,279 @@
+# 47. GitOps Onboarding Redesign
```
ifireball, line 1: @brianwcook do you mind letting #236 keep the 47 number?

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,279 @@
...
+
+### Core Shift
+- **Deprecate ability to configure tenant resources via the UI** Users will configure resources directly via Kubernetes or via Gitops
+- **Promot use of Git as the single source of truth** for applications, components, integration test scenarios, releasePlans and RBAC in production environments
+- **Preserve the UI for non-configuration activities** including:
+  - Monitoring and observability: logs, build inspection, viewing metrics, dashboards
+  - Operational actions: triggering builds/tests, starting manual releases, pipeline management
+  - Any other runtime actions that don't involve declarative resource configuration
+- **Commit to single onboarding flow**: UI-based onboarding will be completely deprecated with no fallback option to minimize maintenance overhead and deliver optimal user outcomes
+- **Maintain ease of use and excellent user experience** by providing aVS Code plugin that delivers forms based configuration of validation of tenant resources
```
staticf0x, line 56:
```suggestion
- **Maintain ease of use and excellent user experience** by providing a VS Code plugin that delivers forms based configuration of validation of tenant resources
```

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,279 @@
...
+
+### IDE Selection Rationale
+VS Code was chosen as the primary IDE target because:
```
staticf0x, line 59: My 2 cents, just random thoughts:

It might be a popular text editor, but by moving the onboarding UI into a VS code plugin effectively creates barriers for users who do not use VS code, doesn't it? A web UI would be more accessible in my opinion. I wouldn't expect any other commercial build system to force me to use VS code with their provided plugin(s). A CLI tool to help and validate the config yes, that would be nice. An optional VS code plugin for those who use that editor yes. But to remove the UI in favor of gitops and then use the plugin as a band aid doesn't seem very user oriented to me. If the onboarding experience is not ideal today, I would rather focus on why that is.

I'd like to suggest what has already been said in previous comments. Make a good UI, click on a button to open a PR in the user repo (the facilities should already exist for the pipeline-as-code PRs) and let the user review the config there.

What I am not seeing in this ADR is how do you help users without VS code? Do they just use the CI validation? That's fine I guess, but then, why duplicate the code in CI and the plugin while restricting the user base? Again, I'm not opposed to the idea of a plugin, but in my opinion it should be only treated as an optional - perhaps community developed - helper tool, while the web UI should be improved upon.

I also don't really agree with the early statement that the UI is

> Very easy to use with good validation and feedback

onboarding to Konflux has been pretty difficult for me, especially the first time. Compared to other (commercial) systems I think Konflux should focus on the UI way more than now, but this ADR IIUC goes in the opposite direction - which I don't really understand.

adambkaplan (commented):
I am going to throw a general :+1: on this ADR. My main concern is that the "GitOps Registration Service" and "ArgoCD" are marked as optional, when the mockup and other bits of the UX vision assume these things exist - ex does the UI need to handle scenarios where the registration service + ArgoCD are not deployed?

I see this architecture change laying the foundation for Backstage support, which supports [Git-based component registration](https://backstage.io/docs/getting-started/register-a-component).

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,279 @@
...
+  - Reduced disaster recovery capabilities due to lack of automated drift detection
+
+**Tenant Model:**
+- **Tenant Definition**: A tenant is equivalent to a Kubernetes namespace
+- **1:1 Mapping**: Each GitOps repository maps to exactly one Kubernetes namespace
+- **Namespace Declaration**: The namespace name is a parameter to the registration service. Once the repo is registered, the namespace it is immutable.
+- **Registration Immutability**: Re-registering the same repo is rejected to prevent namespace conflicts
+
+**Resource Scope:**
+- **Multi-Resource Support**: One GitOps repo can contain multiple applications, components, integrationTestScenarios, and releasePlans
```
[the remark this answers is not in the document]
gbenhaim: So maybe mention that those are just examples?

ADR/0047-gitops-onboarding-redesign.md
gbenhaim: The UI can also create the ArgoCD applications directly through the k8s api.

ADR/0047-gitops-onboarding-redesign.md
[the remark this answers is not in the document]
gbenhaim: How would you ensure that people don't create namespaces for their own use and by that fill up the cluster capacity?

gbenhaim:
I want to summarize the items which are important to me (without order of importance):

1. A mechanism for controlling who can create namespaces and how many. I don't want to be again in a situation where clusters are overflowed with personal namespaces.

2. Explain how resource quota is going to be managed and by who.

3. Konflux resources (application, component..) can be stored in the same repository which contains the components code (the openshift-pipelines team is doing it today)

4. Following up on 3, it should be possible to have multiple repositories with konflux resources which deploy to the same namespace.

5. It should be possible to manage multiple namespaces from a single gitops repository.

6. It should be possible to manage registration using gitops.

7. It should be possible to configure automatic registration, for example register all the repositories in a given github organization.

@brianwcook thank you for pushing this one!

brianwcook:
Most of this makes sense to me. I think that we are on hold for just a bit as one member of the KGC is out. I don't understand your #4 though. I had the opposite opinion. In my design, a namespace is "claimed" and associated with a gitops repo. Why would you want to have more than one gitops repo deploying to the same namespace?

gbenhaim:
> Most of this makes sense to me. I think that we are on hold for just a bit as one member of the KGC is out. I don't understand your item (4) though. I had the opposite opinion. In my design, a namespace is "claimed" and associated with a (one, only) gitops repo. Why would you want to have more than one gitops repo deploying to the same namespace?

(3) and (4) are very useful when used together. Imagine a team which owns multiple repositories, want to store konflux resources in the same repository with their code and have a single view in konflux for all of their pipelines (so they need to have everything in the same namespace).

tal-hason:
After discussing this with @ralphbean and others, I gave it some thought over the weekend.
One of the basic rules in GitOps is to protect the organization's structure via GitOps structure and controllers.

This will assist with how it manages what and where.

So from my POV, a namespace isn't a tenant. A namespace isn't a tenant, as a Product can have multiple namespaces to spread the builds across different "responsibilities" if the product has multiple teams working on it.

So this should be managed from the product's scope and not the "Konflux" configuration scope.

So a Product should be isolated from other products, via ArgoProjects or a dedicated argoCD instance.

Then each product should have its own GitOps repo(one or more), as this now becomes the product's responsibility for how to separate configuration.

The current namespaces are also here to isolate builds and resources from other teams and scale as needed.
But they should be nested under the "product" team and not as individual namespaces that later need to be aggregated by the user.

All this will allow the Product team to:
1. Get faster feedback regarding the sync state of Konflux objects.
2. Build better internal abstractions to manage multiple build processes for multiple versions.
3. leverage Conway's law in practies

ralphbean:
@konflux-ci/kgc talked today and agreed that this one isn't ready. We extended the request for comments period out 2 weeks to October 8.

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,279 @@
...
+
+### Negative Consequences
+
+1. **Tool Dependency**: Requires VS Code and plugin installation for optimal experience
+2. **Learning Curve**: Developers must adapt to new plugin-based workflow
+3. **Development Overhead**: Requires building and maintaining VS Code plugin and optional services
+4. **Manual Operations**: Without optional services, teams must manually manage GitOps operations, reducing automation benefits
+5. **Migration Complexity**: Existing users must migrate from current UI/GitOps hybrid approach
+6. **Configuration Complexity**: Teams must choose between manual setup (simple but limited) vs. optional services (full-featured but more complex)
+
```
mmorhun, line 225: Currently, some controllers use spec and annotations to communicate. For example, when one creates ImageRepository, there could be annotation that define for which Component this IR is created and operator automatically updates the required Component fields.
Another example with Image Controller is credentials rotation. A user needs to set a spec field to `true` and when the operator refresh the tokens, the field is returned to `false`. Will this play well with pure GitOps approach?

ADR/0047-gitops-onboarding-redesign.md
mmorhun: I agree that Konflux UI would better suit than an IDE plugin.

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,279 @@
...
+- Konflux needs a strong disaster recovery story and having Gitops-based tenant configuration is a significant advantage there.
+
+**Chosen approach** replaces both current flows with a unified Git-centric solution that preserves GitOps benefits while providing the usability improvements users expect.
+
+### Core Shift
+- **Deprecate ability to configure tenant resources via the UI** Users will configure resources directly via Kubernetes or via Gitops
+- **Promote use of Git as the single source of truth** for applications, components, integration test scenarios, releasePlans and RBAC in production environments
+- **Preserve the UI for non-configuration activities** including:
+  - Monitoring and observability: logs, build inspection, viewing metrics, dashboards
+  - Operational actions: triggering builds/tests, starting manual releases, pipeline management
```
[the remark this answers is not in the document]
brianwcook: such as manually starting a pipeline run.

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,279 @@
...
+- **Extensibility**: Enables building sophisticated onboarding wizards and configuration interfaces directly within the familiar IDE environment
+
+### Component Architecture
+
+| Component | Role | Optional |
+|-----------|------|----------|
+| VS Code Plugin | Main interface for onboarding and configuration | No |
+| Git Repos | Source of truth for Konflux config | No |
+| GitOps Registration Service | API for automated registration of GitOps repos as tenants | **Yes** (recommended) |
+| ArgoCD Service | Continuous GitOps sync and deployment | **Yes** (recommended) |
```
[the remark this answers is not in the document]
brianwcook: I'm sure someone could replace argocd with flux, but we are going to support argo

ADR/0047-gitops-onboarding-redesign.md
brianwcook: yeah, you aren't the only one to bring it up. I am updating the relationship for gitops repo to namespace to be 1:many

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,279 @@
...
+- Despite continuous UI improvements, many users still choose GitOps onboarding for its inherent disaster recovery and change management benefits
+- UI enhancements don't address the fundamental issue of maintaining two separate onboarding flows
+- Users who prefer GitOps principles cannot benefit from UI improvements, creating a permanent split in the user experience
+- Konflux needs a strong disaster recovery story and having Gitops-based tenant configuration is a significant advantage there.
+
+**Chosen approach** replaces both current flows with a unified Git-centric solution that preserves GitOps benefits while providing the usability improvements users expect.
+
+### Core Shift
+- **Deprecate ability to configure tenant resources via the UI** Users will configure resources directly via Kubernetes or via Gitops
+- **Promot use of Git as the single source of truth** for applications, components, integration test scenarios, releasePlans and RBAC in production environments
```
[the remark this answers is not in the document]
brianwcook: I don't understand the question. We will have a list of allowed objects for argoCD to sync. It should be the intersection of what konflux-admin is allowed to create + objects which are long lived (e.g., it won't include "snapshot" or any other objects which are garbage collected)

ADR/0047-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,279 @@
...
+
+## Security Implications
+
+A critical security consideration in any GitOps implementation is ensuring that the GitOps sync process cannot create arbitrary resources beyond what a namespace administrator would normally be permitted to create. Without proper constraints, GitOps operators could potentially be exploited to escalate privileges or deploy unauthorized resources.
+
+### Security Mitigation Approaches
+
+**ArgoCD Application Sync Using Impersonation:**
+ArgoCD impersonation allows GitOps sync operations to use dedicated, limited-privilege service accounts per namespace rather than the highly privileged control plane service account. This ensures that sync operations can only create resources that a namespace administrator would normally be allowed to create, providing strong tenant isolation through the principle of least privilege.
+
```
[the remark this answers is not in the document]
brianwcook: added

ADR/####-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,279 @@
...
+
+### Alternative Approaches Considered
+
+**Improving the existing UI** was considered but rejected because:
+- Despite continuous UI improvements, many users still choose GitOps onboarding for its inherent disaster recovery and change management benefits
+- UI enhancements don't address the fundamental issue of maintaining two separate onboarding flows
+- Users who prefer GitOps principles cannot benefit from UI improvements, creating a permanent split in the user experience
+- Konflux needs a strong disaster recovery story and having Gitops-based tenant configuration is a significant advantage there.
+
+**Chosen approach** replaces both current flows with a unified Git-centric solution that preserves GitOps benefits while providing the usability improvements users expect.
```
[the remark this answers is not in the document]
brianwcook: yes we do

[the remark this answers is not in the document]
mmorhun: Let's document it here then.

ADR/####-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,280 @@
...
+- **Promote use of Git as the single source of truth** for applications, components, integration test scenarios, releasePlans and RBAC in production environments
+- **Preserve the UI for non-configuration activities** including:
+  - Monitoring and observability: logs, build inspection, viewing metrics, dashboards
+  - Operational actions: triggering builds/tests, starting manual releases, pipeline management
+  - Any other runtime actions that don't involve declarative resource configuration
+- **Commit to single onboarding flow**: UI-based onboarding will be completely deprecated with no fallback option to minimize maintenance overhead and deliver optimal user outcomes
+- **Maintain ease of use and excellent user experience** by providing a VS Code plugin that delivers forms based configuration of validation of tenant resources
+
+### IDE Selection Rationale
+VS Code was chosen as the primary IDE target because:
```
mmorhun, line 57: While VSCode is popular choice, it still leaves behind many users. I think a CLI is a better option here.

ADR/####-gitops-onboarding-redesign.md
```diff
@@ -0,0 +1,280 @@
...
+7. Developer commits and pushes code (optionally via PR, with CI,reviews etc.)
+9. ArgoCD automatically syncs merged changes to the namespace
+10. sync status is reported back to PR, if there was one
+
+### Multi-Repo GitOps Model & Optional Services
+
+**GitOps Registration Service (Optional):**
+Konflux will optionally provide a **GitOps Registration Service** API that automates the registration and lifecycle of GitOps repositories as first-class tenants. This service is recommended for production deployments but not required for basic Konflux functionality.
+
+**ArgoCD Service (Optional):**
```
mmorhun, line 99: As of now, we already have some infinite loops between cluster and ArgoCD. They are caused by invalid yamls in user's GitOps repositories. It creates a lot of unwanted workload.
I think we should agree and document, that ArgoCD (or another GitOps tool) should not endlessly re-sync, but do it on commit or user request.

cit1zen: Maybe also give user ability to trigger sync manually from the UI/cli? So there is a simple and easy way how to trigger sync if the automatic one failed.

ADR/####-gitops-onboarding-redesign.md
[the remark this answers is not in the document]
brianwcook: done

ADR/####-gitops-onboarding-redesign.md
brianwcook: isn't there some way to tell argo not to keep retrying to sync something which already failed? this seems preventable

PR #241, opened 2025-07-30
ADR 50: Install a Backing Broker for Knative Eventing

skoved:
This ADR is to propose that Konflux pick a backing broker implementation for Knative Eventing.

**Before this ADR is accepted** Konflux needs to decide whether to use **Kafka** or **RabbitMQ**

### DIFF
--- a/ADR/0050-knative-kafka.md
+++ b/ADR/0050-knative-kafka.md
@@ -1,0 +1,34 @@
+# 50. Install a Backing Broker for Knative Eventing
+
+Date: 2025-07-30
+
+## Status
+
+Draft
+
+## Context
+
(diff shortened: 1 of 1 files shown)


ralphbean:
Some notes from the @konflux-ci/kgc discussion on this:

* This was submitted before our new process. We're following up on it now.
* We recognize that this is trying to address a gap, but we're not sure an ADR is the right venue.
* We heard that kubearchive has a controller-mode feature which may make this ADR less relevant.
* In any case, the upstream installer doesn't currently install add-ons (other than image-controller). How will this ADR be implemented, in practice?

@skoved do you think it still makes sense to pursue this as an ADR?

PR #254, opened 2025-09-05
Introduce ADR on the events API

ralphbean:
We talked about this in Brno in May of this year.


ralphbean:
Full disclosure: I'm a user and a lover of the events API. I introduced changes to MPC to expose *more* events because I found it useful to see what it was doing: https://github.com/konflux-ci/multi-platform-controller/pull/351

However, @gbenhaim argued the case made in this ADR.

For situations where users need troubleshooting information that they cannot get anywhere else, we should close those gaps. Let's list some of those here as considerations for acceptance or non-acceptance of this ADR.

In the MPC case, I'd like MPC to grow an explicit API so that you can see information about its success/failure/progress in the `.status` of that resource.

gbenhaim (approved):

ADR/nnnn-exclude-events-api-from-user-rbac.md
```diff
@@ -0,0 +1,33 @@
...
+
+Relates to [ADR 11. Roles and Permissions](0011-roles-and-permissions.md)
+
+## Context
+
+Konflux users require RBAC permissions to interact with various Kubernetes resources through our defined roles (Viewer, Contributor, Maintainer, Admin). The Kubernetes `events` API provides access to cluster events, which can contain information about resource changes and system activities. Users may find that API useful, but it raises other problems.
+
+## Decision
+
+We will exclude the Kubernetes `events` API from all Konflux user RBAC roles.
```
ifireball, line 17: What does this mean for resourc-attached events like the ones we are using to e.g. report on ProjectDevelopmentStreamTemplate application status?

ifireball: project-controller only has events RN, so we can't really implement this unless we make an alternative.

ifireball: Ack.

ifireball:
@gbenhaim maybe a little OT question - WRT MPC - how does our ability to report status change if/when the VM allocation logic will move into Kueue? Does Kueue provide for us with a natural place to report these things (Assuming we don't want to introduce many new CRs).

brianwcook (approved):

PR #255, opened 2025-09-08
feat(KFLUXUI-736): Create ADR for POC KITE project

CryptoRodeo:
An ADR was requested for this project: https://github.com/redhat-appstudio/infra-deployments/pull/8000#issuecomment-3260761640

### DIFF
--- a/ADR/xxxx-kite-poc-architecture-and-components.md
+++ b/ADR/xxxx-kite-poc-architecture-and-components.md
@@ -8,7 +8,7 @@ Proposed
 
 ## Context
 
-The [Kluster Issue Tracking Engine](https://github.com/Kluster-Issue-Tracking-Engine/kite) (KITE) is a proof-of-concept designed to detect, create, and track issues that block application releases in Konflux.
+The [KITE](https://github.com/Issues-Dashboard/kite) is a proof-of-concept designed to detect, create, and track issues that block application releases in Konflux.
 
 It prevents duplicate issue records, automates issue creation and resolution, and powers the **Issues Dashboard** where teams can view and manage disruptions.
 
@@ -46,9 +46,9 @@ The Konflux Issues Dashboard serves a fundamentally different purpose from tradi
 
 **With Issues Dashboard**:
 - Single issue: "Shared library v2.1 causing build failures"
-- **Scope**: Components affected by the same issue can fall under the same [scope](https://github.com/Kluster-Issue-Tracking-Engine/kite/blob/main/packages/backend/docs/Webhooks.md#issue-grouping-with-scope-objects).
+- **Scope**: Components affected by the same issue can fall under the same [scope](https://github.com/Issues-Dashboard/kite/blob/main/packages/backend/docs/Webhooks.md#issue-grouping-with-scope-objects).
 - **Grouping**: By grouping the issues, multiple teams are alerted of the same error, preventing separate debugging efforts.
-**Auto-resolution**: When the library is fixed, all related issues can be [resolved automatically](https://github.com/Kluster-Issue-Tracking-Engine/kite/blob/main/packages/backend/docs/Webhooks.md#automatic-issue-resolution-via-custom-webhooks) because of the shared scope.
+**Auto-resolution**: When the library is fixed, all related issues can be [resolved automatically](https://github.com/Issues-Dashboard/kite/blob/main/packages/backend/docs/Webhooks.md#automatic-issue-resolution-via-custom-webhooks) because of the shared scope.
 
 **3. User Support Rotation**
 
@@ -63,7 +63,7 @@ The Konflux Issues Dashboard serves a fundamentally different purpose from tradi
 **With Issues Dashboard**:
 - A non-technical associate can point to the Issues Dashboard and point them towards three issues related to failed builds.
   - Each issue has logs to those builds, allowing for deeper investigation.
-- A power-user associate can use the [KITE CLI tool](https://github.com/Kluster-Issue-Tracking-Engine/kite/tree/main/packages/cli) to quickly check if any existing issue records exist in the namespace(s) the user has access to.
+- A power-user associate can use the [KITE CLI tool](https://github.com/Issues-Dashboard/kite/tree/main/packages/cli) to quickly check if any existing issue records exist in the namespace(s) the user has access to.
   - The CLI tool reports three build failures in namespace `team-gamma`, with logs for each failure.
 
 **Value**:
@@ -186,7 +186,7 @@ We will implement KITE as a distributed system with the following key architectu
 
 ### Bridge Operator Architecture
 
-The **[KITE Bridge Operator](https://github.com/Kluster-Issue-Tracking-Engine/kite/tree/main/packages/operator)** implements the "bridge operator" pattern, which connects a Kubernetes environment with external systems not natively managed by Kubernetes.
+The **[KITE Bridge Operator](https://github.com/Issues-Dashboard/kite/tree/main/packages/operator)** implements the "bridge operator" pattern, which connects a Kubernetes environment with external systems not natively managed by Kubernetes.
 
 The operator:
 
@@ -199,10 +199,10 @@ The operator runs as a standard Kubernetes controller with cluster-wide permissi
 
 ### KITE Backend Service
 
-The **[KITE Backend](https://github.com/Kluster-Issue-Tracking-Engine/kite/tree/main/packages/backend)** is a Go-based REST API service that:
+The **[KITE Backend](https://github.com/Issues-Dashboard/kite/tree/main/packages/backend)** is a Go-based REST API service that:
 
-- **Provides API Endpoints**: Offers [REST API](https://github.com/Kluster-Issue-Tracking-Engine/kite/blob/main/packages/backend/docs/API.md) for creating, updating, and querying issues
-- **Webhook Support**: Includes [specialized webhook endpoints](https://github.com/Kluster-Issue-Tracking-Engine/kite/blob/main/packages/backend/docs/Webhooks.md) for simplified issue creation/resolution
+- **Provides API Endpoints**: Offers [REST API](https://github.com/Issues-Dashboard/kite/blob/main/packages/backend/docs/API.md) for creating, updating, and querying issues
+- **Webhook Support**: Includes [specialized webhook endpoints](https://github.com/Issues-Dashboard/kite/blob/main/packages/backend/docs/Webhooks.md) for simplified issue creation/resolution
 - **Issue Management**: Handles the complete lifecycle of issues (creation, updates, resolution)
 - **Database Integration**: Manages all database operations and data persistence
 - **Namespace isolation**: Issue access is namespace-restricted for isolation and security. (WIP)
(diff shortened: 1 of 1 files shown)


ADR/xxxx-kite-poc-architecture-and-components.md
```diff
@@ -0,0 +1,244 @@
...
+The backend is built using the [Gin](https://github.com/gin-gonic/gin) HTTP web framework and follows standard HTTP API patterns.
+
+### Team Integration Strategy
+
+KITE provides two primary integration paths for teams to onboard their services and start tracking issues:
+
+#### Recommended Integration Path
+
+**1. Build Custom Controllers**
+Teams [develop custom controllers](https://github.com/konflux-ci/kite/blob/main/packages/operator/docs/ControllerDevelopmentGuide.md) for the KITE Bridge Operator that:
```
ifireball, line 104: So all this service-specific knowledge would have to live in the Kite repo instead of the service repo?

ADR/xxxx-kite-poc-architecture-and-components.md
```diff
@@ -0,0 +1,244 @@
...
+- **Watch Specific Resources**: Monitor the Kubernetes resources relevant to their services (e.g., Deployments, Jobs, Custom Resources)
```
ifireball, line 105: K8s already have a well-known and well-understood issue detection and alerting system built-into Prometheus. While resource analisys can surface issues that we ma not be able to detect from existing Prometheuse metrics, I think a better way to do this would be to write custom exporters to bridge such issues into the Prometheus world.

ADR/xxxx-kite-poc-architecture-and-components.md
```diff
@@ -0,0 +1,244 @@
...
+- **Reports to Backend**: Sends failure information to the KITE backend service for persistence
+- **Extensible Design**: Can be extended with additional controllers for monitoring other resource types
+
+The operator runs as a standard Kubernetes controller with cluster-wide permissions to monitor resources across namespaces.
+
+### KITE Backend Service
+
+The **[KITE Backend](https://github.com/konflux-ci/kite/tree/main/packages/backend)** is a Go-based REST API service that:
+
+- **Provides API Endpoints**: Offers [REST API](https://github.com/konflux-ci/kite/blob/main/packages/backend/docs/API.md) for creating, updating, and querying issues
```
ifireball, line 89: Can we leverage existing issue management systems like Jira instead of making a custom one?

ADR/xxxx-kite-poc-architecture-and-components.md
```diff
@@ -0,0 +1,244 @@
...
+### Positive
+
+- **Scalability**: External database can handle large volumes of issue data while not overloading etcd
+- **Performance**: Database optimizations enable fast queries and reporting
+- **Extensibility**: Bridge operator pattern allows easy addition of new resource monitors
+- **Data Integrity**: Strong consistency guarantees prevent duplicate issues
+
+### Negative
+
+- **Complexity**: Additional infrastructure component (PostgreSQL) to manage
```
ifireball, line 231: And additional controllers and an additional stand-alone API. And there is no mention about how any of that is secured.

CryptoRodeo: > And additional controllers and an additional stand-alone API.

Yes good point, I'll add that.
> And there is no mention about how any of that is secured.

For the Backend API we do have a ticket to add a namespace check to ensure that a user has access to issues in a namespace: https://issues.redhat.com/browse/KFLUXUI-510

Are there any other security concerns you have?

ADR/xxxx-kite-poc-architecture-and-components.md
CryptoRodeo: I would imagine that the custom controllers would have to live within the Kite repository, unless there is some way to dynamically add a controller to an operator? Also which service repo are you referring to?

ADR/xxxx-kite-poc-architecture-and-components.md
CryptoRodeo: Prometheus exporters are great for alerting and trend monitoring. For our use case we need issue semantics, correlation, filtering and automatic closure (see https://issues.redhat.com/browse/KONFLUX-4103)

I went with a bridge operator with small controllers because it:
- **Has full Kubernetes contex**t: watches spec/status/events and can fetch related artifacts (for example: log snippets) to aid diagnosis.
- **Lowers the contribution barrier**: teams can implement a small reconciler with Operator SDK without needing PromQL/Alertmanager expertise.
- **Allows for easier, more extensive querying**: issue records include type, severity, scope, and other attributes, so the issues dashboard can filter exactly as required.
- **Allows for complex business logic**: teams can have full control over the logic of what is a success or failure

ADR/xxxx-kite-poc-architecture-and-components.md
```diff
@@ -0,0 +1,365 @@
...
+
+## Status
+
+Proposed
+
+## Context
+
+The [Konflux Issue Tracking Engine](https://github.com/konflux-ci/kite) (KITE) is a proof-of-concept designed to detect, create, and track issues that block application releases in Konflux.
+
+It prevents duplicate issue records, automates issue creation and resolution, and powers the **Issues Dashboard** where teams can view and manage disruptions.
```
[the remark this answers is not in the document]
CryptoRodeo: I added some use-cases: https://github.com/konflux-ci/architecture/pull/255/commits/f8378c6c34ad1e3e80552ee2156b3d27f5ea2bd9

ADR/xxxx-kite-poc-architecture-and-components.md
```diff
@@ -12,6 +12,121 @@ The [Konflux Issue Tracking Engine](https://github.com/konflux-ci/kite) (KITE) i
...
+**Value**:
+- Non-technical support associates can leverage the issues dashboard to quickly show them what is happening where and why.
+- Power-user support associates can quickly find the information they need using the CLI tool, getting them closer to where deeper analysis is needed.
+- Regular users also have access to these tools, potentially reducing the amount of support tickets filed.
+
+**4. Security Vulnerability Response**
+
+**Scenario**: A critical CVE is discovered in a widely-used dependency (e.g. a container base image, Python package, or RPM package) that affects multiple components across different teams.
+
+**Current Problem**:
```
gbenhaim, line 78: IMO, this should be solved by running pipelines that the user need to have access to in the user's namespace and include the CR which is responsible for triggering the pipelines in it as well. This will give the user visabilllity on the the plrs and the starus of the cr
 Another approach is to includec additional information in the status field of the component.

ADR/xxxx-kite-poc-architecture-and-components.md
```diff
@@ -12,6 +12,121 @@ The [Konflux Issue Tracking Engine](https://github.com/konflux-ci/kite) (KITE) i
...
+```
+
+**Resolution Workflow**:
+1. **Immediate Alert**: All affected teams see the security issue in their dashboard when MintMaker detects failures
+2. **Targeted Remediation**: Teams can address specific failure causes (refresh tokens, fix registry auth, resolve conflicts)
+3. **Progress Tracking**: As MintMaker successfully creates security PRs for each component and those PRs get merged, the issue scope automatically shrinks
+4. **Auto-Resolution**: Issue resolves completely when all components have successful security updates merged
+
+**Key Value**:
+- Instead of X teams independently discovering a critical security vulnerability days or weeks later through security scanners, all teams are immediately aware with specific remediation steps.
```
gbenhaim, line 107: Today konflux allows to upload the sbom of a component to an SBOM server (for example RHTPA) that will show the cves related to the artifact. Isn't it enough?

ADR/xxxx-kite-poc-architecture-and-components.md
```diff
@@ -12,6 +12,121 @@ The [Konflux Issue Tracking Engine](https://github.com/konflux-ci/kite) (KITE) i
...
+
+**Current Problem**:
+- Depending on the supporting associates skill set, they may or may not have the knowledge of how to investigate such issues.
+  - Querying a Konflux cluster and checking the logs for a failing pipelinerun can require multiple skills that are out of reach for a supporting associate.
+- The user might have to navigate through multiple views in order to get a clear picture of what could be the cause.
+- A power user capable of investigating logs for a CR on a Konflux cluster still needs to figure out where to check.
+
+**With Issues Dashboard**:
+- A non-technical associate can point to the Issues Dashboard and point them towards three issues related to failed builds.
+  - Each issue has logs to those builds, allowing for deeper investigation.
```
gbenhaim, line 65: It is also possible to show only failed pipelines and click on them and see the logs without KITE

ADR/xxxx-kite-poc-architecture-and-components.md
```diff
@@ -12,6 +12,121 @@ The [Konflux Issue Tracking Engine](https://github.com/konflux-ci/kite) (KITE) i
...
+- They have a summary of all existing issues related to their applications, components, etc
+- Rather than having to dig through each issue individually across multiple views and then decide what to work on first, issues can be filtered by severity.
+  - This allows them to work on critical issues first, potentially unblocking them from moving forward.
+
+**Value**:
+- Time saved (e.g. 5-30 minutes daily)
+- Immediate awareness of all blocking issues
+
+
+**2. Cross-Component Failure Correlation**
```
gbenhaim, line 41: Any real user reported such issue in the past?

ADR/xxxx-kite-poc-architecture-and-components.md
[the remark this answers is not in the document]
gbenhaim: Thanks for adding those. I posted some comments.

ADR/xxxx-kite-poc-architecture-and-components.md
```diff
@@ -12,6 +12,121 @@ The [Konflux Issue Tracking Engine](https://github.com/konflux-ci/kite) (KITE) i
...
 
+## Concrete Uses Cases and User Workflows
+## Why Konflux Issues Dashboard vs External Issue Trackers
+
+The Konflux Issues Dashboard serves a fundamentally different purpose from traditional issue tracking systems like Jira. Rather than replacing external issue management, it acts as an **operational monitoring dashboard**, similar to a car dashboard that alerts you to immediate mechanical problems that need attention *now*, not planning or project management issues.
+
+## Primary Use Cases
+**1. Developer Morning Check-in Workflow**
+
+**Scenario**: A developer is managing 8 components across 3 different applications.
```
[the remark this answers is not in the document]
CryptoRodeo: This is news to me, yeah if component groups become a thing and all the components in that group are in the same view then the use case of having more visibility across multiple components is covered IMO.

ADR/xxxx-kite-poc-architecture-and-components.md
gbenhaim: | failed releases, failed builds

Those are already visible in the UI
 
| MintMaker fails

As I mentioned in the other comments, I think that the problem is that mintmaker doesn't run the dependency update pipelines in the user's namespace. In addition, if the is any configuration issue, it can report it through the status section of the component resource.

ADR/xxxx-kite-poc-architecture-and-components.md
brianwcook: I don't think we intend to replicate the features of an SBOM analysis service; we have data from both minitmaker and Clair which are not surfaced to users in such a way that it is easily actionable and this is an example of how to surface that data better.

ADR/xxxx-kite-poc-architecture-and-components.md
```diff
@@ -0,0 +1,480 @@
...
+
+Date: 2025-09-08
+
+## Status
+
+Proposed
+
+## Context
+
+The [Konflux Issue Tracking Engine](https://github.com/konflux-ci/kite) (KITE) is a proof-of-concept designed to detect, create, and track issues that block application releases in Konflux.
```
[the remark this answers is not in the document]
arewm: I am also curious what other solutions exist in this problems space. If any exist, why are they insufficient? Would it be challenging to contribute to any of these to make them sufficient?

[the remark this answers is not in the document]
CryptoRodeo: No objections from me, it can follow the same route as KubeArchive. 
Any objections @cit1zen ?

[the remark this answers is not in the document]
CryptoRodeo: @cit1zen is not near a PC, but he said he agrees and has no objections too

PR #262, opened 2025-09-30
Introduce ADR for Trusted Tasks model

chmeliik:
The current Trusted Tasks model will not hold up after the decentralization of build-definitions.

Propose an alternative model for trusting Tasks.

Note: this ADR aims to get rid of `data-acceptable-bundles`, but does not propose changes to Task versioning. While those would be related, mixing both concepts in one ADR would not be effective.


ADR/nnnn-trusted-task-model.md
```diff
@@ -0,0 +1,238 @@
...
+### Git references to upstream Konflux Tasks no longer trusted
+
+In the current model, any change that gets merged to build-definitions is considered
+trusted, so the `trusted_tasks` data includes both `git` references and `oci` bundle
+references.
+
+The new trust model for upstream Konflux Tasks is based around trusting the bundles
+released through Red Hat's deployment of Konflux. We should not try to establish
+trust for the `git` references to the source code for those tasks, because there will
+be no way to know whether the reference passed release-time verifications or not.
```
ralphbean, line 153: I'd love to keep the option to still trust git references. Can you think of a reasonable way to still make it possible?

What if the release process signed git tags and pushed them back to the git repository?

We could leave this out of scope for _this_ decision and follow up with another one later that adds this capability.

I know, at least in the Fedora space, users have expressed a desire to see git references rather than bundle references because they're easier to follow as a user to understand what's happening. Easier to hack on.

ADR/nnnn-trusted-task-model.md
```diff
@@ -0,0 +1,238 @@
...
+
+- We implicitly trust all tasks in `quay.io/konflux-ci/tekton-catalog/*`
+- We trust anyone who can write to any of those repositories
+  - Which is the same list as for `data-acceptable-bundles`
+
+With signing key verification:
+
+- We trust only the release pipelines in the Red Hat deployment of Konflux
+
+### Git references to upstream Konflux Tasks no longer trusted
```
ralphbean, line 144: I like the "voice" of the following bullet about the expiry mechanism. Can we copy it here? It is direct.

```suggestion
### Decentralized Tasks lose capability to trust Git references
```

ADR/nnnn-trusted-task-model.md
lcarva: One of the sticky points with git is that for any single file, it could have the same exact content across different commit IDs. (A new commit ID can modify other files, for example.) In this repo, we chose to only use the [git commit ID that modified the file](https://github.com/konflux-ci/build-definitions/blob/8fcc48dc4f857db446be0c92c544937fc36ad6e0/.tekton/scripts/build-acceptable-bundles.sh#L26). It's not very intuitive but it mostly works :)

With the caveat above, using signatures on git commits is probably a good way forward to establish trust, but this would have to be done by an external entity, like a release pipeline. It's definitely worth investigating more.

ADR/nnnn-trusted-task-model.md
```diff
@@ -0,0 +1,238 @@
...
+  - Release pipelines in the Red Hat deployment of Konflux
+  - The CI pipelines for build-definitions
+  - The CI pipelines for other repos that happen to use the same OpenShift namespace
+    as build-definitions
+
+The proposed trust model, without signing key verification:
+
+- We implicitly trust all tasks in `quay.io/konflux-ci/tekton-catalog/*`
+- We trust anyone who can write to any of those repositories
+  - Which is the same list as for `data-acceptable-bundles`
```
lcarva, line 138: I was going to object to this but you're right. There are folks that can already update the list of trusted tasks today outside of CI. So this wouldn't be a different trust model.

ADR/nnnn-trusted-task-model.md
chmeliik: Rephrased some parts of this section and mentioned the suggested approaches (signed tags, separate "release repo") as possible future directions

ADR/nnnn-trusted-task-model.md
```diff
@@ -0,0 +1,249 @@
...
+### Decentralized Tasks lose the trust in their Git references
+
+*...For now.*
+
+In the current model, any change that gets merged to build-definitions is considered
+trusted, so the `trusted_tasks` data includes both `git` references and `oci` bundle
+references.
+
+The new trust model for upstream Konflux Tasks is based around trusting the bundles
+released through Red Hat's deployment of Konflux. Initially, we should not try to
```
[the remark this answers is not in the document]
chmeliik: See "Start signing upstream Konflux tasks with a common release key"

ADR/nnnn-trusted-task-model.md
```diff
@@ -0,0 +1,249 @@
...
+Currently, it's common for Konflux Tasks to release over and over again with the
+same version tag. When the version tag moves to a new revision of the Task, the
+older revision begins its 60 day expiry period.
+
+Such a mechanism would be difficult to replicate with the rule-based model.
+
+An alternative expiry mechanism could be version based. This ADR does not aim to
+define the mechanism, because the current versioning practices do not allow a
+version based expiry to be effective. This will be addressed, along with the expiry
+mechanism, in a future ADR about Task versioning. The example in the Appendix
```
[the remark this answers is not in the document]
chmeliik: There's still the `expires-on` annotation for the important cases. Updated the doc to clarify that that would still work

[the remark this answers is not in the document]
gbenhaim: Conforma checks this annotation?

ADR/nnnn-trusted-task-model.md
```diff
@@ -0,0 +1,245 @@
+# NN. Trusted Tasks model after build-definitions decentralization
+
+Date: 2025-09-30
+
+## Status
+
+Proposed
+
```
ralphbean: I want to cross-link this the "partner tasks" ADR from way back when. De-centralizing build-definitions is essentially a new way to "do partner tasks".

ADR/nnnn-trusted-task-model.md
```diff
@@ -0,0 +1,245 @@
...
+defining a rule that allows all task bundles from `quay.io/konflux-ci/tekton-catalog/*`
+with no further conditions. The `trusted_task_rules` mechanism must support this.
+
+### Start signing upstream Konflux tasks with a common release key
+
+Trusting all task bundles from `quay.io/konflux-ci/tekton-catalog/*` is not a strong
+trust model (but isn't weaker than the current one - see [Consequences](#consequences)).
+To let policy authors define a stronger trust model, we should aim to start signing
+the upstream Konflux task bundles with a common release key and extend the
+`trusted_task_rules` mechanism to support signature verification.
```
arewm: Should we have specifics about how to establish trust in these tasks as a minimum suggested gate? This has some relevancy with the partner task ADR as well as @ralphbean already mentioned.

ADR/nnnn-trusted-task-model.md
```diff
@@ -0,0 +1,245 @@
...
+
+```yaml
+trusted_task_rules:
+  allow:
+    - name: Implicitly trust all tasks from konflux-ci/tekton-catalog
+      pattern: oci://quay.io/konflux-ci/tekton-catalog/*
+
+    - name: Require common signing key  # starting in 2026
+      pattern: oci://quay.io/konflux-ci/tekton-catalog/*
+      signing_key: <common public key for konflux-ci Tasks>
```
ralphbean: There's a weakness we have today that we should try to close while we're making this change (or a subsequent change).

Imagine we're trusting tasks from quay.io/konflux-ci/tekton-catalog/* and quay.io/third-party/cool-tasks/* That third-party team made a cool new task that can build microsoft installer artifacts. Great! What stops that third-party team from publishing bundles of tasks that include a *buildah* task, that does not great things.

We should add a criterion to these allow blocks that restrict the **task names** that we'll accept from each of these locations / signing combinations.

ADR/nnnn-trusted-task-model.md
chmeliik: Do you mean rules for letting tasks get released to `quay.io/konflux-ci/tekton-catalog`?

We probably should have those, but I don't know how to go about it. It seems like that is up to the reviewers in the konflux-release-data repo in an internal gitlab instance :/

ADR/nnnn-trusted-task-model.md
chmeliik: I think the rule-based mechanism in this example would support that natively. If you don't implicitly trust the whole organization, then you wouldn't want to use a pattern to allow `quay.io/third-party/cool-tasks/*`, but instead list out the individual allowed tasks.

Maybe that was more obvious in an earlier iteration of the doc, which had
```yaml
    - task_refs:
        - oci://quay.io/konflux-ci/tekton-catalog/task-buildah
        - oci://quay.io/konflux-ci/tekton-catalog/task-buildah-oci-ta
        - oci://quay.io/konflux-ci/tekton-catalog/task-buildah-remote
        - oci://quay.io/konflux-ci/tekton-catalog/task-buildah-remote-oci-ta
```
Instead of the `pattern`

ADR/nnnn-trusted-task-model.md
arewm: It would be related to the conforma policy rules required for accepting/publishing those task bundles by some entity that runs a Konflux instance. Another alternative for partner tasks would be to let those to be kept separate and still establish them as a trusted task.

ADR/nnnn-trusted-task-model.md
ralphbean: I see what you mean. I think I would still put the replaced-by link in the partner task one, just so that people can find their way here when they stumble on that one.

ralphbean (approved):

arewm (approved):

PR #266, opened 2025-10-06
ADR for Task versioning

chmeliik:
Proposes Task versioning changes that have been needed for a while. The need has reached critical mass with the build-definitions decentralization.

Related to https://issues.redhat.com/browse/KONFLUX-4531

### DIFF
--- a/ADR/nnnn-task-versioning.md
+++ b/ADR/nnnn-task-versioning.md
@@ -1,0 +1,447 @@
+# NN. Start versioning Tekton Tasks responsibly
+
+Date: 2025-10-06
+
+## Status
+
+Proposed
+
+Relates to
+[ADR NN. Trusted Tasks model after build-definitions decentralization](nnnn-trusted-task-model.md).
+
+## Context
+
+The upstream Tasks that Konflux releases today typically do not follow a sensible
+versioning scheme. They do have version numbers, but the numbers serve only two
+purposes:
+
+- Want to make breaking changes? Bump the minor or major version number so that
+  you can add a MIGRATION.md document that users will see in the MintMaker PR that
+  updates their Tasks.
+- Want to attach a migration script that MintMaker will execute when updating
+  users' Tasks? Bump any segment of the version number.
+
+For other kinds of changes, whether they be new features, bug fixes or periodic
+dependency updates, Task maintainers leave version numbers unchanged.
+
+This has consequences.
+
+### Users get opaque updates
+
+Most of the time, when a Konflux user receives a PR that updates their Tasks, they
+see something like this:
+
+| Package                                                      | Change                 | Notes                                      |
+| ---                                                          | ---                    | ---                                        |
+| quay.io/konflux-ci/tekton-catalog/task-apply-tags            | `e0de426` -> `f44be1b` |                                            |
+| quay.io/konflux-ci/tekton-catalog/task-build-image-index     | `3bf6e4e` -> `79784d5` |                                            |
+| quay.io/konflux-ci/tekton-catalog/task-buildah-remote-oci-ta | `0.4` -> `0.5`         | :warning:[migration][migration-1]:warning: |
+| quay.io/konflux-ci/tekton-catalog/task-clair-scan            | `0.2` -> `0.3`         | :warning:[migration][migration-2]:warning: |
+
+For the tasks with breaking changes, they get *some* (usually insufficient) information
+about what changed.
+
+For those without breaking changes, they get nothing. What changed between
+`e0de426` and `f44be1b`? Not even the maintainers know.
+
+It would not be at all unreasonable for a user to refuse to merge these kinds of
+PRs (except that Conforma would eventually flag their tasks as outdated.)
+
+### Maintainers cannot easily communicate critical fixes
+
+When a Task maintainer discovers and fixes a critical bug, they typically wish
+to announce this publicly, e.g. on the Konflux mailing list.
+
+With the current versioning practices, how does the maintainer go about this?
+
+> We've discovered a bug in task-foo version 0.2. Please upgrade to version 0.2.
+
+Hmm, no, that's not right.
+
+> We've discovered a bug in task-foo:0.2@sha256:baadbaad. Please upgrade to sha256:deadbeef
+
+Better. But a user has 0.2@sha256:baddecaf. Are they affected? Who knows.
+
+> We've discovered a bug in task-foo version 0.2, revisions released between 2025-08-01
+  and 2025-10-01. Please upgrade to a sufficiently new revision.
+
+Okay, at least it's an interval now. But are we really going to ask the user to run
+`skopeo inspect --no-tags --format '{{.Created}}' docker://${my_task_ref}` to figure
+out if they're affected?
+
+> We've discovered a bug in task-foo versions >=0.2.2,<0.2.5. Please upgrade to
+  version 0.2.5.
+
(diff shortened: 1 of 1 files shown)


ADR/nnnn-task-versioning.md
```diff
@@ -0,0 +1,361 @@
+# NN. Start versioning Tekton Tasks responsibly
+
+Date: 2025-10-06
+
+## Status
+
+Proposed
+
+Relates to [ADR NN. Trusted Tasks model after build-definitions decentralization](ADR/tbd.md).
```
chmeliik, line 9: https://github.com/konflux-ci/architecture/pull/262

ADR/nnnn-task-versioning.md
```diff
@@ -0,0 +1,361 @@
...
+MintMaker links MIGRATION.md documents in the PRs it sends to Konflux users (see
+the example [above](#users-get-opaque-updates)). This feature assumes that each
+`major.minor` version comes with a MIGRATION.md document, that this document is
+at `/task/${task_name}/${version}/MIGRATION.md` in the source repository and that
+the source repository is <https://github.com/konflux-ci/build-definitions>. The
+feature will not work after decentralization.
+
+<!-- TODO: How will MintMaker figure out a link to the changelog? -->
+MintMaker will stop trying to link MIGRATION.md files and instead link the changelog.
+In case of breaking changes, the changelog should explain how the user can migrate.
```
chmeliik, line 139: @konflux-ci/mintmaker-maintainers I would appreciate your help with this.

Did you have some plans for how the MIGRATION.md links would keep working after decentralization? Or any ideas how we could make CHANGELOG.md discovery work?

chmeliik: I know that Renovate also has some support for discovering CHANGELOG.md files in the source repo (this ADR proposes we should use those): https://docs.renovatebot.com/key-concepts/changelogs/

But I imagine this looks for CHANGELOG.md in the repo root or some other known locations and also wouldn't know where to find the correct file for a Tekton task.

> We can potentially implement a new source for the releases in our downstream fork of Renovate.

Yeah, that might be the way. I'm thinking we would include a link to the CHANGELOG.md as an annotation on the task bundle (similar to [this one](https://github.com/konflux-ci/build-definitions/blob/9989c73a5246e6b52f53ff4dff2ffc678c8d1630/task/tkn-bundle/0.2/tkn-bundle.yaml#L404)) and MintMaker would use that

ADR/nnnn-task-versioning.md
```diff
@@ -0,0 +1,361 @@
...
+#### Dealing with MintMaker updates
+
+Automated PRs for dependency updates may become a problem. They're already seen
+as a chore today by many Task maintainers. If manual action is required before
+a MintMaker PR can merge, the most likely outcome is that MintMaker PRs will stop
+getting merged.
+
+There's a few options to deal with this:
+
+- Do not consider them meaningful. Merge the PR without updating the changelog or
```
mmorhun, line 268: > Do not consider them meaningful

Unless they break stuff. But then a developer need to step in and fix the problem updating the change log. So, it should work if we don't release changes without version bump.

ADR/nnnn-task-versioning.md
```diff
@@ -0,0 +1,427 @@
...
+change, we cannot ask Task maintainers to follow this layout. See Appendix 1:
+[The technicals of Task versioning](#the-technicals-of-task-versioning) for more details.
+
+### Keep a changelog
+
+Version numbers on their own are nice, but without an easy way to see what changed
+in each version, they're not that useful.
+
+Document the meaningful changes in each version in a CHANGELOG.md file. The changelog
+for a specific Task will be at `/task/${task_name}/CHANGELOG.md` in the repository.
```
mmorhun: Are we gonna always require it at this specific path?

ADR/nnnn-task-versioning.md
```diff
@@ -0,0 +1,427 @@
...
+
+Today, MintMaker links MIGRATION.md documents in the PRs it sends to Konflux users
+(see the example [above](#users-get-opaque-updates)). This feature assumes that each
+`major.minor` version comes with a MIGRATION.md document, that this document is
+at `/task/${task_name}/${version}/MIGRATION.md` in the source repository and that
+the source repository is <https://github.com/konflux-ci/build-definitions>. The
+feature will not work after decentralization.
+
+MintMaker will stop trying to link MIGRATION.md files and will instead link the changelog.
+In case of breaking changes, the changelog should explain how the user can migrate.
```
arewm: Is this proposing to do with automated remediation via the migration scripts?

ADR/nnnn-task-versioning.md
chmeliik: The relation from the other direction is only tangential. Task versioning would be needed to make version-based rules work better, but apart from that the trusted task rules would work on their own

ADR/nnnn-task-versioning.md
```diff
@@ -0,0 +1,427 @@
...
+  updates their Tasks.
+- Want to attach a migration script that MintMaker will execute when updating
+  users' Tasks? Bump any segment of the version number.
+
+For other kinds of changes, whether they be new features, bug fixes or periodic
+dependency updates, Task maintainers leave version numbers unchanged.
+
+This has consequences.
+
+### Users get opaque updates
```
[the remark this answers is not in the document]
chmeliik: > Should we suggest that tasks use a non-0 minor version unless it is still in its initial active development?

The suggestion was there, but a bit weak. Re-worded the relevant phrase

ADR/nnnn-task-versioning.md
```diff
@@ -0,0 +1,427 @@
...
+  as git is concerned, the files have nothing to do with each other. Version 1.2
+  is now missing a bugfix from version 1.1.
+- It's enough of a chore to discourage developers from versioning.
+
+The suggested repository layout going forward would be:
+
+```text
+task
+└── hello
+    └── hello.yaml    (version defined only by the app.kubernetes.io/version label)
```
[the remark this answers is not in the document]
chmeliik: I think the _requirement_ would be to have the `org.opencontainers.image.version` annotation on the task bundle (where Conforma can read it) and to tag the bundle with the same version (so that MintMaker can update versions). I'll document this.

In practice, we'll most likely end up implementing it in the tekton-bundle-builder pipeline based on the `app.kubernetes.io/version` label, and it may be good to standardize it. But I can also think of a reason to have the version elsewhere, e.g. a separate VERSION file. That would enable a trivial implementation of the "build the task only if the version number changed" CI approach (trigger the pipeline on changes to the VERSION file).

ADR/nnnn-task-versioning.md
chmeliik: (But migration scripts will also still work)

ADR/nnnn-task-versioning.md
chmeliik: On second thought, it doesn't hurt to have the link. But I think I'll add it after the PRs are merged

ADR/nnnn-task-versioning.md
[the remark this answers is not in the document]
chmeliik: Documented the task bundle requirements in the "Mark each meaningful change with a version update" section. Also added a note about the current setup, which does use `app.kubernetes.io/version` as the main source of truth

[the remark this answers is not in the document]
ralphbean: Thank you!

PR #268, opened 2025-10-09
ADR on source provenance verification

ralphbean:


ADR/####-slsa-source-provenance-verification.md
```diff
@@ -0,0 +1,133 @@
...
+- **Tekton Chains**: Existing Chains infrastructure generates attestations about the verification task
+- **Enterprise Contract**: Conforma policies updated to verify the attestation chain
+- **Trusted Artifacts** ([ADR 36](0036-trusted-artifacts.md)): Source VSA artifacts follow the similar OCI storage patterns to other trusted artifacts
+
+## Consequences
+
+### Positive
+
+- **Consistent trust model**: All Conforma verification continues through Tekton Chains attestations
+- **Storage model alignment**: Source attestations stored as git notes (close to source), build attestations about them stored in OCI (close to artifacts)
```
chmeliik: Should be removed for the same reasons as https://github.com/konflux-ci/architecture/pull/268#discussion_r2419811989 I think

ADR/####-slsa-source-provenance-verification.md
```diff
@@ -0,0 +1,133 @@
...
+   - These VSAs are stored as git notes in the source repository, contemporaneously with commits
+   - VSAs attest to source control practices (e.g., two-person review, branch protection, audit logs)
+
+2. **Source provenance verification task** (in Konflux pipeline):
+   - A Konflux pipeline task reads the source VSA from git notes for the cloned commit
+   - The task accepts a public key parameter (provided via IntegrationTestScenario configuration) and verifies the source VSA signature against that key
+   - The task verifies the VSA contents (format, subject matching the commit, etc.)
+   - The task pushes the source VSA to the OCI registry as a separate artifact
+   - The task exposes the pullspec of the OCI representation of the VSA (digest reference) in its task results using `*_ARTIFACT_OUTPUTS` format
+   - The task records the public key used for verification in its results
```
chmeliik: Nitpick: does the task need to return it as a result if it's already recorded in parameters? It's unforgeable in the attestation either way

One can never be too careful about returning too many bytes from Tekton Tasks

chmeliik (approved):
This makes sense to me. Do we also want to say something about git submodules, or leave that for the future?

ADR/####-slsa-source-provenance-verification.md
```diff
@@ -0,0 +1,133 @@
...
+
+1. **Source VSA generation** (external to Konflux):
+   - Source repository maintainers configure automation (GitHub Actions, GitLab CI/CD pipelines, etc.) to generate source VSAs
+   - These VSAs are stored as git notes in the source repository, contemporaneously with commits
+   - VSAs attest to source control practices (e.g., two-person review, branch protection, audit logs)
+
+2. **Source provenance verification task** (in Konflux pipeline):
+   - A Konflux pipeline task reads the source VSA from git notes for the cloned commit
+   - The task accepts a public key parameter (provided via IntegrationTestScenario configuration) and verifies the source VSA signature against that key
+   - The task verifies the VSA contents (format, subject matching the commit, etc.)
```
ifireball: Do we also want to verify if what the Source VSA claims is actually implemented, e.g. if requiring two reviews is actually enforced in GitHub?

ADR/####-slsa-source-provenance-verification.md
ralphbean: Done!

ADR/####-slsa-source-provenance-verification.md
```diff
@@ -0,0 +1,130 @@
...
+
+1. **Direct reference**: Include source provenance attestations directly in build provenance
+2. **Parallel verification**: Verify source provenance independently alongside build provenance
+3. **Chained attestations**: Create build provenance attestations about source provenance attestations
+
+The challenge is that the source track's storage model (git notes in repositories) differs significantly from our build artifact storage model (OCI registries), making it non-obvious how to integrate source provenance verification into our existing Conforma policies and attestation chains.
+
+### References
+
+- [SLSA Source Track Requirements](https://slsa.dev/spec/draft/source-requirements) - Source provenance specification
```
arewm: The draft spec is more accurate even if it is moving. The RC1 spec is out of date. We should update this to either the RC2 reference (once created) or just wait until we have created the official SLSA v1.2 release.

ADR/####-slsa-source-provenance-verification.md
```diff
@@ -0,0 +1,130 @@
...
+1. **Source VSA generation** (external to Konflux):
+   - Source repository maintainers configure automation (GitHub Actions, GitLab CI/CD pipelines, etc.) to generate source VSAs
+   - These VSAs are stored as git notes in the source repository, contemporaneously with commits
+   - VSAs attest to source control practices (e.g., two-person review, branch protection, audit logs)
+
+2. **Source provenance verification task** (in Konflux pipeline):
+   - A Konflux pipeline task reads the source VSA from git notes for the cloned commit
+   - The task accepts a public key parameter (provided via IntegrationTestScenario configuration) and verifies the source VSA signature against that key
+   - The task verifies the VSA contents (format, subject matching the commit, etc.)
+   - The task pushes the source VSA to the OCI registry as a separate artifact
```
arewm: How will this be discovered? Do we attach it with cosign or use the referrer's API directly? If we want to push this so that Chains can generate an attestation for it, it must be pushed as its own manifest. That means that we cannot use the backwards compatible tag-based cosign pattern. Should the task check to see if the referrer's API is available before it tries to export the digest reference of the VSA?

Also note that if we don't modify the VSA, the subject will be the git commit, not the built artifact.

ADR/####-slsa-source-provenance-verification.md
```diff
@@ -0,0 +1,130 @@
...
+
+3. **Build provenance chain** (via Tekton Chains):
+   - Tekton Chains observes the task results containing the source VSA artifact reference
+   - Chains generates and signs a SLSA build provenance attestation **about** the source provenance attestation
+   - This creates an attestation chain: `Build Provenance → Source VSA → Source Commit`
+
+4. **Verification in Conforma**:
+   - Conforma policies can verify source provenance by checking for the presence of the attestation chain
+   - Conforma trusts Tekton Chains' signatures (existing trust root)
+   - Conforma reads the public key from the test attestation's task parameters and verifies it matches the expected signer identity for the source repository
```
arewm: This would be fetched from the provenance of that attestation?

ADR/####-slsa-source-provenance-verification.md
```diff
@@ -0,0 +1,130 @@
...
+   - Through the chain, Conforma gains transitive trust in the source VSA's signature verification (performed by the task) without needing to re-verify the signature itself
+   - This bootstraps trust in source provenance through our existing build-provenance infrastructure
+
+### Rationale
+
+This approach provides several advantages:
+
+- **Preserves existing trust model**: Conforma continues to trust Tekton Chains as its primary attestation source
```
arewm: Is this trust model also based on our trust of the tasks?

ADR/####-slsa-source-provenance-verification.md
```diff
@@ -0,0 +1,130 @@
...
+  - The verification task attempts to read git notes that don't exist yet
+
+  **Mitigation**: The movable test pattern from ADR 48 significantly helps with this issue. Because source verification runs as an independent integration test scenario rather than an embedded build task, it can be easily re-run once the source VSA is available. This is much easier than if the verification were embedded in the build pipeline, which would require a complete rebuild.
+
+- **Additional task required**: New verification task must be implemented and maintained (poc in [konflux-ci/build-definitions#2867](https://github.com/konflux-ci/build-definitions/pull/2867))
+- **OCI storage overhead**: Source VSAs are duplicated (git notes + OCI registry)
+- **Complexity in Conforma**: Policies must understand attestation chains rather than direct attestations
+- **External dependencies**: Requires source repository owners to set up VSA generation automation
+
+### Implementation Requirements
```
arewm: These may need to be updated based on the above discussions.

ADR/####-slsa-source-provenance-verification.md
```diff
@@ -0,0 +1,130 @@
...
+   - Verify build provenance attestations about source VSAs
+   - Read the public key from test attestation task parameters and verify it matches expected signer identities for source repositories
+   - Determine if source verification levels meet release requirements
+
+3. **Documentation for source repository owners**:
+   - How to generate source VSAs (GitHub Actions, GitLab CI templates)
+   - How to store VSAs as git notes
+   - What source verification levels mean
+
+4. **ITS Provisioning**: Update Konflux onboarding mechanisms to provision a source verification ITS automatically
```
arewm: Are you proposing to enable this globally?

ADR/####-slsa-source-provenance-verification.md
```diff
@@ -0,0 +1,130 @@
...
+- **Race condition risk**: Source VSAs should be generated contemporaneously with commits (typically via GitHub Actions or GitLab pipelines), but Konflux's verification task may execute before the external automation completes. This can cause initial verification failures when:
+  - A commit is pushed to a repository
+  - Konflux immediately triggers a build pipeline
+  - The source VSA generation (GitHub Action/GitLab pipeline) is still running
+  - The verification task attempts to read git notes that don't exist yet
+
+  **Mitigation**: The movable test pattern from ADR 48 significantly helps with this issue. Because source verification runs as an independent integration test scenario rather than an embedded build task, it can be easily re-run once the source VSA is available. This is much easier than if the verification were embedded in the build pipeline, which would require a complete rebuild.
+
+- **Additional task required**: New verification task must be implemented and maintained (poc in [konflux-ci/build-definitions#2867](https://github.com/konflux-ci/build-definitions/pull/2867))
+- **OCI storage overhead**: Source VSAs are duplicated (git notes + OCI registry)
```
arewm: Is this really a negative?

ADR/####-slsa-source-provenance-verification.md
```diff
@@ -0,0 +1,130 @@
...
+   - How to generate source VSAs (GitHub Actions, GitLab CI templates)
+   - How to store VSAs as git notes
+   - What source verification levels mean
+
+4. **ITS Provisioning**: Update Konflux onboarding mechanisms to provision a source verification ITS automatically
+
+### Future Considerations
+
+- **Standard VSA format**: As SLSA source track specifications stabilize, we may need to update VSA formats
+- **Multiple source systems**: Different source control systems (GitHub, GitLab, Gitea) may need special support.
```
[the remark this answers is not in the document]
ralphbean: Yes.

ADR/####-slsa-source-provenance-verification.md
ralphbean: Maybe not.

ADR/####-slsa-source-provenance-verification.md
arewm: yeah, sounds good. Should we make this explicit? Or modify the ADR to state minimum criteria that would be needed for this to be further used... 

But the global deployment of this also doesn't need to happen with Konflux-CI itself. It could be done for a specific deployment of Konflux? Organizations would need to have this lever if we they would require this on the policy verification.

ADR/####-slsa-source-provenance-verification.md
arewm: We cannot use the tag-based fallback as I think that has all attestations in a single manifest so Chains would not be able to generate provenance for it.

But I think the last point is what I was saying as well. Conforma would need to be able to understand that this is a source VSA attestation and then verify that the keyid is the same as what is required from that source VSA attestation's provenance.

ADR/####-slsa-source-provenance-verification.md
```diff
@@ -0,0 +1,130 @@
...
+   - The task verifies the VSA contents (format, subject matching the commit, etc.)
+   - The task pushes the source VSA to the OCI registry as a separate artifact
+   - The task exposes the pullspec of the OCI representation of the VSA (digest reference) in its task results using `*_ARTIFACT_OUTPUTS` format
+
+3. **Build provenance chain** (via Tekton Chains):
+   - Tekton Chains observes the task results containing the source VSA artifact reference
+   - Chains generates and signs a SLSA build provenance attestation **about** the source provenance attestation
+   - This creates an attestation chain: `Build Provenance → Source VSA → Source Commit`
+
+4. **Verification in Conforma**:
```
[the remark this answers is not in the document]
ralphbean: Added a bullet point clarifying that Conforma verifies trust in the source verification task itself using the Trusted Tasks mechanisms described in ADR 53.

ADR/####-slsa-source-provenance-verification.md
```diff
@@ -0,0 +1,130 @@
...
+
+- **Race condition risk**: Source VSAs should be generated contemporaneously with commits (typically via GitHub Actions or GitLab pipelines), but Konflux's verification task may execute before the external automation completes. This can cause initial verification failures when:
+  - A commit is pushed to a repository
+  - Konflux immediately triggers a build pipeline
+  - The source VSA generation (GitHub Action/GitLab pipeline) is still running
+  - The verification task attempts to read git notes that don't exist yet
+
+  **Mitigation**: The movable test pattern from ADR 48 significantly helps with this issue. Because source verification runs as an independent integration test scenario rather than an embedded build task, it can be easily re-run once the source VSA is available. This is much easier than if the verification were embedded in the build pipeline, which would require a complete rebuild.
+
+- **Additional task required**: New verification task must be implemented and maintained (poc in [konflux-ci/build-definitions#2867](https://github.com/konflux-ci/build-definitions/pull/2867))
```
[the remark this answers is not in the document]
ralphbean: The task will be maintained in build-definitions for now, with the possibility of relocating it to a different repository in the future. Updated the text to reflect this.

ADR/####-slsa-source-provenance-verification.md
ralphbean: Updated the text to clarify that this is a capability for Konflux deployments to enable if they choose (e.g., for policy requirements), not a mandatory global rollout. Deployments wanting to adopt this can update their onboarding mechanisms accordingly.

ADR/####-slsa-source-provenance-verification.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Negative
+
+- **Race condition risk**: Source VSAs should be generated contemporaneously with commits (typically via GitHub Actions or GitLab pipelines), but Konflux's verification task may execute before the external automation completes. This can cause initial verification failures when:
+  - A commit is pushed to a repository
+  - Konflux immediately triggers a build pipeline
+  - The source VSA generation (GitHub Action/GitLab pipeline) is still running
+  - The verification task attempts to read git notes that don't exist yet
+
+  **Mitigation**: The movable test pattern from ADR 48 significantly helps with this issue. Because source verification runs as an independent integration test scenario rather than an embedded build task, it can be easily re-run once the source VSA is available. This is much easier than if the verification were embedded in the build pipeline, which would require a complete rebuild.
```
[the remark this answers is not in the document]
ralphbean: Added this timing consideration as an additional mitigation factor - while the race condition always occurs, typical build times reduce the likelihood of an unfavorable outcome.

ADR/####-slsa-source-provenance-verification.md
ralphbean: The current ADR approach (validating externally-generated source VSAs) addresses this correctly. No changes needed to the ADR text based on this discussion.

ADR/####-slsa-source-provenance-verification.md
```diff
@@ -0,0 +1,130 @@
...
+   - The task pushes the source VSA to the OCI registry as a separate artifact (manifest) using the OCI referrers API (via `oras attach`), with the container image as the subject of the OCI reference (the git commit remains the subject within the VSA payload itself)
+   - The task exposes the pullspec of the OCI representation of the VSA (digest reference) in its task results using `*_ARTIFACT_OUTPUTS` format
+
+3. **Build provenance chain** (via Tekton Chains):
+   - Tekton Chains observes the task results containing the source VSA artifact reference
+   - Chains generates and signs a SLSA build provenance attestation **about** the source provenance attestation
+   - This creates an attestation chain: `Build Provenance → Source VSA → Source Commit`
+
+4. **Verification in Conforma**:
+   - Conforma policies can verify source provenance by checking for the presence of the attestation chain (supporting both tag-based and referrers-based attestation storage)
```
ralphbean:
```suggestion
   - Conforma policies can verify source provenance by checking for the presence of the attestation chain (using the referrers-based attestation storage)
```

ADR/####-slsa-source-provenance-verification.md
```diff
@@ -0,0 +1,130 @@
...
+   - Fetches git notes from cloned repositories
+   - Verifies source VSA signatures against the provided public key
+   - Verifies source VSA contents (format, subject, etc.)
+   - Pushes VSAs to OCI registry
+   - Exposes pullspec in `*_ARTIFACT_OUTPUTS` results format
+
+2. **Conforma policy updates**: Update Enterprise Contract policies to:
+   - Detect source provenance attestation chains
+   - Verify build provenance attestations about source VSAs
+   - Read the public key from test attestation task parameters and verify it matches expected signer identities for source repositories
```
ralphbean:
```suggestion
   - Read the public key from test attestation task parameters and verify it matches expected signer identities for source repositories, provided to conforma
```

ADR/####-slsa-source-provenance-verification.md
```diff
@@ -0,0 +1,130 @@
...
+3. **Documentation for source repository owners**:
+   - How to generate source VSAs (GitHub Actions, GitLab CI templates)
+   - How to store VSAs as git notes
+   - What source verification levels mean
+
+4. **ITS Provisioning**: Konflux deployments that want to adopt source provenance verification across their instance can update their onboarding mechanisms to provision a source verification ITS automatically
+
+### Future Considerations
+
+- **Standard VSA format**: As SLSA source track specifications stabilize, we may need to update VSA formats
```
ralphbean:
```suggestion
- **Expected Signer Identities**: We may need to think more deeply about how to efficiently configure the expected signer identities at scale.
- **Standard VSA format**: As SLSA source track specifications stabilize, we may need to update VSA formats
```

arewm (approved):
Thanks for the clarifications and updates. Approved.

PR #274, opened 2025-10-23
Added ADR for proxy user configuration flags

ifireball:
Follow-up of ADR-0047, how users are going to be able to enable/disable
the caching proxy.

Assisted-by: Gemini (via Cursor)
Signed-off-by: Barak Korren <bkorren@redhat.com>

### DIFF
--- a/ADR/XXXX-pipeline-caching-feature-flag.md
+++ b/ADR/XXXX-pipeline-caching-feature-flag.md
@@ -30,7 +30,6 @@ To facilitate disabling use of the proxy throughout the cluster, a `allow-cache-
 
 * **Pipeline Integration**: The `enable-cache-proxy` parameter will be passed to the pipeline init task
 * **Configuration Resolution**: The init task will read configuration values from the `cluster-config` config map in the `konflux-info` namespace and resolve the final proxy settings
-* **Environment Variables**: The init task will emit `HTTP_PROXY` and `NO_PROXY` configuration values to be used by the buildah task
 * **Default Values**: If no configuration is found, the system will fall back emitting empty values, effectively switching the proxy off.
 * **Proxy Activation**: The proxy will be enabled if `enable-cache-proxy` is set to `true` and `allow-cache-proxy` is either set to `true` or unset.
 * **Emitted Values**: The init task will emit `HTTP_PROXY` and `NO_PROXY` configuration values be used by the buildah task, The emitted values for `HTTP_PROXY` and `NO_PROXY` will be read from the `cluster-config` config map, from the `http-proxy` and `no-proxy` keys respectively. If the keys are missing from the config map, the following default values would be had-coded in the init task:


amisstea (approved):

ADR/XXXX-pipeline-caching-feature-flag.md
```diff
@@ -0,0 +1,65 @@
+# XXXX. Pipeline Caching Feature Flag Configuration
+
+Date: 2025-01-27
```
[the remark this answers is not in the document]
ifireball: no that would be an AI hallucination... will fix.

ralphbean (approved):

ADR/XXXX-pipeline-caching-feature-flag.md
```diff
@@ -0,0 +1,65 @@
...
+
+The current design requires users to manually configure `HTTP_PROXY` and `NO_PROXY` parameters to the buildah task, which exposes implementation details and creates a poor user experience. We need a more user-friendly configuration mechanism that abstracts away the technical details while providing appropriate levels of control.
+
+## Decision
+
+We will implement a hierarchical feature flag system for controlling pipeline caching proxy usage with the following configuration levels and precedence:
+
+### Configuration Levels
+
+1. **Pipeline Level**: An `ENABLE_CACHE_PROXY` parameter that can be set in individual pipeline definitions
```
chmeliik: nitpick: pipeline parameters use `kebab-case`

ADR/XXXX-pipeline-caching-feature-flag.md
```diff
@@ -0,0 +1,65 @@
...
+2. **Namespace Level**: An `enable-cache-proxy` value in the `konflux-config` config map in the pipeline's namespace
+3. **Cluster Level**: An `enable-cache-proxy` value in the `cluster-config` config map in the `konflux-info` namespace
+
+### Precedence Order
+
+Configuration takes effect based on the following precedence (highest to lowest):
+
+1. Cluster-level configuration (`cluster-config` in `konflux-info` namespace)
+2. Namespace-level configuration (`konflux-config` in pipeline's namespace)
+3. Pipeline-level configuration (`ENABLE_CACHE_PROXY` parameter)
```
chmeliik: Why is this the precedence order? That seems backwards, I would expect the precedence to go `global < namespace < pipeline`

ADR/XXXX-pipeline-caching-feature-flag.md
```diff
@@ -0,0 +1,65 @@
...
+* **Configuration Values**: Each level supports `true`, `false`, `""` (empty), `"defer"`, or unset values
+* **Proxy Activation**: The proxy will be enabled if any of the configuration levels is set to `true`, following the precedence order above
+
+### Configuration Logic
+
+The proxy will be enabled if any of these conditions are met (in order of precedence):
+
+* Cluster flag is `true`
+* Cluster flag is unset/empty/deferred AND namespace flag is `true`
+* Cluster flag is unset/empty/deferred AND namespace flag is unset/empty/deferred AND pipeline parameter is `true`
```
chmeliik: If we go solely by the text of this ADR, then setting the flag to `false` doesn't do anything. One has to read between the lines to infer that a cluster-level `false` disables everything etc.

Maybe a simpler way to phrase this would be: "The proxy will be enabled if the first non-empty flag (in order of precedence) is `true`"

chmeliik: It may also be worth mentioning that they will default to `unset, unset, true` respectively (I assume)

ADR/XXXX-pipeline-caching-feature-flag.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Consequences
+
+* **Improved User Experience**: Users can control proxy usage through simple boolean flags without needing to understand proxy technical details like ports or addresses
+* **Flexible Control**: Three levels of configuration provide appropriate control for different user roles (pipeline authors, namespace administrators, cluster administrators)
+* **GitOps Integration**: Configuration can be managed through GitOps processes using config maps, consistent with other Konflux configuration elements
+* **Implementation Complexity**: The init task becomes more complex as it needs to resolve configuration from multiple sources and apply precedence rules
+* **Configuration Management**: Cluster and namespace administrators need to understand the precedence rules to effectively manage proxy usage
+* **Documentation Requirements**: Clear documentation will be needed to explain the configuration options and precedence rules for different user roles
+* **Migration Path**: The implementation can be done incrementally, starting with basic pipeline-level support and adding cluster/namespace configuration later to avoid breaking changes
```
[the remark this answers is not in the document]
MartinBasti: make it mandatory in conforma?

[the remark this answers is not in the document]
ifireball: Well having the proxy cache stuff for them might be a good "carrot" approach to convince them to bring it back (As opposed to the Conforma "stick" approach)

ADR/XXXX-pipeline-caching-feature-flag.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Implementation Details
+
+* **Pipeline Integration**: The `ENABLE_CACHE_PROXY` parameter will be passed to the pipeline init task
+* **Configuration Resolution**: The init task will read configuration values from the appropriate config maps and resolve the final proxy settings
+* **Environment Variables**: The init task will emit `HTTP_PROXY` and `NO_PROXY` configuration values to be used by the buildah task
+* **Default Values**: If no configuration is found, the system will fall back to default values:
+  * `HTTP_PROXY`: `squid.caching.svc.cluster.local:3128`
+  * `NO_PROXY`: *<TBD>*
+* **Configuration Values**: Each level supports `true`, `false`, `""` (empty), `"defer"`, or unset values
```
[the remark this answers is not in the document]
ifireball: well, I like "defer" because it describe what its doing most accurately and explicitly. Some people like to adhere to "Explicit is better then implicit", so I want to provide an option to do that.

Then again some people like to omit stuff to allow it to "fall through" so I want to allow for that as well (and it avoid needing to define defaults as edge cases)

ADR/XXXX-pipeline-caching-feature-flag.md
ifireball: I'll try to add some clarification, though I'd prefer to also keep the strictl logical definition in place as well. As for the defaults, those would be "unset" for all values which effectively mean the proxy is off by default (As per ADR 0047)

ADR/XXXX-pipeline-caching-feature-flag.md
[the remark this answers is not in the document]
ifireball: I think the ADR is short enough that I can afford to keep some space between them, but I'll state the motivation for this unintuitive precedence choice to appease the impatient ;)

ADR/XXXX-pipeline-caching-feature-flag.md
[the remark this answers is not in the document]
chmeliik: Yeah, I don't think enforcement is needed at this time. Just consider mentioning this in any announcement emails related to this functionality

ADR/XXXX-pipeline-caching-feature-flag.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Decision
+
+We will implement a hierarchical feature flag system for controlling pipeline caching proxy usage with the following configuration levels and precedence:
+
+### Configuration Levels
+
+1. **Pipeline Level**: An `enable-cache-proxy` parameter that can be set in individual pipeline definitions
+2. **Namespace Level**: An `enable-cache-proxy` value in the `konflux-config` config map in the pipeline's namespace
+3. **Cluster Level**: An `enable-cache-proxy` value in the `cluster-config` config map in the `konflux-info` namespace
```
mmorhun: Are those `*-config` Config Maps introduced here or they already exist?

ADR/XXXX-pipeline-caching-feature-flag.md
```diff
@@ -0,0 +1,79 @@
...
+
+### Precedence Order
+
+Configuration takes effect based on the following precedence (highest to lowest):
+
+1. Cluster-level configuration (`cluster-config` in `konflux-info` namespace)
+2. Namespace-level configuration (`konflux-config` in pipeline's namespace)
+3. Pipeline-level configuration (`ENABLE_CACHE_PROXY` parameter)
```
mmorhun: I think it will be very confusing to users if they have a pipeline parameter that says one thing, but actually does completely different stuff because different settings are used. We should avoid such behavior as much as possible.

ADR/XXXX-pipeline-caching-feature-flag.md
```diff
@@ -0,0 +1,79 @@
...
+leaving an unset or empty value (Or explicitly deferring) lets the lower levels
+decide.
+
+The default behavior is a consequence of having the value be unset in all
+levels. That results in the proxy being "off".
+
+
+## Consequences
+
+* **Improved User Experience**: Users can control proxy usage through simple boolean flags without needing to understand proxy technical details like ports or addresses
```
mmorhun: That's not really true, users might be forced to use proxy against their will.

ADR/XXXX-pipeline-caching-feature-flag.md
```diff
@@ -0,0 +1,79 @@
...
+* **Flexible Control**: Three levels of configuration provide appropriate control for different user roles (pipeline authors, namespace administrators, cluster administrators)
```
mmorhun: I disagree that this approach is flexible. Imagine situation when I want to use the proxy for all Components in my namespace except a few. For current configuration order, the only way is to specify the option in every pipeline.

ADR/XXXX-pipeline-caching-feature-flag.md
```diff
@@ -0,0 +1,79 @@
...
+1. Cluster-level configuration (`cluster-config` in `konflux-info` namespace)
+2. Namespace-level configuration (`konflux-config` in pipeline's namespace)
+3. Pipeline-level configuration (`ENABLE_CACHE_PROXY` parameter)
+
+The motivation for this seemingly unintuitive precedence order is to allow for
+overriding the per-pipeline configuration from the namespace or cluster level.
+This serves as a quick "emergency escape hatch" in case of proxy-based failure
+scenarios. See the "Configuration Logic" section below for details.
+
+### Implementation Details
```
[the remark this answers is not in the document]
ifireball: very well, if we don't get logging right once we implement this, let us know when you review the PR ;)  (I don't think this kind of detail really belongs in the ADR itself, logging should be good for everything).

ADR/XXXX-pipeline-caching-feature-flag.md
```diff
@@ -0,0 +1,79 @@
...
+The proxy will be enabled if any of these conditions are met (in order of precedence):
+
+* Cluster flag is `true`
+* Cluster flag is unset/empty/deferred AND namespace flag is `true`
+* Cluster flag is unset/empty/deferred AND namespace flag is unset/empty/deferred AND pipeline parameter is `true`
+
+What this means in practice is that setting the flag to `false` switches the
+proxy off regardless of the configuration in the lower precedence levels, while
+leaving an unset or empty value (Or explicitly deferring) lets the lower levels
+decide.
```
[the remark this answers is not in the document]
ifireball: it doesn't make sense to me to make it only work one way and not the other. 

if you compromised the cluster to the point that you can override the proxy settings, you can probably do as much harm by disabling it as you can do by enabling it. 

either way the wither the proxy settings will be recorded in the attestation data, and we can have Conforma or some other tool tell us if they are not what they are supposed to be (If you can trust the attestation data from a compromized cluster, that is).

ADR/XXXX-pipeline-caching-feature-flag.md
ifireball: right now, they have to know its address to use it. 

"Forced againset their will" - by the infra team you mean? The can already force anything they want. I doubt they will do anything like that, unless its needed to mitigate some greater issue.

ADR/XXXX-pipeline-caching-feature-flag.md
ifireball: you can always just not pass the init parameters to buildah for the "special" pipelines.

ADR/XXXX-pipeline-caching-feature-flag.md
MartinBasti: I'm already confused by this.

I may understand need of global turn-off of proxy, but  local namespace overriding pipeline config is too unexpected.

IMHO, we should have a separate admin flag in cluster config that is true/false/defer, and keep namespace with lower precedence than pipeline (and allow only true/false)

ADR/XXXX-pipeline-caching-feature-flag.md
```diff
@@ -0,0 +1,79 @@
...
+### Implementation Details
+
+* **Pipeline Integration**: The `ENABLE_CACHE_PROXY` parameter will be passed to the pipeline init task
+* **Configuration Resolution**: The init task will read configuration values from the appropriate config maps and resolve the final proxy settings
+* **Environment Variables**: The init task will emit `HTTP_PROXY` and `NO_PROXY` configuration values to be used by the buildah task
+* **Default Values**: If no configuration is found, the system will fall back emitting empty values, effectively switching the proxy off.
+* **Configuration Values**: Each level supports `true`, `false`, `""` (empty), `"defer"`, or unset values
+* **Proxy Activation**: The proxy will be enabled if any of the configuration levels is set to `true`, following the precedence order above. The emitted values will be:
+  * `HTTP_PROXY`: `squid.caching.svc.cluster.local:3128`
+  * `NO_PROXY`: *&lt;TBD&gt;*
```
MartinBasti: How NO_PROXY values will be defined? On cluster level? will we allow users to customize them?

You need to export also HTTPS_PROXY, because HTTP_PROXY is used only for HTTP traffic, which should be none.
Also variations of envvars with lowercase must be exported, so all tools are compatible.

ADR/XXXX-pipeline-caching-feature-flag.md
MartinBasti: How init tasks will get access to those config maps?

Will we update PLR to mount configmaps from tenant?

And the global configmap, we cannot mount it, are we going to add access to each SA to fetch it via k8s API?

ADR/XXXX-pipeline-caching-feature-flag.md
ifireball: I've added text to spell-out the distinction between the pipeline and task parameters and variables this ADR concerns itself with and the actual environment variable that are controlled by internal task code.

ADR/XXXX-pipeline-caching-feature-flag.md
ralphbean: I'll think this is satisifed now.

* The user elects to use the cache with a param on their pipeline.
* The admin can disallow use of the cache globally.

The user sets *enable* and *disable* (like requests) and the admin sets *allow* and *disallow* (like a limit or quota).

PR #279, opened 2025-10-31
Add google analytics for the architecture docs

ralphbean:
So we can try to understand the traffic.

Related: https://github.com/konflux-ci/docs/pull/501

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -6,6 +6,16 @@
   <title>{{ title or "Architecture Of Konflux" }}</title>
   <meta name="description" content="{{ description or 'Technical documents about the Managed Developer Platform' }}">
 
+  <!-- Google Analytics tracking code -->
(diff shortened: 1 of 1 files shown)


_layouts/base.html
```diff
@@ -6,6 +6,16 @@
   <title>{{ title or "Architecture Of Konflux" }}</title>
   <meta name="description" content="{{ description or 'Technical documents about the Managed Developer Platform' }}">
 
+  <!-- Google Analytics tracking code -->
+  <script async src="https://www.googletagmanager.com/gtag/js?id=G-381LYFWTP4"></script>
```
gbenhaim: Do we need to ask for consent from the user before running this?

ifireball: I think we do, and probably fill-in one of those three-letter forms about privacy protections.

PR #280, opened 2025-11-17
Introduce ADR: Schema-governed correlation ID for workflows

ci-operator:
This ADR introduces a schema-governed correlation ID for tracking workflows across multiple stages in Konflux, addressing the need for a deterministic identifier spanning multiple Tekton PipelineRuns. It outlines the generation and propagation of correlation IDs (ECID and SCID) without modifying existing CRDs, ensuring compatibility and observability.

### DIFF
--- a/ADR/0056-schema-goverened-correlation-id.md
+++ b/ADR/0056-schema-goverened-correlation-id.md
@@ -10,7 +10,7 @@ Konflux CI orchestrates multi-stage workflows involving Build PipelineRuns for i
 
 Tekton’s built-in identifiers (such as event IDs) are random and cannot be reproduced. CEL templates can extract event metadata but cannot compute stable hashes. PipelineRuns may be short-lived or created in remote clusters, preventing reliable correlation based on their presence in the cluster. Downstream observability systems that consume PipelineRun CloudEvents require a durable, deterministic identifier spanning the full Build -> Integration -> Release lifecycle.
 
-A correlation mechanism is needed that is deterministic across retries and replays, configurable on a per-namespace basis, SCM-agnostic, and capable of expressing both single-event and multi-event fan-in, all without modifying existing CRDs.
+A correlation mechanism is needed that is deterministic across retries and replays, configurable on a per-namespace basis, SCM-agnostic, and capable of expressing both single-event and multi-event fan-in, all without modifying existing CRDs. Because the correlation-id is a deterministic function of a configurable set of SCM fields (e.g. provider, repo, PR number, commit SHA), operators can define what constitutes a ‘unit of work’ (per webhook, per PR+commit, etc.) and use correlation-id as a stable join key in observability systems. This enables per-change or per-delivery latency metrics (e.g. commit -> first successful integration) without PaC needing to know anything about specific downstream controllers.
 
 ## Decision / Proposal
 


ADR/0056-schema-goverened-correlation-id.md
```diff
@@ -0,0 +1,57 @@
...
+
+**Event Correlation ID (ECID).**
+Generated by Pipelines-as-Code (PaC) per webhook event. Each namespace may define a `konflux-correlation-policy` Secret containing a UUIDv5 namespace, a selected set of normalized SCM fields (provider, repository, commit, PR number, event ID, etc.), and a canonical JSON template. PaC watches and caches this policy; for each webhook event, it normalizes the selected fields, renders the template, computes a UUIDv5 hash, and labels all PipelineRuns created for that event with the resulting ECID.
+
+**Snapshot Correlation ID (SCID).**
+Generated by Integration when a Snapshot contains components produced from multiple ECIDs. Build-service records the ECID in each Component’s build metadata when a Build PipelineRun completes. When Integration creates a Snapshot, it copies these ECIDs into the Snapshot’s component list. On reconciliation, if all components share the same ECID, that value becomes the Snapshot’s correlation ID; if multiple ECIDs are present, Integration computes a UUIDv5 over the sorted distinct ECIDs and uses it as the SCID. Parent ECIDs are stored in paged Snapshot annotations along with page count, total count, and a fingerprint.
```
arewm: > Build-service records the ECID in each Component’s build metadata when a Build PipelineRun completes.

What build metadata are you referring to? The build service doesn't propagate anything to resources on the cluster. It is actually the integration service which updates the Component CR with the latest known image build, but this is only so that it knows what references to use for the Snapshot generation.

arewm: > On reconciliation, if all components share the same ECID, that value becomes the Snapshot’s correlation ID; if multiple ECIDs are present, Integration computes a UUIDv5 over the sorted distinct ECIDs and uses it as the SCID. Parent ECIDs are stored in paged Snapshot annotations along with page count, total count, and a fingerprint.

The integration service has _some_ functionality for creating group snapshots for all components built from the same commit which feels similar to some of the desired functionality here. This is generally not data that the integration service has available to it since its only state is the set of latest Component build as recorded in the Component CR. 

How do you envision that the ECIDs would be stored in the Snapshot? We don't have a database available to ensure that we have all possible Snapshots available. If you are proposing to using existing CRs for this data, we are generally trying to reduce the amount of data stored in etcd across all controllers. This might be problematic for getting some of this data unless the integration service is aware of KubeArchive and can fetch data out of it.

ADR/0056-schema-goverened-correlation-id.md
```diff
@@ -0,0 +1,57 @@
...
+Konflux introduces a **schema-governed correlation identifier** carried through the system using a single label:
+
+```
+pipelinesascode.tekton.dev/correlation-id
+```
+
+Two identifiers are defined:
+
+**Event Correlation ID (ECID).**
+Generated by Pipelines-as-Code (PaC) per webhook event. Each namespace may define a `konflux-correlation-policy` Secret containing a UUIDv5 namespace, a selected set of normalized SCM fields (provider, repository, commit, PR number, event ID, etc.), and a canonical JSON template. PaC watches and caches this policy; for each webhook event, it normalizes the selected fields, renders the template, computes a UUIDv5 hash, and labels all PipelineRuns created for that event with the resulting ECID.
```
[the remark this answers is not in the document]
ci-operator: Good call – setting this in the Repository spec (e.g. spec.correlationPolicy.secretRef.name) is the right place, even if it breaks our “no CRD modifications” guarantee 😅. That also lets each Repository pick its own correlation policy, which is useful both for mixed SCM providers (different webhook payload shapes) and for teams that want different correlation semantics (e.g. per-webhook vs per-PR+commit) even within the same namespace.

If that field isn’t set, PaC simply doesn’t label PipelineRuns with ECIDs for that repository. If it is set but the Secret is missing or invalid, PaC treats correlation as disabled for that repo and logs a warning rather than silently falling back to a default policy.  Such implementation-specifics may be detailed a bit more in the PaC-specific [discussion](https://github.com/openshift-pipelines/pipelines-as-code/discussions/2333).

Longer term, I imagine PaC could support a cluster-level default policy that applies only when the Repository doesn’t specify one and correlation is enabled globally, but that feels additive and not required for a first iteration.

ADR/0056-schema-goverened-correlation-id.md
ci-operator: The intent is that Integration still only uses the state it has today: the latest build info on `Component` plus the current `Snapshot`. When a build `PipelineRun` completes, Integration reads the ECID from the `PipelineRun` and writes it as an annotation on the corresponding `Component`; when a `Snapshot` is reconciled, it looks up those Components, reads their ECIDs, computes the distinct ECID set for *that Snapshot only*, and either reuses the single ECID or derives an SCID and writes the SCID plus the parent ECID list as paged annotations on that Snapshot. There’s no extra database, no need to discover “all Snapshots”, and no KubeArchive lookup in the reconcile path.

On the etcd side, this is bounded to one correlation label per object and parent ECID annotations only on multi-origin Snapshots and Release PipelineRuns, with paging over the distinct ECIDs to stay under a configurable size cap. KubeArchive is still useful for long-term archival, but for this we need correlation metadata on live CRs and `PipelineRun` events so we can do near-real-time aggregation and alerting, which makes depending on an eventually consistent archive in the reconcile loop a non-starter.

ADR/0056-schema-goverened-correlation-id.md
```diff
@@ -0,0 +1,55 @@
+# ADR-0056: Schema-Governed Correlation ID for Workflow Traceability
```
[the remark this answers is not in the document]
raks-tt: @ggallen - Can you explain what do you mean by nudging?

PR #281, opened 2025-11-19
Add ADR for ComponentGroups

14rcole:
Relates to #277

### DIFF
--- a/ADR/0060-component-groups.md
+++ b/ADR/0060-component-groups.md
@@ -1,0 +1,282 @@
+# 60. ComponentGroups
+
+Date: 2025-11019
+
+## Status
+
+Proposed
+
+## References
+To be implemented alongside [ADR-0056](./0056-revised-component-model.md)
+
+## Context
+
+The existing application/component model in Konflux limits our ability to
+support different types of workflows. New versions of a particular Component
+must be onboarded as a separate Component. Additionally, each Component can
+belong to exactly one application. Both of these issues hinder reusability and
+introduce additional work for users.
+
+Furthermore, ownership of the current APIs is unclear. They were originally
+owned by HAS (which is now deprecated) and exhibit excessive coupling across
+multiple services.
+
+The current application/component model also limits the integration service's
+ability to implement test ordering. Because the current model does not
+facilitate having a central place from which to define the test ordering, both
+the user experience and the performance of test ordering would be poor. We
+must make sure that the new Custom Resources support test ordering.
+
+Users also want testing on one snapshot to trigger the creation and testing of
+other snapshots with different groups of components. This too must be
+accounted for in the new model.
+
+Finally, Snapshot creation is hard-coded into the integration service.
+Snapshots are always created from the full set of Components in an Application.
+This is limiting for users with more complex test and release use-cases who may
+want to aggregate subsets of these sets of Components for testing. While this
+problem is out of the scope of the current design, we need to consider it when
+desigining the ComponentGroup spec.
+
+## Decision
+
+We have decided to _deprecate_ and eventually _remove_ the Application CR and
+replace it with a new CR called ComponentGroup.
+
+The ComponentGroup spec will contain a Component/Branch list which will tell
+the integration service how to generate snapshots. This is the reverse of how
+the relationship is defined in the old model, in which the Component CR
+contained a `spec.Application` field naming its parent Application. Defining
+the relationship "top-down" from parent to child will allow Components to
+belong to multiple ComponentGroups. To enable test ordering the ComponentGroup
+spec will also include an optional field called `testGraph` which will contain
+a map thatdefines the order of IntegrationTestScenarios to be run. Each node in
+the map will be the name of an IntegrationTestScenario and will contain a list
+of other IntegrationTestScenarios. The scenario named in the node runs after
+those in the list below it have completed. Each item in the list will contain
+the name of another IntegrationTestScenario (required) and whether the parent
+scenario should still be run if that scenario fails (optional). Finally, the
+ComponentGroup spec will contain a list called `dependents` that contains the
+names of other ComponentGroups for which snapshots should be created when a
+snapshot is created for the parent ComponentGroup.
+
+### Example of ComponentGroup Spec
+
+```
+apiVersion: konflux-ci.dev/v1alpha1
+kind: ComponentGroup
+metadata:
+  name: sample-cg
+  namespace: default
+spec:
+  components:
+    - name: first-component
+      componentVersion:
+        name: "testing"
+    - name: second-component
+      componentVersion:
+        name: v1
+    - name: python-component
+      componentVersion:
+        name: "3.12.4"
+  dependents:
+    - child-cg
+  testGraph:
+    verify:
+      - name: clamav-scan
+      - name: dast-tests
+        onFail: skip
+      - name: operator-scorecard
+    e2e-test:
+      - name: clamav-scan
+        onFail: run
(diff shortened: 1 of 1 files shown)


ralphbean:
@14rcole when you're ready for review to start, announce this on the community call, per [the process](https://github.com/konflux-ci/community/blob/main/ADRs.md).

ADR/0057-component-groups.md
```diff
@@ -0,0 +1,211 @@
...
+
+* **spec.components[].componentBranch.name**
+  * Type: string
+  * Required field
+  * The name of the branch
+
+* **spec.components[].componentBranch.lastPromotedImage**
+  * Type: string
+  * Optional field
+  * The name of the last image that was promoted to the GCL. Updated by the integration service and used to create snapshots. Cannot be updated by the user
```
tkdchen: What is GCL?

ADR/0057-component-groups.md
```diff
@@ -0,0 +1,211 @@
...
+kind: ComponentGroup
+metadata:
+  name: sample-cg
+  namespace: default
+spec:
+  components:
+    - name: first-component
+      componentBranch:
+        name: "testing"
+        lastPromotedImage: "quay.io/sampleorg/first-component@sha256:1b29..."
```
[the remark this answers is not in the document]
adambkaplan: Learn the lesson from ImageStreams - do not put information that updates as part of a controller reconciliation loop in `spec`. Otherwise you cannot use GitOps tools to manage ComponentGroups.

ADR/0057-component-groups.md
```diff
@@ -0,0 +1,211 @@
...
+
+* **spec.testGraph.{}.onFail**
+  * Type: string
+  * Optional field
+  * Denotes whether a failure in the IntegrationTestScenario should prevent the next scenario from running. Can be set to `run` or `skip`. Defaults to `run`
+
+* **spec.snapshotCreator**
+  * Type: object
+  * Optional field
+  * Placeholder field. In the future this will be the home of custom snapshot creation logic if the user wishes to define any.
```
adambkaplan: Per @lcarva 's recommendation - drop this and add to an "Alternatives Considered" section. This is something that can be considered as a follow-up.

ADR/0057-component-groups.md
```diff
@@ -0,0 +1,211 @@
...
+Users also want testing on one snapshot to trigger the creation and testing of
+other snapshots with different groups of components. This too must be
+accounted for in the new model.
+
+Finally, Snapshot creation is hard-coded into the integration service.
+Snapshots are always created from the full set of Components in an Application.
+This is limiting for users with more complex test and release use-cases who may
+want to aggregate subsets of these sets of Components for testing. While this
+problem is out of the scope of the current design, we need to consider it when
+desigining the ComponentGroup spec.
```
adambkaplan: I'm a bit confused - from what I understand, this ADR effectively adds a secondary means of creating `Snapshots`. It is intended to replace `Applications`, however we can't just drop it.

ADR/0057-component-groups.md
```diff
@@ -0,0 +1,211 @@
...
+
+## Decision
+
+We have decided to do away with the Application CR and replace it with a new CR
+called ComponentGroup. The ComponentGroup spec will contain a Component/Branch
```
adambkaplan: We have community members operating Konflux at scale - simply dropping `Applications` and replacing them with `ComponentGroup` will not be sufficient.

Our plan of action must be to _deprecate_ and _remove_ the `Application` custom resource and respective controllers. We also need a plan for migrating current `Application` users to `ComponentGroup` with either no loss of functionality, or well documented impacts and workarounds. Otherwise I don't think we graduate this ADR to "implementable".

ADR/0057-component-groups.md
```diff
@@ -0,0 +1,211 @@
...
+    - name python-component
+      componentBranch:
+        name: "3.12.4"
+        lastPromotedImage: ''
+        lastPromotedCommit: ''
+        lastBuildTime: ''
+  dependents:
+    - child-cg
+  testGraph:
+    verify:
```
adambkaplan: I like the idea of the `testGraph`, but do not like this particular API implementation for two reasons:

- The API uses map keys to imply a reference to an `IntegrationTestScenario`. This locks the API to a specific custom resource implementation that we may very well want to change. As a matter of an academic exercise, why not allow references to Tekton pipelines directly?
- The direction of the DAG does not match how a person naturally reads the data structure.

Potential alternatives for inspiration (that are also "battle tested"):

- Tekton's [runAfter](https://tekton.dev/docs/pipelines/pipelines/#using-the-runafter-field)
- Argo Workflows [DAG model](https://argo-workflows.readthedocs.io/en/latest/walk-through/dag/)

ADR/0057-component-groups.md
```diff
@@ -0,0 +1,211 @@
...
+        lastBuiLdTime: "2025-08-13T05:13:25Z"
+    - name python-component
+      componentBranch:
+        name: "3.12.4"
+        lastPromotedImage: ''
+        lastPromotedCommit: ''
+        lastBuildTime: ''
+  dependents:
+    - child-cg
+  testGraph:
```
adambkaplan: I cannot help but think - is this a problem that only exists because Tekton does not support [nested pipelines](https://tekton.dev/docs/pipelines/pipelines-in-pipelines/)?

cc @vdemeester

adambkaplan (commented):
This ADR appears to be biting off a lot by proposing the removal of the `Application` CR. I am in favor of stating this as a long term goal, however its role is so critical to Konflux's software model that I don't think we can drop it completely in one go.

I would rather that this ADR and the `ComponentGroup` API narrow its scope to the following:

- Snapshot creation based on more loosely coupled `Components`
- Orchestration of `IntegrationTestScenarios`

I think this will provide immediate value that addresses the primary gaps in our current App/Component model.

I would also like to see - as a matter of a thought exercise - an alternative API that has `Component` objects reference one or more `ComponentGroups`. This is closer to the existing App/Component model and is akin to how Kubernetes defines parent/child relationships (OwnerReferences).

ADR/0058-component-groups.md
```diff
@@ -0,0 +1,211 @@
...
+Snapshots are always created from the full set of Components in an Application.
+This is limiting for users with more complex test and release use-cases who may
+want to aggregate subsets of these sets of Components for testing. While this
+problem is out of the scope of the current design, we need to consider it when
+desigining the ComponentGroup spec.
+
+## Decision
+
+We have decided to do away with the Application CR and replace it with a new CR
+called ComponentGroup. The ComponentGroup spec will contain a Component/Branch
```
ralphbean: From the @konflux-ci/kgc discussion today, we agree that the intent to deprecate
 and then remove the Application CR must be made clear in the ADR, but we don't think that details of the migration need to be documented here. We're okay leaving that to the discretion of downstream operators of the Konflux platform.

ADR/0058-component-groups.md
```diff
@@ -0,0 +1,211 @@
...
+* **spec.snapshotCreator**
+  * Type: object
+  * Optional field
+  * Placeholder field. In the future this will be the home of custom snapshot creation logic if the user wishes to define any.
+
+## Consequences
+
+Once this is implemented users will be able to more easily test different
+configurations of their applications. Duplication of applications and
+components will be significantly reduced. There will still be some duplication
```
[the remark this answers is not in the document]
ralphbean: Yes, I think that moved forwards in https://github.com/konflux-ci/architecture/pull/277

PR #283, opened 2025-11-24
Official addons

gbenhaim:
1. Document the reasons why Konflux chose to support the Add-on approach.
2. Create a separation between official/unofficial addons.

If this change will be accepted I will update https://github.com/konflux-ci/community/blob/main/ADRs.md as well.

Assisted-By: Cursor


gbenhaim:
> Were you prompted to clarify this after seeing #282 in the review queue?

Actually now. I saw it after I've opened this pr. I'm glad that #282 was opened and I suggest to repeat the process for all add-ons.

arewm (approved):

arewm:
Is there any guidance for what should be documented in this architecture repository? Which should be enabled for configuration in konfux-ci/konflux-ci?

PR #303, opened 2026-01-14
ADR: Generalize Build and Release Pipelines

adambkaplan:
This decision record generalizes the patterns and practices of container image build and release on Konflux, and generalizes them to any software package artifact. The document uses SLSA terminology to define the meanings of "artifacts" and "packages" to provide consistent understanding. An additional term - "package identity" - is introduced to address a gap in the current SLSA v1.2 vocabulary. The core of the decision record declares that OCI images must be used to _store_ build artifacts, which are then _distributed_ using the Konflux release service.

A core challenge adapting Konflux to non-container software artifacts is addressing the storage and distribution of built outputs. For container images, the storage and distribution are handled by the same entity - the container registry. With non-container software, we need a reliable way to store software artifacts such that downstream consumers can reliably prove the software was not tampered with. This document assumes that no such system exists outside of the OCI standard because OCI images are guaranteed to be identified by their content-addressable digest. Distribution is then handled by the release pipeline, which is able to publish the artifacts to a destination package registry.

A secondary challenge is that in many non-container ecosystems, a single "build" can produce multiple package artifacts. This ADR addresses this scenario by defining the "Component Build Artifact", the output of a Konflux build that contains all software artifacts that are intended to be distributed. The release pipeline is responsible for distributing the artifacts within the Component Build Artifact, and as such is also responsible for documenting the identities of the released package artifacts. This ensures the full set of artifacts released by Konflux are able to be documented.

Current known Konflux build pipelines meet the requirements specified in this ADR. Release pipelines likewise meet the general requirements in this ADR, with the exception of adding type checking for component OCI images contained in the provided Snapshot object. These checks are considered best practice and are documented as a recommended improvement.

### DIFF
--- a/ADR/0063-generalized-build-release-pipelines.md
+++ b/ADR/0063-generalized-build-release-pipelines.md
@@ -1,0 +1,358 @@
+---
+title: "63. Generalized Build and Release Pipelines"
+status: Implementable
+applies_to:
+  - pipeline-service
+  - build-service
+  - integration-service
+  - release-service
+topics:
+  - build
+  - rpm
+  - python
+  - java
+  - release
+---
+
+# 63. Generalized Build and Release Pipelines
+
+Date: 2026-03-17
+
+## Status
+
+Implementable
+
+## Context
+
+Konflux was originally designed to produce container images. As a result, all of Konflux's core
+software modeling APIs (`Application`, `Component`, `Snapshot`) assume a container image is used
+to _store_ and _distribute_ software artifacts. Dependencies between components were assumed
+to be represented as container image references in source code. Container image standards (OCI) are
+also used to associate secure software supply chain metadata (signatures, attestations, and SBOMs)
+with executable code.
+
+Over the past year, the Konflux community has added build and release pipelines for non-container
+software, such as [RPMs](https://github.com/konflux-ci/rpmbuild-pipeline) and [Python wheels](https://github.com/calungaproject/plumbing).
+It is clear that Konflux aspires to be a _universal_ secure software factory, and therefore should
+support additional software ecosystems such as npm (NodeJS) and Maven (Java).
+
+This ADR codifies and generalizes the patterns for building and releasing non-container software.
+
+### Definitions
+
+For the remainder of this ADR, the term _"OCI image"_ will refer to any artifact that meets the OCI
+[image specification](https://specs.opencontainers.org/image-spec/). This can refer to runnable
+images, [index images](https://specs.opencontainers.org/image-spec/image-index/?v=v1.1.1), or
+non-runnable [OCI artifacts](https://specs.opencontainers.org/image-spec/manifest/?v=v1.1.1#guidelines-for-artifact-usage).
+
+The terms "artifact" and "package" will use the [SLSA v1.2 definition of these terms](https://specs.opencontainers.org/image-spec/manifest/?v=v1.1.1#guidelines-for-artifact-usage),
+namely:
+- An "artifact" is any immutable blob of data, whether that be a single file, a git commit, or a
+  container image.
+- A "package artifact" is an artifact that is _distributed_ as an output of a build process.
+- A "package name" is the primary identifier for a _mutable_ collection of artifacts used to obtain
+  software. More than one package artifact can be referenced or implied by a package name.
+- A "package registry" is an entity that maps package names and identities to package artifacts.
+
+This ADR will also introduce the term "package identity", with the following definition:
+
+```
+A "package identity" is a reference to a specific package artifact within or belonging to a package
+name. This is often known as a "package version" within package ecosystems, however:
+
+- The format and requirements of such identity can vary across package ecosystems. SemVer v2 format
+  is often encouraged, but not always required.
+- A package artifact may have one or more package identities at a given point in time, subject to
+  the requirements of the artifact's package ecosystem.
+- Package identities are not required to be immutable, depending on the requirements of its related
+  package ecosystem.
+```
+
+"Software" may be used as a preceding adjective for the above terms to distinguish artifacts
+containing executable code from supplementary metadata artifacts, such as Software Bill of
+Materials (SBOMs), signatures, and security scan reports.
+
+## Decision
+
+- The definition of a `Component` (its current form, or successor as defined in [ADR-56](./0056-revised-component-model.md))
+  is updated as follows:
+
+  ```
+  A "Component" is a set of one or more software package artifacts that are meant to be assembled
+  and distributed through a single action. This set of package artifacts is stored as an OCI image,
+  and is referred to as the _Component Build Artifact_.
+  ```
+
+- The definition of a `Release` is updated as follows:
+
+  ```
+  A "Release" is the process of verifying and distributing the software package artifacts within an
+  `Application` or `ComponentGroup`, which in turn are comprised of one or more Components and
+  their related Component Build Artifacts. A successful "Release" publishes all identified software
+  package artifacts for an `Application` or `ComponentGroup` to a destination package registry.
+  ```
+
+- All Konflux build pipelines MUST produce an OCI image that _stores_ all content that is intended
+  to be _distributed_ through a single action (the "Component Build Artifact"). The particulars of
+  the OCI image can vary as needed, for example:
+  - A containerized multi-arch application build produces an OCI image _index_ which contains
+    references to architecture-specific OCI container images.
+  - An no-arch RPM build produces an OCI artifact with a single layer containing the compressed
+    (tar+gzip) contents of the RPM package.
+
+- The _Component Build Artifact_ MUST be stored on a container registry that
+  enables the [OCI Distribution Referrers API](https://github.com/opencontainers/distribution-spec/blob/v1.1.1/spec.md#listing-referrers).
+
+- The _Component Build Artifact_ MUST provide sufficient metadata that allows its contents to be
+  analyzed and extracted, namely:
+  - Provide a `mediaType` of either `application/vnd.oci.image.index.v1+json` (Image Index) or
+    `application/vnd.oci.image.manifest.v1+json` (Image Manifest)
+  - If the _Component Build Artifact_ is an Image Manifest, provide one of the following:
+    - A value for `artifactType`, OR
+    - A value for `config.mediaType` that is not set to the [empty value](https://github.com/opencontainers/image-spec/blob/main/manifest.md#guidance-for-an-empty-descriptor)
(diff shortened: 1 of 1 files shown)


ADR/0059-generalized-build-release-pipeliness.md
```diff
@@ -0,0 +1,231 @@
...
+containing executable code from supplementary metadata artifacts, such as Software Bill of
+Materials (SBOMs), signatures, and security scan reports.
+
+## Decision
+
+- The definition of a `Component` (its current form, or successor as defined in [ADR-56](./0056-revised-component-model.md))
+  is updated as follows:
+
+  ```
+  A "Component" is a set of one or more software package artifacts that are meant to be assembled
```
gbenhaim: 1. So IIUC, the purpose of this adr is to make it official that a build pipeline can produce multiple artifacts which are referenced by an index and the index would be the output of the build pipeline?

2. I thing that a a word about component groups (or as as we call them today application) should be added since their purpose is to also distribute multiple artifacts together.

ADR/0059-generalized-build-release-pipeliness.md
```diff
@@ -0,0 +1,231 @@
...
+  and distributed through a single action. This set of package artifacts is stored as an OCI image,
+  and is referred to as the _Component Build Artifact_.
+  ```
+
+- The definition of a `Release` is updated as follows:
+
+  ```
+  A "Release" is the process of verifying the software package artifacts within one or more
+  Component Build Artifacts, and upon verification publishing these software package artifacts to a
```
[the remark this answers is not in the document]
adambkaplan: I would consider (additional) signing an implementation detail of the "publication" process.

ADR/0059-generalized-build-release-pipeliness.md
```diff
@@ -0,0 +1,258 @@
...
+  - If the _Component Build Artifact_ is an Image Manifest, provide one of the following:
+    - A value for `artifactType`, OR
+    - A value for `config.mediaType` that is not set to the [empty value](https://github.com/opencontainers/image-spec/blob/main/manifest.md#guidance-for-an-empty-descriptor)
+  - If the `artifactType` value is set, it SHOULD use a recognized OCI MIME type for the artifact's
+    package ecosystem (ex: runnable container images, Helm charts). If no MIME type for OCI exists
+    within the package ecosystem, the build pipeline can set an arbitrary media type value that
+    uses the `vnd.konflux` namespace prefix.
+
+- The build pipeline MUST document the expected `mediaType`, `artifactType`, and/or
+  `config.mediaType` for the _Component Build Artifact_.
```
adambkaplan: @bmbouter I have added this section here to clarify the role of the component build artifact, and the requirements with respect to using OCI with package ecosystems that are reluctant to adopt it.

The downside of this approach is that to get software in the hands of developers in these non-OCI package ecosystems, it needs to be released. Continuous release is possible in Konflux, and I document a potential workflow for active/"pre-release" development later in this ADR. What I don't know is how widely this approach is used by current Konflux users, if at all.

ralphbean (changes requested):
Please:

* Move the source container bullet up to treat it like any other supplementary artifact.
* Remove the suggestion around storing policies alongside pipeline definitions. It's a cool idea, but we don't know enough about what it looks like to include it here confidently.

With that, I think we're :+1: to merge soon.

bmbouter:
Thank you for this thorough ADR—I appreciate the work that went into codifying these patterns as Konflux evolves toward supporting broader software ecosystems. I particularly like the framing around the Component Build Artifact concept and how it addresses the real challenge of packaging systems that produce multiple artifacts from a single build. That's a solid solution.

However, I have a significant concern about one of the core assumptions in the proposal that I'd like to raise before this is ratified.

**The assumption that OCI images are the only content-addressable storage mechanism doesn't align with the current state of artifact repositories.** Multiple mature systems already implement content-addressable storage with cryptographic integrity verification:

- **Artifactory** explicitly stores binaries by checksum, providing deduplication and tamper-resistant references
- **Go modules** uses SHA-256 hashes in `go.sum` with immutable, content-addressable semantics
- **Maven ecosystems** verify artifacts through stored checksums
- **npm, pip, and Ruby Gems** all maintain hash-verified package integrity
- **GitHub Artifacts** provides SHA-256 digest verification

I ran through a quick survey of these (sources: [Artifactory](https://jfrog.com/blog/database-wrong-fud/), [Go modules](https://go.googlesource.com/proposal/+/master/design/25530-sumdb.md), [npm](https://medium.com/node-js-cybersecurity/lockfile-poisoning-and-how-hashes-verify-integrity-in-node-js-lockfiles-0f105a6a18cd), [pip](https://peps.python.org/pep-0458/), [Bundler](https://mensfeld.pl/2025/01/the-silent-guardian-why-bundler-checksums-are-a-game-changer-for-your-applications/)), and all of them support content-addressable artifact storage.

**What I think this ADR is actually solving** (and does well):
- Unified supply chain metadata (signatures, attestations, SBOMs via OCI referrers)
- Consistent provenance tracking across heterogeneous build types
- A single interface for downstream consumers to understand artifacts

**What I'd suggest reconsidering:**
Rather than arguing that OCI is the *only* option for tamper-proof storage, frame this as an architectural choice for *unified provenance and attestation tracking*. That's a completely legitimate reason to wrap artifacts in OCI format—many Konflux users will want that unified supply chain metadata. But it's not because other systems lack content-addressable storage.

The difference matters because:
1. It acknowledges that wrapping (e.g., RPMs in OCI artifacts) adds overhead and complexity
2. It opens the door for future alternative approaches (e.g., some builds might opt for native package registry publication with equivalent provenance metadata)
3. It's technically more defensible if someone asks why we're not using their ecosystem's native storage

I'm not suggesting this changes the decision—unified metadata tracking is valuable. But I think the justification should be more precise about what problem OCI actually solves here versus what other systems already provide.

ADR/0062-generalized-build-release-pipeliness.md
```diff
@@ -0,0 +1,342 @@
...
+  garbage collected. As above, commercial package registries _may_ provide this as a feature in
+  practice.
+- The build pipeline would need to generate a list of artifacts published to the development
+  registry in a machine-readable format.
+
+There are multiple benefits to this approach:
+
+- "Development" artifacts are easier to consume natively in dependent `Components`.
+- Component "nudging" through Renovate can be implemented with native tooling, rather than obscure
+  OCI artifacts or a scheduled Renovate run that obtains the latest "released" artifact.
```
adambkaplan: This touches on the primary benefits @bmbouter raised (and conversely the largest drawback of this ADR's approach).

adambkaplan (commented):
I added additional consequences for release pipelines and significantly expanded the "native package registry" alternative based on @bmbouter 's feedback.

Publishing "candidate builds" to candidate package registries is not a "hard no" for me. There are a few problem areas that I have identified which _may_ be resolved in specific package ecosystems, but are harder to assess more broadly. The chosen direction in this ADR prioritizes "works everywhere with our existing known build pipelines and tools" over developer experience.

If the community ends up hating this, we have a well documented "Plan B" that we can pivot to.

ralphbean (approved):

ADR/0063-generalized-build-release-pipeliness.md
```diff
@@ -0,0 +1,342 @@
...
+  - Pre-release artifacts MUST have at least one immutable package identifier. These
+    identifiers SHOULD be generated as part of the build process.
+  - The format of these identifiers and mechanisms to enforce immutability MAY be provided by the
+    package registry and/or package ecosystem tooling.
+
+- Current known build pipelines for container images, RPMs, and Python wheels meet the metadata
+  requirements for _Component Build Artifacts_. Future build pipelines for non-container artifacts
+  must adhere to these requirements.
+
+- Release pipelines for a given component MUST verify that they are capable of releasing the
```
johnbieren: So if a release pipeline has 20 tasks in it, we have to check we have all required data for each task before we do anything? That is going to be very ugly

johnbieren (approved):
lgtm, thanks for explanations @adambkaplan 
nit - you have a typo in the ADR file name (pipeliness)

cit1zen (approved):

ralphbean (approved):

arewm (approved):
Thanks for this. Just a couple changes to indicate that we should have documentation for build tasks or pipelines.

ralphbean (approved):

ADR/0063-generalized-build-release-pipelines.md
```diff
@@ -0,0 +1,358 @@
...
+- Release pipelines for non-container artifacts MUST extract content from the provided _Component Build_
+  _Artifact_ and publish these to appropriate "native" package registries. The release pipeline
+  MUST publish this content in the manner expected of the respective package ecosystem. The set of
+  artifacts published during the release process MUST be documented on the associated Konflux
+  `Release` object.
+
+- Current release pipelines update the `Release` object's `status.artifacts` field in-place to
+  document the set of released artifacts. This practice can remain for the time being.
+
+## Alternatives Considered
```
bmbouter, line 187: Apologies for being late to this. Can I get some clarity here on the state of these artifacts being available in native package managers? This section is in the alternatives category, does that mean this is preferred/allowed/disallowed?

I can understand not making it required by ADR policy, but in practice it kind of is... Without built artifacts being available in language native package managers, this ADR is incompatible with the very stakeholders it mentions here as motivations, e.g. Calunga (requires python pip support), RHEL on Konflux (requires rpm metadata and dnf support).

I'm probably reading it wrong. Please set me straight.

@adambkaplan @ralphbean @arewm

bmbouter (commented):
followup question -- sorry!

ADR/0063-generalized-build-release-pipelines.md
arewm: The ADR isn't specific about it. It is possible that the artifacts can also be available from other sources, but we need to have them as OCI artifacts as that is what we pin the metadata to.

adambkaplan: There's nothing that prevents a build pipeline from adding a step that publishes to a native package registry. However this results in content being duplicated, and other parts of the Konflux test + release workflows will not source content from said package registry.

The "automatic tenant release" [workaround](https://github.com/konflux-ci/architecture/blob/main/ADR/0063-generalized-build-release-pipelines.md?plain=1#L267-L275) tries to alleviate this concern. This would allow vetted code to be published to a "dev" package registry in a more rapid fashion, and allow downstream components to consume these "dev" builds in a native fashion.

PR #305, opened 2026-02-03
docs: add ADR for mandating conventional commits across Konflux repos

ergonzal22:
Proposes adopting the Conventional Commits specification to enable automated changelog generation for Konflux-CI Installer releases.

Ref: KFLUXVNGD-728

Assisted-by: Claude Code

### DIFF
--- a/ADR/NNNN-conventional-commits.md
+++ b/ADR/NNNN-conventional-commits.md
@@ -1,0 +1,102 @@
+# N. Mandate Conventional Commits Across Konflux Repos
+
+Date: 2026-02-03
+
+## Status
+
+Proposed
+
+## Context
+
+Konflux currently lacks a standardized commit message format across its repositories. This inconsistency creates several challenges:
+
+* **Manual changelog generation**: Release notes must be compiled manually by reviewing commit histories, which is time-consuming and error-prone
+* **Inconsistent commit quality**: Without guidelines, commit messages vary widely in format and informativeness
+* **Difficulty tracking changes**: Understanding what changed between releases requires reading through unstructured commit messages
+* **Limited automation opportunities**: Tools that could automate versioning and changelog generation cannot be used effectively
+
+The [Conventional Commits](https://www.conventionalcommits.org/) specification provides a lightweight convention for commit messages that enables automated tooling while remaining human-readable. It structures commits as:
+
+```
+<type>[optional scope]: <description>
+
+[optional body]
+
+[optional footer(s)]
+```
+
+Where `type` indicates the nature of the change (e.g., `feat`, `fix`, `docs`, `chore`).
+
+## Decision
+
+All Konflux repositories will adopt the Conventional Commits specification for commit messages. This decision includes:
+
+### Commit Message Format
+
+Commits must follow the Conventional Commits 1.0.0 specification with the following types:
+
+| Type | Description |
+|------|-------------|
+| `feat` | A new feature |
+| `fix` | A bug fix |
+| `docs` | Documentation only changes |
+| `style` | Changes that do not affect the meaning of the code (formatting, etc.) |
+| `refactor` | A code change that neither fixes a bug nor adds a feature |
+| `perf` | A code change that improves performance |
+| `test` | Adding missing tests or correcting existing tests |
+| `build` | Changes that affect the build system or external dependencies |
+| `ci` | Changes to CI configuration files and scripts |
+| `chore` | Other changes that don't modify src or test files |
+| `revert` | Reverts a previous commit |
+
+Breaking changes must be indicated by appending `!` after the type/scope or by including `BREAKING CHANGE:` in the footer.
+
+### Enforcement Mechanism
+
+Repositories will enforce conventional commits through CI checks using [commitlint](https://commitlint.js.org/) with the `@commitlint/config-conventional` preset. The enforcement will be implemented as:
+
+1. **PR title validation**: GitHub Actions workflow that validates PR titles follow conventional commit format (for squash-merge workflows)
+2. **Commit message validation**: For repositories using merge commits, individual commit messages will be validated
+3. **Pre-commit hooks**: Optional local enforcement via Husky and commitlint for developer convenience
+
+### Rollout Timeline
+
+The rollout will proceed in phases:
+
+1. **Phase 1 - Documentation and Tooling**: Publish adoption guide, create reusable GitHub Actions workflow, document PR template updates
+2. **Phase 2 - Opt-in Adoption**: Enable enforcement in willing repositories, gather feedback, refine tooling
+3. **Phase 3 - Mandatory Adoption**: Enable enforcement across all Konflux repositories, provide migration support for teams
+
+Specific timelines for each phase will be determined based on team capacity and feedback during the adoption process.
+
+### Adoption Guide for Teams
+
+Teams adopting conventional commits should:
+
+1. Add the shared commitlint GitHub Actions workflow to their repository
+2. Update PR templates to include conventional commit format guidance
+3. Optionally configure local pre-commit hooks for immediate feedback
+4. Review the [Conventional Commits specification](https://www.conventionalcommits.org/) and team-specific scope conventions
+
+## Consequences
+
+### Benefits
+
+* **Automated changelog generation**: The [Konflux-CI Installer](https://github.com/konflux-ci/konflux-ci) release notes automation will leverage conventional commits to generate meaningful changelogs that aggregate `feat` and `fix` commits from all upstream Konflux repositories
+* **Semantic versioning automation**: Version bumps can be determined automatically based on commit types (feat = minor, fix = patch, breaking = major)
+* **Improved commit hygiene**: The required structure encourages more thoughtful, descriptive commit messages
+* **Better searchability**: Structured commits are easier to filter and search (e.g., find all `fix` commits)
+* **Clearer release communication**: Stakeholders can quickly understand what changed in each release
+
+### Costs
+
+* **Learning curve**: Contributors unfamiliar with conventional commits will need to learn the format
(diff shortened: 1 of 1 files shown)


ADR/NNNN-conventional-commits.md
```diff
@@ -0,0 +1,102 @@
...
+2. Update PR templates to include conventional commit format guidance
+3. Optionally configure local pre-commit hooks for immediate feedback
+4. Review the [Conventional Commits specification](https://www.conventionalcommits.org/) and team-specific scope conventions
+
+## Consequences
+
+### Benefits
+
+* **Automated changelog generation**: The [Konflux-CI Installer](https://github.com/konflux-ci/konflux-ci) release notes automation will leverage conventional commits to generate meaningful changelogs that aggregate `feat` and `fix` commits from all upstream Konflux repositories
+* **Semantic versioning automation**: Version bumps can be determined automatically based on commit types (feat = minor, fix = patch, breaking = major)
```
tnevrlka: What version bumps do you have in mind?

ADR/NNNN-conventional-commits.md
```diff
@@ -0,0 +1,102 @@
...
+1. Add the shared commitlint GitHub Actions workflow to their repository
+2. Update PR templates to include conventional commit format guidance
+3. Optionally configure local pre-commit hooks for immediate feedback
+4. Review the [Conventional Commits specification](https://www.conventionalcommits.org/) and team-specific scope conventions
+
+## Consequences
+
+### Benefits
+
+* **Automated changelog generation**: The [Konflux-CI Installer](https://github.com/konflux-ci/konflux-ci) release notes automation will leverage conventional commits to generate meaningful changelogs that aggregate `feat` and `fix` commits from all upstream Konflux repositories
```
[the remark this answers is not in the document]
gbenhaim: I don't think that we can or should commit to a certain structure with this ADR. Showing the changes by services is an option, but it means that we expect users to know more about Konflux internals. Another approach might be sections for features/bugs/security fixes.

ADR/NNNN-conventional-commits.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Status
+
+Proposed
+
+## Context
+
+Konflux currently lacks a standardized commit message format across its repositories. This inconsistency creates several challenges:
+
+* **Manual changelog generation**: Release notes must be compiled manually by reviewing commit histories, which is time-consuming and error-prone
```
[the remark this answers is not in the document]
chmeliik: :+1: 

This is why I think changelogs generated from commits are not a good idea

ADR/NNNN-conventional-commits.md
```diff
@@ -0,0 +1,102 @@
...
+
+* **Automated changelog generation**: The [Konflux-CI Installer](https://github.com/konflux-ci/konflux-ci) release notes automation will leverage conventional commits to generate meaningful changelogs that aggregate `feat` and `fix` commits from all upstream Konflux repositories
+* **Semantic versioning automation**: Version bumps can be determined automatically based on commit types (feat = minor, fix = patch, breaking = major)
+* **Improved commit hygiene**: The required structure encourages more thoughtful, descriptive commit messages
+* **Better searchability**: Structured commits are easier to filter and search (e.g., find all `fix` commits)
+* **Clearer release communication**: Stakeholders can quickly understand what changed in each release
+
+### Costs
+
+* **Learning curve**: Contributors unfamiliar with conventional commits will need to learn the format
```
ralphbean: What percentage of repositories currently use and enforce conventional commits?

The answer implies how large of a learning curve we're looking at here. Do 10% of repos use it today, which will mean a change for 90% of repos? If the gap is large, perhaps we should encourage incremental adoption and learn from that as an experiment.

MartinBasti: quite a lot, but they are adding jira ID instead of scope

tnevrlka:
> For that reason, I think Conventional Commits are good "bang for the buck": most teams are using them to some extent already, it can provide reasonable outcome with small effort, both on the individual components level and on the aggregation level (Konflux Operator release). 

> Yeah, that's fair. The cost of trying this out is relatively small, and the generated release notes could turn out "good enough"

I am not trying to bash this, but I'm still not sure what that "bang" for the buck is. Could somebody please provide a simple specific concrete example of a potential situation where it would be beneficial?

ADR/NNNN-conventional-commits.md
```diff
@@ -0,0 +1,102 @@
...
+## Context
+
+Konflux currently lacks a standardized commit message format across its repositories. This inconsistency creates several challenges:
+
+* **Manual changelog generation**: Release notes must be compiled manually by reviewing commit histories, which is time-consuming and error-prone
+* **Inconsistent commit quality**: Without guidelines, commit messages vary widely in format and informativeness
+* **Difficulty tracking changes**: Understanding what changed between releases requires reading through unstructured commit messages
+* **Limited automation opportunities**: Tools that could automate versioning and changelog generation cannot be used effectively
+
+The [Conventional Commits](https://www.conventionalcommits.org/) specification provides a lightweight convention for commit messages that enables automated tooling while remaining human-readable. It structures commits as:
```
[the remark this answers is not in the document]
chmeliik: The "we" in that comment was meant to be "we the maintainers who care about commit messages". This gives stronger justification to pester contributors about commit message quality.

Most contributors don't understand the benefit, they see it as a chore and don't realize that taking the time to provide context will make their PR get processed far quicker.

With this, we have an easily understandable reason: "Fix your commit messages or you'll make Konflux release notes horrible." It's a big hammer to teach commit message standards.

yftacherzog:
> I am not trying to bash this, but I'm still not sure what that "bang" for the buck is. Could somebody please provide a simple specific concrete example of a potential situation where it would be beneficial?

Initially, we configured releases in konflux operator so they include all commits to the konflux-ci/konflux-ci repo in the release notes (see [example](https://github.com/konflux-ci/konflux-ci/releases/tag/v0.0.9)). This is not very helpful, as being an integration repo, in normal times, it will mainly include bot updates from the upstream repos.

What we want to do instead is for the release process to go over all changes in upstream repositories and collect all the commits that were made to each repo since the last operator release and are meaningful for whoever is going to look at the operator release notes.

Suppose that all those repos follow conventional commits, this process can collect every feat and fix which was not added by a bot, and use that in the operator release notes.

With such approach, we can move from release notes which are almost useless today to something which will be far more meaningful with limited additional effort than what most teams are already doing today.

yftacherzog:
> I would assume it should be somewhat straightforward to filter out commits made by a bot even without conventional commits.

Yes, and we would like to filter it further to only include ones which are more likely for users to find meaningful, and this is where conventional commits become relevant.

> That's what I'm interested in, who's the intended audience and for what purpose would they look at those?
> 
> I would assume predominantly users who want to see new features? Wouldn't listing the PRs make more sense then? I'm not sure if they'd realistically get anything from a single commit

The audience will initially be upstream users, but in the long term, it depends on how we're going to use the operator downstream or productize it.

I think that in many cases, a PR is a single commit, but in any case, I'm not saying that this approach is perfect, but I do think it can get something meaningful with relatively limited extra effort for the average developer.

yftacherzog:
> Yes! And I don't think it's just for "our particular workflow", I can't imagine a situation where you would actually benefit from having the specific commits rather than the PR which usually encapsulates the bigger idea behind the changes which is also what fits the idea for the changelog more imo. Correct me if I'm wrong please.

it's a matter of perspective and preference I suppose. I prefer having a PR including a single commit, as it keeps it limited in scope -- and I can talk about it a lot :) -- but this is the sort of discussion we should not aim to settle here, as I don't think we should try to lead teams one way or the other in that level.

> I'm not sure about that, this seems like mishmash to me. I would rather consider whether using the PR title would suffice.

I don't think we have an easy way to go about it. It will require more changes from teams, which will make it harder to agree upon. I imagine it will also require more API calls eventually when creating the operator release notes, which can lead to a more fragile process.

All in all, I don't want to try the conventional commits approach because it's perfect, but because it will be easy to try on and proceed from there.

eskultety:
Apart from all the comments that have been raised already:

> it's a matter of perspective and preference I suppose. I prefer having a PR including a single commit, as it keeps it limited in scope

What if a PR is still limited in scope but **requires ** more than a single commit?

[Disclaimer: I don't advocate for squash merging]
What if a project uses squash merging? Let's say a PR adds a new feature, but the first commit is some kind of fix rather than the feature itself and hence the commit message is marked as a fix accordingly? What happens when GH squashes everything at the end? The first commit message is what will define the whole PR in terms of change type detection, right? IOW it's going to be a `fix` and then the commit message body will contain all the other messages, but that won't matter for your intended automation, or will it?

> All in all, I don't want to try the conventional commits approach because it's perfect, but because it will be easy to try on and proceed from there.

IIUC we're talking GH native automation here, IOW not something that is multi-platform, correct? In that case, why not building something around labels? That way, the control fully remains in hands of the maintainers rather than developers/contributors who may have not used the conventional commit structure properly.

ADR/NNNN-conventional-commits.md
eskultety: NB: Jira is a resource which more often than not is private --> context is 0 and so if JIRA IDs are being used their use of should be solely some additional automation for a particular entity's use case rather than context replacement.

gbenhaim:
> Personally, I think we should reject this one.
> 
> Adopting standards across the project is a reasonable thing to propose as an ADR, but I don't see the value introduced by conventional commits in particular.
> 
> The use case is producing release notes. I'm open to seeing another ADR proposed in the future for how we should produce release notes for the operator, but massive behavior changes across teams to adopt conventional commits feels like too big of a lift for the value it would provide. What other options do we have?

Let's see how good are the release notes for the operator and what are the gaps and proceed from there. If there is no convention, the release notes of some of the components would be better from the others.

PR #311, opened 2026-03-04
Add Hermeto registry proxy configuration ADR

taylormadore:
- `allow-package-registry-proxy` in the `cluster-config` ConfigMap lets administrators enable or disable package registry proxy usage cluster-wide (default: disabled)
- `enable-package-registry-proxy` as a pipeline parameter lets users opt out of the package registry proxy for individual pipelines (default: enabled)
- The prefetch-dependencies task reads these two flags from the ConfigMap into a single effective setting
- Per-package-manager package registry proxy URL keys in the ConfigMap specify the endpoints for each package ecosystem. These are read directly by the prefetch-dependencies task, which avoids pipeline wiring that scales with package manager count

### DIFF
--- a/ADR/00NN-hermeto-package-registry-proxy.md
+++ b/ADR/00NN-hermeto-package-registry-proxy.md
@@ -1,0 +1,114 @@
+# NN. Package Registry Proxy Configuration for Hermeto
+
+Date: YYYY-MM-DD
+
+## Status
+
+Proposed
+
+## Context
+
+Hermeto is the content prefetching tool used in Konflux build pipelines to fetch application dependencies before the container image build step. It supports multiple package managers including npm, yarn, gomod, pip, cargo, and bundler. Currently, Hermeto fetches dependencies directly from upstream package registries such as registry.npmjs.org, proxy.golang.org, PyPI, crates.io, and rubygems.org.
+
+This direct fetching from upstream sources exposes the build process to supply chain attacks. Malicious packages published to public registries can be consumed by builds without any policy evaluation or control. Recent high-profile incidents have demonstrated that attackers actively target package registries to distribute malware.
+
+To address this security gap, organizations can deploy a package registry proxy with a policy enforcement layer that:
+
+1. **Evaluates all consumed artifacts** against organizational governance policies during ingestion, automatically quarantining components that violate policy before they are available for use in builds.
+2. **Stores all consumed artifacts** in a controlled repository, providing an audit trail and enabling reproducible builds.
+
+By routing Hermeto's dependency fetching through such a package registry proxy, malicious or non-compliant components can be blocked before they ever reach the build environment.
+
+### Architecture Overview
+
+The deployment architecture involves two components between Hermeto and upstream package registries:
+
+1. **Package registry proxy** (e.g., Sonatype Nexus Repository Server): An external service that proxies upstream package registries, enforces governance policies on ingested artifacts, and stores them in controlled repositories.
+2. **On-cluster reverse proxy**: An nginx reverse proxy deployed on the Konflux cluster that transparently injects package registry proxy credentials into outbound requests. Hermeto fetches packages through this proxy, which forwards the authenticated requests to the package registry proxy.
+
+The design and deployment of the on-cluster reverse proxy is outside the scope of this ADR.
+
+**Users** need the ability to:
+
+* Use the package registry proxy automatically when available, without manual configuration
+* Opt out of package registry proxy use for individual pipelines when needed
+
+**Platform administrators** need the ability to:
+
+* Enable/disable package registry proxy usage at the cluster level for all pipelines
+* Configure package registry proxy URLs for different package managers (npm, pip, cargo, etc.) since they may be served from different repositories
+* Apply GitOps processes for this configuration, similar to other Konflux configuration elements
+
+## Decision
+
+We will extend the `cluster-config` ConfigMap described in [ADR 0057](0057-pipeline-caching-feature-flag.md) with an `allow-package-registry-proxy` flag and per-package-manager package registry proxy URL keys. We will also introduce an `enable-package-registry-proxy` pipeline parameter that gives users control over whether their pipeline uses the package registry proxy.
+
+The package registry proxy is used only when both the administrator and user permit it: `allow-package-registry-proxy` must be `true` in the ConfigMap and `enable-package-registry-proxy` must be `true` in the pipeline. The administrator flag defaults to `false`, so the package registry proxy is disabled until explicitly enabled at the cluster level. The pipeline parameter defaults to `true`, so once an administrator enables it, all pipelines use the package registry proxy automatically. Users who need to bypass it for debugging or compatibility reasons can set their pipeline parameter to `false`.
+
+This is the opposite of the OCI cache proxy model in [ADR 0057](0057-pipeline-caching-feature-flag.md), where each user must enable the OCI cache proxy for their own pipeline. The difference reflects the nature of the feature: OCI caching is a performance optimization individual users choose to adopt, while package registry proxying is a security control that should apply globally unless a user explicitly opts out.
+
+The following diagram shows how the package registry proxy fits into the build pipeline alongside the OCI cache proxy described in [ADR 0057](0057-pipeline-caching-feature-flag.md):
+
+```mermaid
+flowchart LR
+    subgraph "Konflux Cluster"
+        subgraph "User Namespace"
+            subgraph "Build PipelineRun"
+                Init["init Task"]
+                Prefetch["prefetch-dependencies Task"]
+                Buildah["buildah Task"]
+                ContainerImage["Completed Container Image"]
+            end
+        end
+        subgraph "konflux-info Namespace"
+            ClusterConfig["ConfigMap/cluster-config"]
+        end
+        subgraph "caching Namespace"
+            OCICacheProxy["OCI Cache Proxy<br/>(squid)"]
+            ReverseProxy["Reverse Proxy for<br/>Package Registry Proxy<br/>(nginx)"]
+        end
+    end
+    subgraph "External Services"
+        PackageRegistryProxy["Package Registry Proxy<br/>(e.g. Sonatype Nexus)"]
+        PackageRegistry["Upstream Package Registries<br/>(npm, PyPI, crates.io, etc.)"]
+        OCIRegistry["OCI Registries"]
+    end
+
+    Init -->|reads| ClusterConfig
+    Init --> |provides<br/>OCI Cache Proxy Settings| Buildah
+    Prefetch -->|reads| ClusterConfig
+    Prefetch -->|provides<br/>package content| Buildah
+    Prefetch -->|fetches packages via| ReverseProxy
+    ReverseProxy -->|forwards authenticated<br/>requests to| PackageRegistryProxy
+    PackageRegistryProxy -->|reads from| PackageRegistry
+    Buildah -->|pulls from| OCICacheProxy
+    Buildah --> |produces| ContainerImage
+    OCICacheProxy -->|pulls from| OCIRegistry
+```
+
+### Implementation Details
+
+Each supported package manager gets its own package registry proxy URL key in the ConfigMap, named `package-registry-proxy-<pkg>-url`. The initially supported package managers are npm, yarn, gomod (with a separate `package-registry-proxy-gomod-sum-url` for the checksum database), pip, cargo, and bundler. Multiple package managers may share the same URL if they use the same upstream registry. The configured URLs point to the on-cluster reverse proxy, which handles authentication to the package registry proxy transparently.
+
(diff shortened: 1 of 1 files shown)


ADR/00NN-hermeto-registry-proxy-feature-flag.md
```diff
@@ -0,0 +1,74 @@
...
+
+**Single base URL**: One URL (e.g., `https://registry-proxy.example.com`) would be configured and per-package-manager paths constructed at runtime. While simpler to configure, this would hardcode path assumptions in the prefetch-dependencies task and prevent administrators from selectively enabling the proxy for some package managers while leaving others disabled, or from routing different package managers to different proxy servers.
+
+**Per-namespace ConfigMap**: Proxy URLs would be placed in a namespace-scoped ConfigMap directly available to tasks. However, [ADR 0057](0057-pipeline-caching-feature-flag.md) already rejected per-namespace configuration for the cache proxy.
+
+## Consequences
+
+* **Attestation**: The `use-registry-proxy` setting flows through Tekton task parameters, making it visible in Chains attestation data. Conforma policies can use this to quickly determine whether the registry proxy was enabled for a build before doing package-level analysis.
```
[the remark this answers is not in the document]
taylormadore: Added

ADR/00NN-hermeto-registry-proxy-feature-flag.md
```diff
@@ -0,0 +1,74 @@
...
+
+The registry proxy is used only when both the administrator and user permit it: `allow-registry-proxy` must be `true` in the ConfigMap and `enable-registry-proxy` must be `true` in the pipeline. The administrator flag defaults to `false`, so the proxy is disabled until explicitly enabled at the cluster level. The pipeline parameter defaults to `true`, so once an administrator enables the proxy, all pipelines use it automatically. Users who need to bypass the proxy for debugging or compatibility reasons can set their pipeline parameter to `false`.
+
+This is the opposite of the cache proxy model in [ADR 0057](0057-pipeline-caching-feature-flag.md), where each user must enable the proxy for their own pipeline. The difference reflects the nature of the feature: caching is a performance optimization individual users choose to adopt, while registry proxying is a security control that should apply globally unless a user explicitly opts out.
```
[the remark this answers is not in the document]
taylormadore: Updated the references. Let me know if that helped make it clearer. Also added a diagram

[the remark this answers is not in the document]
chmeliik: Much better :+1:

ADR/00NN-hermeto-package-registry-proxy.md
```diff
@@ -0,0 +1,114 @@
...
+2. **Stores all consumed artifacts** in a controlled repository, providing an audit trail and enabling reproducible builds.
+
+By routing Hermeto's dependency fetching through such a package registry proxy, malicious or non-compliant components can be blocked before they ever reach the build environment.
+
+### Architecture Overview
+
+The deployment architecture involves two components between Hermeto and upstream package registries:
+
+1. **Package registry proxy** (e.g., Sonatype Nexus Repository Server): An external service that proxies upstream package registries, enforces governance policies on ingested artifacts, and stores them in controlled repositories.
+2. **On-cluster reverse proxy**: An nginx reverse proxy deployed on the Konflux cluster that transparently injects credentials into outbound requests. Hermeto fetches packages through this proxy, which forwards the requests to the package registry proxy with the appropriate authentication.
```
adambkaplan: How does the right credential get associated with the correct workload? I assume (perhaps naively) that these package registry proxies are capable of logging which user requested a vulnerable or compromised package. If all workloads share the same credential, it will be much harder for incident responders to identify which components are affected.

As an industry, we are moving away from static credentials to ephemeral tokens tied to workload identities. I'd rather Hermeto accept a `Bearer: Authorization` token for its requests and not care where it comes from. We can mount this as a Kubernetes Secret for now. In the future this can be replaced with a Kubernetes service account token, or another token generated by a workload identity provider.
I don't see the reverse proxy as being necessary (and it adds yet another man in the middle in our network chain).

Most package managers have standard ways to declare proxy/mirrors - can't Hermeto be configured to understand these configuration options?

taylormadore: > How does the right credential get associated with the correct workload? I assume (perhaps naively) that these package registry proxies are capable of logging which user requested a vulnerable or compromised package. If all workloads share the same credential, it will be much harder for incident responders to identify which components are affected.

The package registry proxy itself can be the first line of defense in terms of evaluating packages against governance policies and blocking vulnerable or compromised packages before they are consumed by a build. For anything that gets through, I believe the SBOM should be the source of truth for incident response rather than an audit trail from the package registry proxy.

> I don't see the reverse proxy as being necessary (and it adds yet another man in the middle in our network chain).

From Hermeto's perspective, there is no authentication in this design. Hermeto simply sends requests to the package registry proxy URLs configured in the ConfigMap. The reverse proxy is really just an implementation detail of the package registry proxy deployment. It was only mentioned in this ADR to explain why there is no authentication configuration for Hermeto in the pipeline at this time. Maybe that is a good case for removing any discussion of it here for now :thinking:.

As to that implementation detail: We tried to avoid per-tenant credentials because workloads run in user-owned tenant namespaces, so users would need to be issued and have access to the credentials. Given that the proxy repositories are read-only, the overhead didn't seem to justify the benefit. Nothing in this design prevents us from changing this in the future if needed.

> Most package managers have standard ways to declare proxy/mirrors - can't Hermeto be configured to understand these configuration options?

That is exactly what we're doing on the Hermeto side. When Hermeto calls a package manager that supports native proxy/mirror configuration, we'll have Hermeto make use of it. When Hermeto handles the fetching itself, we'll add that support directly.

ralphbean: If I can try to paraphrase Adam's idea:

> What if hermeto in the prefetch-deps task didn't talk to the proxy, but it instead talked directly to the package registry proxy, using OIDC federation credentials? We could get rid of the proxy and long-lived credentials.

I think that's a valid aspiration, but:

* The package registry proxy (e.g. Nexus) would have to support it and support authorizing requests from pods in certain clusters in certain AWS accounts.
* The individual package ecosystem toolchains would need to support it.

I bet (but haven't verified) that at least one link in that chain isn't ready or available.

cit1zen: Are the credentials that are being injected Nexus credentials? If yes, can it be explicitly mentioned?

ADR/00NN-hermeto-package-registry-proxy.md
```diff
@@ -0,0 +1,114 @@
...
+This direct fetching from upstream sources exposes the build process to supply chain attacks. Malicious packages published to public registries can be consumed by builds without any policy evaluation or control. Recent high-profile incidents have demonstrated that attackers actively target package registries to distribute malware.
+
+To address this security gap, organizations can deploy a package registry proxy with a policy enforcement layer that:
+
+1. **Evaluates all consumed artifacts** against organizational governance policies during ingestion, automatically quarantining components that violate policy before they are available for use in builds.
+2. **Stores all consumed artifacts** in a controlled repository, providing an audit trail and enabling reproducible builds.
+
+By routing Hermeto's dependency fetching through such a package registry proxy, malicious or non-compliant components can be blocked before they ever reach the build environment.
+
+### Architecture Overview
```
cit1zen, line 22: How are we going to handle situations where we are using authentication when accessing packages?

More specifically:
1. How are we going to handle packages that are only accessible with authorization?
2. How are we going to handle situations like Docker, where we use authentication to access publicly available packages for one reason or the other (in case of Docker, it is higher rate-limits)?

ADR/00NN-hermeto-package-registry-proxy.md
```diff
@@ -0,0 +1,114 @@
...
+
+**Single base URL**: One package registry proxy URL (e.g., `https://registry-proxy.example.com`) would be configured and per-package-manager paths constructed at runtime. While simpler to configure, this would hardcode path assumptions in the prefetch-dependencies task and prevent administrators from selectively enabling the package registry proxy for some package managers while leaving others disabled, or from routing different package managers to different proxy servers.
+
+**Per-namespace ConfigMap**: Package registry proxy URLs would be placed in a namespace-scoped ConfigMap directly available to tasks. However, [ADR 0057](0057-pipeline-caching-feature-flag.md) already rejected per-namespace configuration for the OCI cache proxy.
+
+## Consequences
```
cit1zen, line 105: Can you add a point saying that Hermeto has to be able to recognize if the package pull failed because the Nexus things the package is problematic and show this information to the user?

If this information is not easily visible to the users, they could think that there is some networking or other issue, which can lead to confusion and unnecessary support requests.

cit1zen (commented):
The KGC is inclined to approve this ADR after the questions and comments by us are answered or resolved.

ADR/00NN-hermeto-package-registry-proxy.md
taylormadore: Yes, they are for the package registry proxy :+1: . I will add a statement about that, although I'd like to keep it implementation agnostic if possible and not mention Nexus specifically if that's ok

ADR/00NN-hermeto-package-registry-proxy.md
taylormadore: > How are we going to handle packages that are only accessible with authorization?

Are we talking about auth to upstream package registries or auth to the package registry proxy?

> How are we going to handle situations like Docker, where we use authentication to access publicly available packages for one reason or the other (in case of Docker, it is higher rate-limits)?

Authentication to to upstream registries (if any) would be the package registry proxy's responsibility. It would depend on the implementation details of that service and I believe it is outside the scope of this ADR. For the docker case specifically, I believe that would need to be handled by the cache-proxy in ADR57.

chmeliik: I recall now that Hermeto supports ~/.netrc for accessing private upstream packages (via the `netrc` workspace on the prefetch Task). Will that be compatible with the package registry proxy?

ADR/00NN-hermeto-package-registry-proxy.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+### Implementation Details
+
+Each supported package manager gets its own package registry proxy URL key in the ConfigMap, named `package-registry-proxy-<pkg>-url`. The initially supported package managers are npm, yarn, gomod (with a separate `package-registry-proxy-gomod-sum-url` for the checksum database), pip, cargo, and bundler. Multiple package managers may share the same URL if they use the same upstream registry. The configured URLs point to the on-cluster reverse proxy, which handles authentication to the package registry proxy transparently.
+
+The init task resolves `allow-package-registry-proxy` and `enable-package-registry-proxy` into a single `use-package-registry-proxy` result that is passed to the prefetch-dependencies task, so that the effective setting is visible in Chains attestation data. When `use-package-registry-proxy` is true, the prefetch-dependencies task reads the per-package-manager package registry proxy URLs directly from the ConfigMap and sets the corresponding Hermeto environment variables. This avoids threading a growing number of URL parameters through the pipeline definition as Hermeto adds support for more package ecosystems.
```
chmeliik: This is currently causing confusion in https://github.com/konflux-ci/konflux-build-cli/pull/90, where the obvious approach would be to resolve these on the prefetch side, not the init side.

How critical is it for the resolved `use-package-registry-proxy` setting to show up in the attestation? Would the presence/absence of proxy URLs in the SBOM be sufficient?

taylormadore: This is a really good point. It may be a vestige of how this ADR was originally conceived following the pattern from ADR 57. Once we broke the precedent of the init task being the sole reader of the cluster-config ConfigMap, and with the introduction of the konflux-build-cli, this may no longer be necessary.

There was one concern raised during the Conforma policy review about having a simple boolean to skip per-package proxy URL processing in the SBOM when the package registry proxy is disabled, but that's a secondary concern we can resolve separately.

I'm certainly supportive of keeping things simple :+1:

ADR/00NN-hermeto-package-registry-proxy.md
taylormadore: Yes, I can add a point about this

ADR/00NN-hermeto-package-registry-proxy.md
taylormadore: Private URL and other non-registry dependencies bypass the package registry proxy and are fetched directly from the original source, so .netrc credentials (for Hermeto backends that support them) authenticate to the upstream host as usual.

Registry dependencies, however, are routed through the package registry proxy when enabled. Any .netrc entries for a private upstream registry would be ignored since Hermeto fetches from the package registry proxy instead. If the upstream registry requires authentication, it would need to be configured in the package registry proxy.

ADR/00NN-hermeto-package-registry-proxy.md
taylormadore: Adjusted the text to reflect this change

ADR/00NN-hermeto-package-registry-proxy.md
chmeliik: Ack. If we're not aware of any users relying on .netrc to authenticate with upstream registries, I think this is fine

ralphbean: Yeah, I'm fine with that.

ralphbean (approved):
LGTM

chmeliik (approved):

cit1zen (approved):

lkolacek (approved):

Issue #307, opened 2026-02-11
KubeArchive Missing From Architecture Records


adambkaplan:
KubeArchive is best categorized as an "addon" that is essential for operating Konflux at scale. It plays the vial role of removing Kubernetes data with finite lifecycles, such as `Snapshots`, `Releases`, and most importantly Tekton `PipelineRuns`. There are multiple references to KubeArchive in konflux-ci documentation, however there is no mention of KubeArchive in the architecture docs or decision records.

This should be addressed both through updated architecture documentation as well as a "retroactive" ADR documenting reasons for adopting KubeArchive as a replacement for Tekton Results.

adambkaplan:
cc @ggallen @skoved @konflux-ci/kgc

rh-hemartin:
Hello! I can take care of it.

rh-hemartin:
Hello! I opened something, please take a look

arewm:
#320 was merged. Closing. Thanks!

PR #320, opened 2026-03-16
Add KubeArchive to the architecture

rh-hemartin:
Assisted-By: Claude Code

Solves https://github.com/konflux-ci/architecture/issues/307

### DIFF
--- a/ADR/0064-kubearchive.md
+++ b/ADR/0064-kubearchive.md
@@ -1,0 +1,39 @@
+---
+title: "64. Adopt KubeArchive for resource lifecycle management"
+status: Implemented
+applies_to:
+  - pipeline-service
+  - kubearchive
+topics:
+  - archival
+  - scalability
+  - resource-lifecycle
+---
+
+# 64. Adopt KubeArchive for resource lifecycle management
+
+Date: 2026-03-16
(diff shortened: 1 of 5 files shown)


ADR/0064-kubearchive.md
```diff
@@ -0,0 +1,39 @@
...
+  - resource-lifecycle
+---
+
+# 64. Adopt KubeArchive for resource lifecycle management
+
+Date: 2026-03-16
+
+## Status
+
+Proposed
```
adambkaplan: Nit - change this to `Implemented` to reflect that many Konflux instances deploy Kubearchive as an add-on. I don't think we should block merge on this.

```suggestion
Implemented
```

adambkaplan (approved):
LGTM - only one nit on the ADR's status.

ADR/0064-kubearchive.md
rh-hemartin: That is true, I'll change it.

ralphbean (approved):

arewm (approved):

cit1zen (approved):

ralphbean (approved):

PR #323, opened 2026-03-18
Add ADR XXXX: Ecosystem-Native Distribution for Non-OCI Artifacts

bmbouter:
## Summary

This ADR builds on [ADR-0063](https://github.com/konflux-ci/architecture/blob/main/ADR/0063-generalized-build-release-pipelines.md) by addressing what must happen at the boundary where Konflux-internal OCI artifacts are distributed to language-native ecosystems.

The core requirement: release pipelines for non-OCI artifacts must publish both the artifacts themselves and their associated supply chain metadata (SBOMs, attestations, signatures) to ecosystem-native locations in ecosystem-native formats — not exclusively as OCI referrers. Consumers using standard tooling (`pip`, `dnf`, `mvn`, etc.) must be able to access both the software and its supply chain metadata without interacting directly with an OCI registry.

## Open questions for reviewers

- **`applies_to`**: This ADR currently lists `release-service` and `build-service`. The pre-release/integration testing requirement (the SHOULD bullet in the Decision section) could reasonably also apply to `integration-service`. Feedback welcome on whether that service should be added.

- **ADR number**: The number is currently `XXXX` pending assignment. Please assign the next available number when merging.

## Relationship to ADR-0063

ADR-0063 establishes OCI as Konflux's internal representation for all artifacts and states that release pipelines must publish to native package registries. This ADR makes that requirement explicit for supply chain metadata and pre-release testing, which ADR-0063 does not address.


ADR/XXXX-ecosystem-native-distribution.md
```diff
@@ -0,0 +1 @@
...bmF0aXZlIHdvdWxkIGhhdmUgcmVsZWFzZSBwaXBlbGluZXMgcHVibGlzaCBTQk9NcywgYXR0ZXN0YXRpb25zLCBhbmQgc2lnbmF0dXJlcwpleGNsdXNpdmVseSBhcyBPQ0kgcmVmZXJyZXJzIGF0dGFjaGVkIHRvIHRoZSBfQ29tcG9uZW50IEJ1aWxkIEFydGlmYWN0XywgcmVnYXJkbGVzcyBvZiB0aGUgdGFyZ2V0CnBhY2thZ2UgZWNvc3lzdGVtLgoKVGhpcyBhcHByb2FjaCBpcyByZWplY3RlZCBiZWNhdXNlOgoKLSBFY29zeXN0ZW0tbmF0aXZlIHRvb2xpbmcgZG9lcyBub3QgcXVlcnkgT0NJIHJlZ2lzdHJpZXMgZm9yIHN1cHBseSBjaGFpbiBtZXRhZGF0YS4gQ29uc3VtZXJzCiAgdXNpbmcgc3RhbmRhcmQgdG9vbGluZyB3b3VsZCBoYXZlIG5vIHZpc2liaWxpdHkgaW50byBTQk9NcyBvciBhdHRlc3RhdGlvbnMgcHJvZHVjZWQgYnkgS29uZmx1eC4KLSBQYWNrYWdlIGVjb3N5c3RlbXMgaGF2ZSBhZG9wdGVkIG9yIGFyZSBhY3RpdmVseSBhZG9wdGluZyB0aGVpciBvd24gc3VwcGx5IGNoYWluIG1ldGFkYXRhCiAgc3RhbmRhcmRzLiBQdWJsaXNoaW5nIG1ldGFkYXRhIG9ubHkgdG8gT0NJIHJlZ2lzdHJpZXMgbWVhbnMgS29uZmx1eC1wcm9kdWNlZCBtZXRhZGF0YSBpcwogIGVmZmVjdGl2ZWx5IGludmlzaWJsZSB0byB0aGUgdXNlcnMgaXQgaXMgbWVhbnQgdG8gcHJvdGVjdC4KLSBSZXF1aXJpbmcgZWNvc3lzdGVtIHN0YWtlaG9sZGVycyB0byBhZG9wdCBPQ0kgY2xpZW50cyB0byBhY2Nlc3Mgc3VwcGx5IGNoYWluIGRhdGEgY3JlYXRlcwogIGZyaWN0aW9uIGFuZCBpcyB1bmxpa2VseSB0byBiZSBhZG9wdGVkIGluIHByYWN0aWNlLCBtYWtpbmcgdGhlIHNlY3VyaXR5IGd1YXJhbnRlZXMgS29uZmx1eAogIHByb3ZpZGVzIG1lYW5pbmdsZXNzIGluIHRob3NlIGVjb3N5c3RlbXMuCg==
```
ralphbean: @bmbouter this is base64 encoded! :D  

I decoded it locally for early review, but - needs to be decoded here.

bmbouter: Fixed :)

ADR/XXXX-ecosystem-native-distribution.md
```diff
@@ -0,0 +1,113 @@
...
+  - **RPM**: GPG-signed packages in a Yum/DNF repository with a signed `repomd.xml`.
+  - **Python**: Attestations published to PyPI per [PEP 740](https://peps.python.org/pep-0740/).
+  - **Maven**: Signatures published to Maven Central per its
+    [publication requirements](https://central.sonatype.org/publish/requirements/).
+
+  OCI referrers MAY be used to store supply chain metadata internally within Konflux, but this
+  does not satisfy the requirement for ecosystem-native distribution.
+
+- Build and integration pipelines that make non-OCI artifacts available for pre-release or
+  integration testing SHOULD expose those artifacts via native package tooling. Consumers of
```
mmorhun: I'd replace `SHOULD` with `should be able to`, because it's not always needed to deploy the built non OCI artifacts to the target ecosystem package tooling.

ADR/XXXX-ecosystem-native-distribution.md
```diff
@@ -0,0 +1,113 @@
...
+  attestations, and signatures -- to the ecosystem-native location and in the ecosystem-native
+  format, to the extent possible as allowed by the standards of that ecosystem, which continue to
+  evolve. Examples include:
+  - **RPM**: GPG-signed packages in a Yum/DNF repository with a signed `repomd.xml`.
+  - **Python**: Attestations published to PyPI per [PEP 740](https://peps.python.org/pep-0740/).
+  - **Maven**: Signatures published to Maven Central per its
+    [publication requirements](https://central.sonatype.org/publish/requirements/).
+
+  OCI referrers MAY be used to store supply chain metadata internally within Konflux, but this
+  does not satisfy the requirement for ecosystem-native distribution.
```
mmorhun: What if ecosystem-native distribution doesn't support certain types of the metadata Konflux provides?

ADR/XXXX-ecosystem-native-distribution.md
```diff
@@ -0,0 +1,113 @@
...
+
+- Release pipeline authors for non-OCI ecosystems must have ecosystem expertise. A pipeline that
+  performs only OCI operations does not satisfy this ADR.
+
+- Supply chain metadata for non-OCI artifacts will exist in two locations: as OCI referrers
+  (Konflux-internal) and in the ecosystem-native location (external). Release pipelines are
+  responsible for this translation.
+
+- Pre-release and integration test infrastructure for non-OCI artifacts SHOULD include
+  ecosystem-native package repositories or indices (e.g., a local Yum repository, a local PyPI
```
mmorhun: I don't think that is should be a strict requirement, otherwise it's always required to deploy / configure target ecosystem distribution services. My suggestion is to `SHOULD` -> `should be able to`.

ADR/XXXX-ecosystem-native-distribution.md
```diff
@@ -0,0 +1,113 @@
...
+  index) so that tests can consume artifacts using standard tooling.
```
mmorhun, line 92: With tests it's more tricky than it seems at first.
Imagine a few parallel PRs are created, so we need to make sure that _in all cases_ built artifacts from one PR are _never_ accessed by a test from another PR. In the systems that doesn't bound artifact version to git commit sha it could be quite hard to do.
On contrary, when using an OCI artifacts internally, even for testing, we never face the above problem. The tests can easily access other native artifacts from already released location(s), and I think we can manage to read the artifact to be tested from internal OCI in our own tests.
Anyway, it's a topic to discuss more. I'd like to hear others' opinion on it.

ADR/XXXX-ecosystem-native-distribution.md
bmbouter: Agreed, updated to "should be able to". This keeps the aspiration without mandating infrastructure for every pipeline.

ADR/XXXX-ecosystem-native-distribution.md
bmbouter: Good question. The "to the extent possible as allowed by the standards of that ecosystem" language in the MUST bullets covers this — if an ecosystem does not support a particular metadata type natively, Konflux cannot publish it there. I have added a clarifying sentence: when an ecosystem does not yet support a particular metadata type natively, OCI referrers remain the sole store for that metadata within Konflux.

ADR/XXXX-ecosystem-native-distribution.md
bmbouter: Agreed, updated to match — "should be able to" in both the Decision and Consequences sections.

ADR/XXXX-ecosystem-native-distribution.md
bmbouter: Valid concern — artifact isolation with native package repos across parallel PRs is genuinely harder than with content-addressable OCI. With OCI you get isolation by default; with native repos you would need to solve version namespacing or per-PR repo instances. This is part of why we have kept this aspirational ("should be able to") rather than mandatory. Worth further discussion as specific ecosystems are onboarded and we learn which isolation patterns work in practice.

ADR/XXXX-ecosystem-native-distribution.md
```diff
@@ -0,0 +1,115 @@
+---
+title: "XXXX. Ecosystem-Native Distribution for Non-OCI Artifacts"
+status: Proposed
+applies_to:
+  - release-service
+  - build-service
```
arewm: I am not sure that this applies to the build-service. This is purely a release-time concern.

ADR/XXXX-ecosystem-native-distribution.md
```diff
@@ -0,0 +1,115 @@
...
+  extent possible as allowed by the standards of that ecosystem, which continue to evolve.
+
+- Release pipelines for non-OCI artifacts MUST publish supply chain metadata -- including SBOMs,
+  attestations, and signatures -- to the ecosystem-native location and in the ecosystem-native
+  format, to the extent possible as allowed by the standards of that ecosystem, which continue to
+  evolve. Examples include:
+  - **RPM**: GPG-signed packages in a Yum/DNF repository with a signed `repomd.xml`.
+  - **Python**: Attestations published to PyPI per [PEP 740](https://peps.python.org/pep-0740/).
+  - **Maven**: Signatures published to Maven Central per its
+    [publication requirements](https://central.sonatype.org/publish/requirements/).
```
arewm, line 64: Is there a reason why we limit to only these? NPM packages also have some current support. We don't have any pipelines to publish those, but it could be included as an example. 

I don't know what the balance is as no list is exhaustive, especially in time.

ADR/XXXX-ecosystem-native-distribution.md
```diff
@@ -0,0 +1,115 @@
...
+
+  OCI referrers MAY be used to store supply chain metadata internally within Konflux, but this
+  does not satisfy the requirement for ecosystem-native distribution. When an ecosystem does not
+  yet support a particular metadata type natively, OCI referrers remain the sole store for that
+  metadata within Konflux.
+
+- Build and integration pipelines that make non-OCI artifacts available for pre-release or
+  integration testing should be able to expose those artifacts via native package tooling.
```
arewm: Is this a build time concern specifically or would it be acceptable to have other post-build pipelines like an Integration Test Scenario or a tenant release pipeline to do this conversion?

ADR/XXXX-ecosystem-native-distribution.md
```diff
@@ -0,0 +1,115 @@
...
+  interaction with an OCI registry.
+
+## Consequences
+
+- Release pipeline authors for non-OCI ecosystems must have ecosystem expertise. A pipeline that
+  performs only OCI operations does not satisfy this ADR.
+
+- Supply chain metadata for non-OCI artifacts will exist in two locations: as OCI referrers
+  (Konflux-internal) and in the ecosystem-native location (external). Release pipelines are
+  responsible for this translation.
```
arewm, line 88: One of the consequences here is that we may have some loss in the metadata. If we produce metadata which an ecosystem doesn't support, we will not push it. If release pipelines don't push the OCI artifacts to some "reserved" location then the only source of this information that we have is in the tenants' OCI locations.

ADR/XXXX-ecosystem-native-distribution.md
```diff
@@ -0,0 +1,115 @@
...
+  - **Python**: Attestations published to PyPI per [PEP 740](https://peps.python.org/pep-0740/).
+  - **Maven**: Signatures published to Maven Central per its
+    [publication requirements](https://central.sonatype.org/publish/requirements/).
+
+  OCI referrers MAY be used to store supply chain metadata internally within Konflux, but this
+  does not satisfy the requirement for ecosystem-native distribution. When an ecosystem does not
+  yet support a particular metadata type natively, OCI referrers remain the sole store for that
+  metadata within Konflux.
+
+- Build and integration pipelines that make non-OCI artifacts available for pre-release or
```
cit1zen: Do we check if the non-OCI artifact that we've built and tested and the OCI artifact that we are trying to release are the same thing? If yes, how? Should we make a requirement for such checks a part of the ADR?

There are serious security and QA implications if misalignment between non-OCI and OCI artifacts can go undetected.

ADR/XXXX-ecosystem-native-distribution.md
mmorhun: In my understanding, build pipeline can reuse corresponding release tasks to push the built artifact into internal native repos, if any. So, involvement is trivial and it's mostly about release.
Is it the intention, @bmbouter ?

arewm:
This ADR has missed its milestone. Let's move its milestone out two weeks from today to give @bmbouter time to respond.

ADR/XXXX-ecosystem-native-distribution.md
bmbouter: Good point. Removed `build-service` from `applies_to`. This ADR is primarily a release-time concern. The pre-release testing bullet is now worded to be pipeline-agnostic — it could be an ITS, tenant release pipeline, or a build task that does the conversion. @mmorhun that addresses your question too — the intent is not to mandate build pipelines do this, but to leave the door open for whichever pipeline is appropriate.

ADR/XXXX-ecosystem-native-distribution.md
bmbouter: No reason to limit — the examples are illustrative. Updated to add NPM provenance as an example and added "Examples include, but are not limited to" plus a sentence clarifying the requirement applies to all non-OCI ecosystems.

ADR/XXXX-ecosystem-native-distribution.md
bmbouter: Good question — it does not need to be build-time specifically. Updated the bullet to be pipeline-agnostic: "whether build pipelines, Integration Test Scenarios, or tenant release pipelines." The point is that consumers should be able to access pre-release artifacts via native tooling, regardless of which pipeline stage does the conversion.

ADR/XXXX-ecosystem-native-distribution.md
bmbouter: Good catch — this is a real consequence worth capturing. Added a new bullet: "Supply chain metadata that Konflux produces but the target ecosystem cannot accept natively will only persist in the tenant's OCI location. Release pipelines do not push OCI artifacts to a separate reserved location, so this metadata is not independently preserved outside the tenant's registry."

ADR/XXXX-ecosystem-native-distribution.md
bmbouter: Important concern. The integrity guarantee here comes from the content-addressable nature of the OCI artifact: the release pipeline extracts content from a specific OCI image referenced by digest in the Snapshot, and Conforma/Enterprise Contract policy verifies that the build pipeline executed correctly to produce that artifact. The release pipeline is extracting from a verified, immutable OCI artifact — not from a separate, potentially divergent source.

That said, adding an explicit verification requirement to this ADR would significantly expand its scope. This ADR is focused on the distribution boundary (OCI-internal to ecosystem-native). Verification of build-to-release integrity is better addressed by Conforma policies associated with each build pipeline, as discussed in ADR-0063. If the community feels a dedicated ADR on this topic is warranted, I would support that as a follow-up.

bmbouter:
Apologies for the delayed responses — I will do a better job of replying to comments promptly. My goal is to not miss another milestone.

ADR/XXXX-ecosystem-native-distribution.md
```diff
@@ -0,0 +1,123 @@
...
+
+Konflux's role is to be a secure, universal software factory whose _internal_ representation is
+OCI, while ensuring that all _external_ distribution conforms to the expectations of the target
+ecosystem.
+
+## Decision
+
+- Release pipelines for non-OCI artifacts MUST publish software artifacts to ecosystem-native
+  package registries, in the format and manner required by the target package ecosystem, to the
+  extent possible as allowed by the standards of that ecosystem, which continue to evolve.
```
adambkaplan, line 55: 👍 - this aligns with the spirit of ADR-0063

ADR/XXXX-ecosystem-native-distribution.md
```diff
@@ -0,0 +1,123 @@
...
+
+- Release pipelines for non-OCI artifacts MUST publish supply chain metadata -- including SBOMs,
+  attestations, and signatures -- to the ecosystem-native location and in the ecosystem-native
```
adambkaplan: I think we should constrain this bullet point to attestations and signatures, for the following reasons:

- The examples below focus on them, and do not mention SBOMs.
- Attestations and signatures are things that are generally produced outside of the build pipeline.

SBOM, on the other hand, is generated from within the build pipeline. I think it makes more sense to amend ADR-0063 (or create a follow-up ADR) on this topic to refine our requirements for native ecosystem SBOM generation. ~~Ex: for Maven, SBOMs need to be generated during the build process and placed within the output Maven repository (per generated JAR artifact - there can be multiple!), which is then packaged as an OCI artifact.~~

EDIT: After further analysis, most package ecosystems do not have standards or even strong conventions for distributing SBOMs. Even the Maven ecosystem does not have strong "agreement" on how to do this, as the CycloneDX and SPDX plugins have different distribution behavior.

ADR/XXXX-ecosystem-native-distribution.md
```diff
@@ -0,0 +1,123 @@
...
+  ensures release pipelines translate that to whatever the target ecosystem requires. Konflux does
+  not require downstream ecosystems to adopt OCI-based distribution.
+
+- Supply chain metadata that Konflux produces but the target ecosystem cannot accept natively
+  will only persist in the tenant's OCI location. Release pipelines do not push OCI artifacts to
+  a separate reserved location, so this metadata is not independently preserved outside the
+  tenant's registry.
+
+- Ecosystem-native supply chain standards will evolve. Release pipelines are expected to track and
+  adopt those standards over time.
```
adambkaplan, line 110: 👍

ADR/XXXX-ecosystem-native-distribution.md
adambkaplan: This is also why I didn't include "release to ecosystem-native tooling" in the predecessor build pipeline ADR. 👍 to keeping this consequence aspirational.

Depending on the ecosystem, you can configure hacky work-arounds that extract content from the OCI artifact and "publish" it to a local filesystem directory.

adambkaplan (approved):
@bmbouter this is a great ADR! I only have one minor point around requirements for SBOM, which I wouldn't call a blocker.

mmorhun (approved):

ADR/XXXX-ecosystem-native-distribution.md
```diff
@@ -0,0 +1,123 @@
+---
+title: "XXXX. Ecosystem-Native Distribution for Non-OCI Artifacts"
+status: Proposed
+applies_to:
+  - release-service
```
adambkaplan, line 5: Is Conforma implied by applying this ADR to the release service? (_I think so_).

Conforma policy act as a bridge between build and release. We configure `IntegrationTestScenario`s by default that run at least one Conforma policy against the contents in the `Snapshot`. Adding `integration-service` wouldn't hurt, but IMO is not a blocker for approving this ADR.

ADR/XXXX-ecosystem-native-distribution.md
adambkaplan: Submitted an update to ADR-0063 on this subject: https://github.com/konflux-ci/architecture/pull/337

ADR/XXXX-ecosystem-native-distribution.md
ralphbean: > That said, adding an explicit verification requirement to this ADR would significantly expand its scope.

Ah, but - maybe we don't need to add an explicit verification requirement, but - the new risk should be called out. It is incumbent upon the tenant who is doing their own testing with the ecosystem-native representation of the artifact to understand that there's a chance that the artifact is different from the OCI-native artifact that they'll be planning to release. We can't require them to verify. But, if it is sufficiently important to them - then they should verify. They need to understand and decide if they'll take extra precautions. That burden is a **consequence** of this ADR. Let's spell it out for them in the consequences section. It doesn't have to update the ADR decision.

ADR/XXXX-ecosystem-native-distribution.md
bmbouter: Good point. Removed SBOMs from this MUST bullet — most ecosystems lack standards or even strong conventions for native SBOM distribution (as you noted, even Maven/CycloneDX/SPDX disagree). Attestations and signatures have clearer ecosystem-native distribution paths. SBOM distribution requirements are better addressed in ADR-0063 amendments or a follow-up ADR. Thanks for the PR at https://github.com/konflux-ci/architecture/pull/337.

ADR/XXXX-ecosystem-native-distribution.md
bmbouter: Yes, Conforma is implied — it acts as the bridge between build and release via IntegrationTestScenarios that run policy against the Snapshot. Agreed that adding integration-service is not a blocker. Keeping applies_to as just release-service for now.

ADR/XXXX-ecosystem-native-distribution.md
bmbouter: Great framing — agreed. Added a consequence bullet calling out this risk: tenants testing with ecosystem-native representations should understand those may differ from the OCI artifact that gets released, and should verify if the risk matters to them. This keeps it as a consequence (awareness + tenant responsibility) rather than a requirement.

ralphbean (approved):

ADR/XXXX-ecosystem-native-distribution.md
```diff
@@ -0,0 +1,129 @@
...
```
ralphbean, line 5:
```suggestion
  - integration-service
  - release-service
```

ralphbean: Nevermind - this doesn't _really_ affect integration-service, only possible integration tests provided by users.

arewm (approved):
Thanks for this ADR!

ADR/XXXX-ecosystem-native-distribution.md
```diff
@@ -0,0 +1,129 @@
...
+topics:
```
arewm, line 6:
```suggestion
applies_to:
topics:
```
This doesn't really apply to the release service either.

cit1zen (approved):

ADR/XXXX-ecosystem-native-distribution.md
arewm: I was hoping to be able to modify this with a suggestion, but I don't seem to be able to. Resolving and we can fix this post-merge.

PR #333, opened 2026-04-14
refactor: Architecture "Goals" -> "Principles"

adambkaplan:
This change rephrases the previous architecture "goals" as "principles" and divides them into six short, high-level phrases:

- Build once, publish many
- Git is the source of truth
- Be opinionated yet flexible
- Enforce compliance at the point of change
- Build for distributed, scalable systems
- Secure the system by default

Each phrase (the "principle") has sub-bullets with specific guidance derived from the existing architecture goals. The term "principle" was inspired by the work of Ray Dalio, as published in "Principles: Life & Work" [1].  The term "compliance at the point of change" is taken directly from Carl Nyard, whose work "Compliance in a DevOps Culture" is a clear inspiration for the Konflux project as a whole [2].

[1] https://web.nypl.org/research/research-catalog/bib/cb13219516
[2] https://martinfowler.com/articles/devops-compliance.html

### DIFF
--- a/architecture/index.md
+++ b/architecture/index.md
@@ -10,33 +10,54 @@ toc: true
 
 ## Overview
 
-Konflux is a platform for building integrated software that streamlines, consolidates, and secures the development lifecycle.
+From ["Why Konflux?"](https://konflux-ci.dev/docs):
 
-### Goals
+> Konflux is an open-source continuous integration and delivery (CI/CD) application that helps you
+> secure and simplify the process of building, testing, and releasing rock-solid software faster.
 
-- Build component artifacts from source.
-- Compose software that consists of multiple components, from multiple repositories.
-- Provide transparency on the software supply chain of artifacts: what makes up the software and how was it built.
-- Provide a way for software teams to release to destinations under the control of their SRE or release engineering team(s).
+Key goals of Konflux:
+
+- Build component artifacts from source
+- Compose software that consists of multiple components, from multiple repositories
+- Provide transparency on the software supply chain of artifacts: what makes up the software and how was it built
+- Provide a way for software teams to publish software to authoritative [distribution platforms](https://slsa.dev/spec/v1.2/terminology#distribution-model)
 - Provide a unified user interface across the entire process
 
-## Architecture Goals
+## Architecture Principles
 
-- Build artifacts once with enough data to determine suitability for releasing.
-- Build artifacts once that can be released to multiple locations, multiple use cases.
-- Specify builds and their dependencies entirely from git and not from transient state of the build system. Employ tools like [renovate](https://docs.renovatebot.com/) to manage dependency updates.
-- Build semantically reproducible artifacts. Any configuration which has the potential to affect the semantic functionality of a build should be recorded in the provenance and source controlled whenever possible.
-- Be extensible. Provide opinionated [build pipelines](https://github.com/redhat-appstudio/build-definitions/) and [release pipelines](https://github.com/redhat-appstudio/release-service-catalog), but let users extend those and create their own.
-- "Shift left" the decisions for releasing into PRs; you should be able to release artifacts from a PR as soon as it is merged.
-- Scale without capacity reserved ahead of time.
-- Static stability: the overall system continues to work when a dependency is impaired.
-- Enhancements to the pipelines (the extensible elements of the system) should be rolled out in such a way that individual users can control **when** they accept the update to their namespaces, their processes. Use policy to drive eventual compliance.
-- Each subservice can fulfill its primary use cases independently, without relying on other systems’ availability. An exception to this is the tekton [pipeline service] which provides foundational APIs on which [build-service], [integration-service], and [release-service] depend.
-- Each sub-service owns its data and logic.
-- Communication among services and participants is always asynchronous.
-- Each sub-service is owned by one team. Ownership does not mean that only one team can change the code, but the owning team has the final decision.
-- Minimize shared resources among sub-services.
-- Security, Privacy, and Governance: Sensitive data is protected by fine-grained access control
+> Principles are fundamental truths that serve as the foundations for behavior that gets you what you want out of life.
+_- Ray Dalio, "Principles: Life & Work"_
+
+- **Build once, publish many:**
+  - Every artifact is a release candidate, with enough data to determine suitability for releasing.
+  - A single artifact can be released to multiple locations which serve multiple use cases.
+  - Publishing software automatically from an approved change in source code should be feasible.
+- **Git is the source of truth:**
+  - Build execution is defined and controlled alongside source code.
+  - Artifacts and their dependency trees should be defined in git to the furthest extent possible.
+  - Control the rate of change through git-based review processes (i.e. pull requests, merge requests, etc.), especially for updates to software dependencies and changes to build execution.
+  - Build semantically reproducible artifacts. Any configuration which has the potential to affect the semantic functionality of a build should be recorded in the provenance and source controlled whenever possible.
+- **Be opinionated yet flexible:**
+  - Provide authoritative build execution pipelines and example release pipelines "out of the box."
+  - Provide example programmatic policies that validate a software artifact is ready for publication "out of the box."
+  - Ensure adopters have sufficient knowledge to extend these tools to suit their own needs.
+- **Enforce compliance at the point of change:**
+  - Use programmatic policies to determine if a software artifact meets publication criteria.
+  - Allow policies to be tailored to specific distribution platform requirements.
+  - Allow policy definition and application to be done separately from the software build process.
+- **Build for distributed, scalable systems:**
+  - Scale without capacity reserved ahead of time.
+  - Communication among services and participants is asynchronous to the furthest extent possible.
+  - Expect failures from downstream/dependent systems, and handle them gracefully.
+  - Each subservice can fulfill its primary use cases independently, with minimal synchronous dependencies.
+  - Each sub-service owns its data and logic.
+  - Minimize shared resources among sub-services.
+  - Each sub-service is owned by one team. Ownership does not mean that only one team can change the code, but the owning team has the final decision.
+- **Secure the system by default:**
+  - Sensitive data is protected by fine-grained access control.
+  - Users should not be able to view, modify, or interfere with builds without explicit authority.
+  - The build system provides cryptographically verifiable evidence of the actions taken that are related to a specific software artifact.
+
 
 ## Architecture Constraints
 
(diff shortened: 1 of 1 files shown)


architecture/index.md
```diff
@@ -10,33 +10,54 @@ toc: true
...
-- Each sub-service is owned by one team. Ownership does not mean that only one team can change the code, but the owning team has the final decision.
-- Minimize shared resources among sub-services.
-- Security, Privacy, and Governance: Sensitive data is protected by fine-grained access control
+## Architecture Principles
+
+> Principles are fundamental truths that serve as the foundations for behavior that gets you what you want out of life.
+_- Ray Dalio, "Principles: Life & Work"_
+
+- **Build once, publish many:**
+  - Every artifact is a release candidate, with enough enough data to determine suitability for releasing.
```
rrosatti: nit: the `enough` word is repeated twice here

architecture/index.md
```diff
@@ -10,33 +10,54 @@ toc: true
...
+  - A single artifact can be released to multiple locations which serve multiple use cases.
+  - Publishing software automatically from an approved change in source code should be feasible.
+- **Git is the source of truth:**
+  - Build execution is defined and controlled alongside source code.
+  - Artifacts and their dependency trees should be defined in git to the furthest extent possible.
+  - Control the rate of change through git-based review processes (i.e. pull requests, merge requests, etc.), especially for updates to software dependencies and changes to build execution.
+  - Build semantically reproducible artifacts. Any configuration which has the potential to affect the semantic functionality of a build should be recorded in the provenance and source controlled whenever possible.
+- **Be opinionated yet flexible:**
+  - Provide authoritative build execution pipelines and example release pipelines "out of the box."
```
rrosatti: nit: in the previous text we had hyperlinks to `build-definitions` and `release-service-catalog` repos, maybe we could do the same here? :)

e.g. - Provide authoritative [build execution pipelines](https://github.com/konflux-ci/build-definitions/) and example [release pipelines](https://github.com/konflux-ci/release-service-catalog) "out of the box."

adambkaplan: cc @eskultety - on the build pipeline side we are decentralizing the catalog. We've already done this for RPMs, and there are folks building Python wheels on Konflux with their pipelines defined completely outside of the konflux-ci org. I am less certain of what we are doing for release pipelines.

architecture/index.md
```diff
@@ -10,33 +10,54 @@ toc: true
 
 ## Overview
 
-Konflux is a platform for building integrated software that streamlines, consolidates, and secures the development lifecycle.
+From ["Why Konflux?"](https://konflux-ci.dev/docs):
```
arewm, line 13: What is the intention here? Are we trying to make the pages consistent or just indicating the relatedness? I would expect that these things are going to drift.

architecture/index.md
```diff
@@ -10,33 +10,54 @@ toc: true
...
-- Minimize shared resources among sub-services.
-- Security, Privacy, and Governance: Sensitive data is protected by fine-grained access control
+## Architecture Principles
+
+> Principles are fundamental truths that serve as the foundations for behavior that gets you what you want out of life.
+_- Ray Dalio, "Principles: Life & Work"_
+
+- **Build once, publish many:**
+  - Every artifact is a release candidate, with enough data to determine suitability for releasing.
+  - A single artifact can be released to multiple locations which serve multiple use cases.
```
arewm: Do we want to say that each of these locations can be driven by their own policy requirements?

architecture/index.md
```diff
@@ -10,33 +10,54 @@ toc: true
...
+  - Publishing software automatically from an approved change in source code should be feasible.
+- **Git is the source of truth:**
+  - Build execution is defined and controlled alongside source code.
+  - Artifacts and their dependency trees should be defined in git to the furthest extent possible.
+  - Control the rate of change through git-based review processes (i.e. pull requests, merge requests, etc.), especially for updates to software dependencies and changes to build execution.
```
arewm: Sources should be stored in git, but it isn't necessarily are requirement for the pipeline definition to be there. Is this too nuanced of a topic to try to address?

architecture/index.md
arewm: Even if the build tasks are decentralized, there should still be one centralized definition of the pipelines that we use for initial configuration.

Specific Konflux instances can create a ConfigMap to provide the default pipelines that are available. This doesn't mean that those are the only ones that _can_ be available. I don't think that @rrosatti is trying to say that users of Konflux _must_ use these pipelines, just that Konflux can provide sensible defaults as an easy onbaording path. As teams onboard more components, it is natural that they will find the desire to customize their pipelines as they are able to as well as [evolve the build pipeline management](https://konflux-ci.dev/docs/patterns/managing-pipeline-changes/).

For release pipelines, we have a repository with many tasks in it. Some of these are general purpose, others are more tailored to specific organizations. Regardless, the release process points to a specific pipeline using a git resolver in the release catalog repo.

architecture/index.md
```diff
@@ -10,33 +10,54 @@ toc: true
...
+  - Build semantically reproducible artifacts. Any configuration which has the potential to affect the semantic functionality of a build should be recorded in the provenance and source controlled whenever possible.
+- **Be opinionated yet flexible:**
+  - Provide authoritative build execution pipelines and example release pipelines "out of the box."
+  - Provide example programmatic policies that validate a software artifact is ready for publication "out of the box."
```
arewm: What does "out of the box" mean here? If you are trying to say that this should be provided to any Konflux instance, then it has to be based on the default pipelines as mentioned in the previous comment.

If you are trying to say something closer to as soon as artifacts are built, I have use a term "policy driven development" in the past.

architecture/index.md
```diff
@@ -10,33 +10,54 @@ toc: true
...
+  - Ensure adopters have sufficient knowledge to extend these tools to suit their own needs.
+- **Enforce compliance at the point of change:**
+  - Use programmatic policies to determine if a software artifact meets publication criteria.
+  - Allow policies to be tailored to specific distribution platform requirements.
+  - Allow policy definition and application to be done separately from the software build process.
```
arewm: Is this an allow or a require?

architecture/index.md
```diff
@@ -10,33 +10,54 @@ toc: true
...
+  - Communication among services and participants is asynchronous to the furthest extent possible.
+  - Expect failures from downstream/dependent systems, and handle them gracefully.
+  - Each subservice can fulfill its primary use cases independently, with minimal synchronous dependencies.
+  - Each sub-service owns its data and logic.
+  - Minimize shared resources among sub-services.
+  - Each sub-service is owned by one team. Ownership does not mean that only one team can change the code, but the owning team has the final decision.
+- **Secure the system by default:**
+  - Sensitive data is protected by fine-grained access control.
+  - Users should not be able to view, modify, or interfere with builds without explicit authority.
+  - The build system provides cryptographically verifiable evidence of the actions taken that are related to a specific software artifact.
```
arewm: This feels incomplete. It is cryptographically verifiable that it was generated by a neutral observer.

architecture/index.md
```diff
@@ -45,7 +66,7 @@ Konflux is a platform for building integrated software that streamlines, consoli
...
 - The cluster is our **unit of sharding**. Each cluster is independent and runs an instance of every subsystem. User namespaces are allocated to one cluster. If we lose a cluster, all namespaces on that cluster are inaccessible, but namespaces on other clusters are not impacted, limiting the blast radius. No subsystem should coordinate across clusters.
 - Artifacts built, tested, and shipped by the system are **OCI artifacts**. SBOMs, attestations, signatures, and other supporting metadata are stored in the registry alongside the artifact, tagged by the `cosign triangulate` convention.
-- While not true today, it should be possible to **install** one subsystem without the others and to replace one subsystem with a new one without affecting the others. See [!148](https://github.com/redhat-appstudio/architecture/pull/148) for an example of attempting to achieve this.
+- While not true today, it should be possible to **install** one subsystem without the others and to replace one subsystem with a new one without affecting the others. See [!148](https://github.com/konflux-ci/architecture/pull/148) for an example of attempting to achieve this.
```
arewm: Is this still relevant? This is a closed and unmerged ADR.

architecture/index.md
adambkaplan: My main goal was consistency. I think it's OK for these to drift, however at this point I think we are at a moment where the two can be aligned.

architecture/index.md
adambkaplan: Yes, we should!

I think this framing also helps with the notion of "managed" release pipelines. Each release destination has an owning team ("release platform") that sets policy and controls the release process.

"Tenant" release pipelines are then a special case where the development team and the "release platform" team are one and the same.

architecture/index.md
adambkaplan: Might be a bit too nuanced.

I'm iterating on an idea right now that would allow the build pipelines to be managed through a separate GitOps repository. This would require a few changes to Tekton that are currently in progress.

architecture/index.md
```diff
@@ -10,33 +10,54 @@ toc: true
...
+  - Provide example programmatic policies that validate a software artifact is ready for publication "out of the box."
+  - Ensure adopters have sufficient knowledge to extend these tools to suit their own needs.
+- **Enforce compliance at the point of change:**
+  - Use programmatic policies to determine if a software artifact meets publication criteria.
+  - Allow policies to be tailored to specific distribution platform requirements.
+  - Allow policy definition and application to be done separately from the software build process.
+- **Build for distributed, scalable systems:**
+  - Scale without capacity reserved ahead of time.
+  - Communication among services and participants is asynchronous to the furthest extent possible.
+  - Expect failures from downstream/dependent systems, and handle them gracefully.
```
ralphbean, line 53: Nit: the sub-bullets under the principles mix imperative ("Scale without capacity…"), declarative ("Each sub-service owns its data…"), and prescriptive ("should be feasible") voices. Not blocking, just noting the inconsistency.

architecture/index.md
ralphbean: Nit: `i.e.` means "that is" (exhaustive), but `etc.` implies the list is non-exhaustive. Should probably be `e.g.` ("for example").

architecture/index.md
ralphbean: Nit: `[!148]` is GitLab merge-request notation. On GitHub, pull requests use `#`, so this would conventionally read `[#148]`. The URL itself is correct.

architecture/index.md
```diff
@@ -10,33 +10,54 @@ toc: true
...
+  - Each subservice can fulfill its primary use cases independently, with minimal synchronous dependencies.
+  - Each sub-service owns its data and logic.
+  - Minimize shared resources among sub-services.
```
ralphbean, line 56: Nit: "subservice" (line 54) and "sub-service" (lines 55, 56, 57) appear in the same section. The Constraints section also uses "subsystem." Might be worth picking one consistent form.

ralphbean (approved):
Thanks so much for this.

arewm:
@adambkaplan , while this is not an ADR, the KGC is treating it as if it is one because of the expanse of the content that it modifies. According to [that process](https://github.com/konflux-ci/community/blob/main/ADRs.md), the author is responsible for responding to all threads.

This is currently scheduled in the KGC review milestone for next week.

architecture/index.md
adambkaplan: I think "default deployment experience" is a better term. An early adopter should be able to use project-maintained build pipelines, release pipelines, and Conforma policies with minimal "extra" configuration. They should then have the ability to inspect, extend, and customize these tools to suit their needs.

architecture/index.md
adambkaplan: Require is probably better.

architecture/index.md
adambkaplan: I can drop the reference.

I actually struggle with this sentence, and perhaps this should be taken up as a separate PR/ADR. If we're thinking about the ideal for a resilient, distributed system:

1. The sub-service should be able to **install** itself with no dependency on other sub-services.
2. The sub-service should be able to **remain operational** when a dependency is not present, is not operational, or is operating in a degraded capacity. Ex: if a dependent CRD is not present, the sub-service Pod should not crashloop on start.
3. The sub-service should be able to **restore full functionality** when a dependent system is present or recovers from an outage. Ex: once the CRD is present, start the controller reconcile loops.

architecture/index.md
adambkaplan: Picking `sub-service`

architecture/index.md
adambkaplan: Fixed in latest push.

architecture/index.md
adambkaplan: Fixed in latest push

architecture/index.md
adambkaplan: Fixed in latest push

architecture/index.md
adambkaplan: I added the links back.

architecture/index.md
adambkaplan: Updated to try and "thread the needle." Build execution and code need to be in git, but they need not be in the same repository/code tree.

architecture/index.md
arewm: Resolving as you removed the ADR reference. I think it is still fine to have the ideal state of independence recorded.

This can certainly be something that can be raised as a separate ADR in the future.

cit1zen (approved):

arewm (approved):

ralphbean (approved):

PR #336, opened 2026-04-22
Use map-based schema for trusted_task_rules

joejstuart:
OPA doesn't merge arrays, so adding a top-level key.

### DIFF
--- a/ADR/0053-trusted-task-model.md
+++ b/ADR/0053-trusted-task-model.md
@@ -115,6 +115,11 @@ Note:
 This ADR doesn't mandate a specific format for `trusted_task_rules`, but does
 come with an example of what it could look like - see the [Appendix](#appendix).
 
+Note: The `trusted_task_rules` structure uses maps (keyed by a descriptive name)
+rather than arrays for the `allow` and `deny` fields. This is because OPA does
+not merge arrays — when multiple data sources contribute rules, array values
+would conflict instead of combining. Map-based keys merge naturally in OPA.
+
 ### Trust upstream Konflux Tasks based on bundle URL
 
 A policy author who wants to trust all the upstream Konflux Tasks should do so by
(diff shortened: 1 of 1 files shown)


ADR/0053-trusted-task-model.md
```diff
@@ -115,6 +115,11 @@ Note:
 This ADR doesn't mandate a specific format for `trusted_task_rules`, but does
 come with an example of what it could look like - see the [Appendix](#appendix).
 
+Note: The `trusted_task_rules` structure uses maps (keyed by a descriptive name)
+rather than arrays for the `allow` and `deny` fields. This is because OPA does
+not merge arrays — when multiple data sources contribute rules, array values
+would conflict instead of combining. Map-based keys merge naturally in OPA.
```
chmeliik: I think this would make more sense in the Appendix, which is the only part of the ADR that mentions the suggested structure

It may also be worth calling out how Conforma is expected to process the map. I assume it should just flatten the values into one big array and process the array

joejstuart: Added to appendix with an explanation - https://github.com/konflux-ci/architecture/pull/336/changes/ed5c74addaabc3f2da8f3590b7b2be8b1eb082cb

ADR/0053-trusted-task-model.md
```diff
@@ -202,6 +202,39 @@ Note that Conforma also supports [setting Task expiry] manually via the
...
+```yaml
+# Bundle A                          # Bundle B
+trusted_task_rules:                  trusted_task_rules:
+  allow:                               allow:
+    tekton-catalog-tasks:                my-org-tasks:
+      - pattern: oci://quay.io/…          - pattern: oci://my.registry/…
+```
+
+Because `allow` is a map, OPA merges the two: the result has both
+`tekton-catalog-tasks` and `my-org-tasks` as keys under `allow`. If `allow`
```
chmeliik: What would happen if the keys were the same? Do the arrays get merged, or does one overwrite the other?

joejstuart: You get a merge error and the policy run fails. It's mentioned just above this line, but let me know if it's not clear enough. 

```
But if two sources define the same key path with a non-map value (an array, a scalar, or null),
OPA treats it as a conflict and **fails the entire data load**.
```

chmeliik: Ah, I see, for some reason I thought that only applied to the top-level keys but not keys nested deeper in the structure. That explains it

ADR/0053-trusted-task-model.md
```diff
@@ -202,6 +202,39 @@ Note that Conforma also supports [setting Task expiry] manually via the
 
 ## Appendix
 
+### Why maps instead of arrays
```
chmeliik: nitpick: maybe move this under the "Example `trusted_task_rules` mechanism" section. Up here, it looks taken out of context.

chmeliik (approved):

ralphbean (approved):

cit1zen (approved):

arewm (approved):
If we are updating ADRs to change the content then we would need to submit a new ADR. Since this is just updating an example, approving.

PR #337, opened 2026-04-22
feat: SBOMs for Package Ecosystems

adambkaplan:
Amend ADR-0063 to require that build pipelines generate SBOMs in a manner that meets the standards of the respective package ecosystem. Python has a standard for distributing SBOMs, whereas other package ecosystems have "patterns" that can be adopted [1].

[1] https://github.com/CycloneDX/cyclonedx-maven-plugin#goals


ralphbean (approved):

arewm (approved):
If we are updating ADRs to change the content then we would need to submit a new ADR. Since this is just offering further clarification for package ecosystems in a non-substantive way, approving.

PR #354, opened 2026-05-11
feat(STONEINTG-1559): add nudging relationship storage via singleton CRD, (ADR-0067)

kasemAlem:
# nudging relationship migration

Date: 2026-05-27

## Status
                 
                 Proposed

## Context

  ### Summary                                                                                                                                                                                                       
                                                                                                                                                                                                                   
  This ADR proposes moving nudging relationship storage from the `build-nudges-ref` field on the Component CR to a new singleton `NudgeConfig` CRD (one per namespace), owned by the integration-service.          
                                                                                                                                                                                                                   
  ### Why                                                                                                                                                                                                           
                                                                                                                                                                                                                   
  - The revised component model (ADR-0056/0060) splits ownership: build-service owns Component, integration-service owns orchestration. Keeping nudge config on Component violates this boundary.                  
  - Current immediate-on-build nudging causes redundant downstream builds when upstream tests fail. The new `validated` mode waits for snapshot tests to pass before nudging.
  - A singleton CRD provides a structured API the UI can consume directly, unlike annotations or ConfigMaps.                                                                                                       
  - etcd impact is minimal: ~48 lightweight CRs across production vs. thousands of PipelineRuns at 165-399 KB each.                                                                                                
                                                                                                                                                                                                                   
  ### Decisions                                                                                                                                                                                                 
                                                                                                                                                                                                                   
  - **Singleton per namespace** — one `NudgeConfig` CR named `nudge-config`, enforced by webhook                                                                                                                   
  - **Two nudge modes** — `immediate` (post-build, current behavior) and `validated` (post-test, reduces redundant builds)
  - **Component-to-component only** (Phase 1) — group relationships deferred to Phase 2                                                                                                                            
  - **Stale reference controller** — watches Component deletions, reports orphaned references via status conditions                                                                                                
  - **Automated migration** from existing `build-nudges-ref` fields                                                                                                                                                
                                                                                                                                                                                                                   
  Addresses [STONEINTG-1559](https://redhat.atlassian.net/browse/STONEINTG-1559)
  
  Related issues: 
  [STONEINTG-1588](https://issues.redhat.com/browse/STONEINTG-1588) / [STONEINTG-1495](https://issues.redhat.com/browse/STONEINTG-1495).   

[STONEINTG-1588]: https://redhat.atlassian.net/browse/STONEINTG-1588?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[STONEINTG-1559]: https://redhat.atlassian.net/browse/STONEINTG-1559?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ADR/0067-nudging-relationship-singleton-crd.md
+++ b/ADR/0067-nudging-relationship-singleton-crd.md
@@ -1,0 +1,421 @@
+---
+title: "67. Nudging Relationship Storage via Singleton CRD"
+status: Proposed
+applies_to:
+  - integration-service
+  - build-service
+topics:
+  - nudging
+  - component-dependencies
+  - orchestrated-builds
+  - singleton-crd
+---
+
+# 67. Nudging Relationship Storage via Singleton CRD
+
+Date: 2026-05-11
+
+## Status
+
+Proposed
+
+Partially supersedes [ADR 29. Component Dependencies](0029-component-dependencies.html)
+with respect to nudging relationship storage and execution. This ADR
+moves both the nudge configuration (from the Component CR's
+`build-nudges-ref` field to a new `NudgeConfig` CRD) and the nudge
+execution (from build-service to integration-service). The nudging
+*concept* — directed component-to-component dependency edges that
+trigger downstream rebuilds via Renovate — remains as defined in ADR 29.
+Build-service retains ownership of the build pipeline and Renovate
+tooling; integration-service becomes the orchestrator that decides
+*when* and *whether* a nudge fires.
+
+Builds upon [ADR 60. ComponentGroups](0060-component-groups.html) and
+[ADR 56. Revised Component Model](0056-revised-component-model.html).
+
+## Context
+
+### Current State
+
+Today, component nudging relationships are defined in the `Component` CR
+via the `build-nudges-ref` field in the spec
+([ADR 29](0029-component-dependencies.html)). When a component build
+succeeds, the build-service triggers Renovate to create pull requests
+that update image references in downstream ("nudged") components.
+
+### Problems with the Current Approach
+
+The [revised component model](0056-revised-component-model.html) and the
+introduction of [ComponentGroups](0060-component-groups.html) create
+several problems for the current nudging storage:
+
+1. **Ownership boundary violation.** Under the revised model, the
+   build-service owns the `Component` CR while the integration-service
+   owns orchestration and testing. Keeping nudging relationships on the
+   Component spec forces the build-service to understand dependency
+   semantics it should not own.
+
+2. **Redundant builds.** The current immediate-nudge-on-build model
+   triggers downstream builds before upstream tests have passed. When
+   upstream tests fail, the downstream builds are wasted. With
+   multi-component groups this problem compounds: an operator bundle
+   component may be nudged N times for N upstream builds that could have
+   been batched into a single nudge after validation.
+
+3. **Component Group ambiguity.** A component can belong to multiple
+   ComponentGroups. Storing nudge relationships on the component makes it
+   impossible to express group-scoped nudging rules without ambiguity.
+
+4. **No structured API for UI consumption.** The current field is a
+   simple list embedded in the Component spec. The UI team cannot easily
+   discover or visualize the full dependency graph across a namespace.
+
+5. **Application-scoped limitations.** Under ADR 29, nudge
+   relationships are Application-scoped — cross-Application nudges are
+   invalid. With the deprecation of the Application CR (ADR 56/60) and
+   the introduction of ComponentGroups where a component can belong to
+   multiple groups, Application-scoping is no longer a meaningful
+   boundary. This ADR intentionally broadens nudge scope to the
+   namespace level, which aligns with Konflux's existing tenant
+   isolation boundary. All components within a namespace may participate
+   in nudge relationships regardless of their ComponentGroup membership.
+
+### Alternatives Considered
+
+| Alternative | Why rejected |
+|---|---|
+| **Annotation on Component CR** | Unstructured data, poor UI discoverability, no schema validation. Breaks backward compatibility when adding future fields (e.g., nudge timing mode). |
+| **Annotation on build PipelineRun** | Requires git-stored configuration, removes the ability to manage dependencies through the UI. |
+| **ConfigMap per namespace** | Same etcd footprint as a CR but without schema validation, RBAC granularity, or structured API. Expensive to watch and cache. |
+| **Dedicated CRD per component pair** | Object explosion risk: namespaces with many components could create thousands of CRs, exacerbating etcd pressure. |
+| **Field on ComponentGroup CR** | A component can belong to multiple ComponentGroups, making it ambiguous where the relationship is defined. Also mixes group topology concerns with dependency concerns. |
+
+### Constraints
+
+- **etcd pressure.** The Konflux infrastructure team has identified etcd
+  storage as the primary scaling bottleneck. Any new CR must minimize
+  object count. The singleton-per-namespace design ensures at most one
+  NudgeConfig CR exists per tenant namespace, and only in namespaces
+  that use nudging. This is negligible compared to the per-build
+  PipelineRun objects that dominate etcd storage.
+
+- **1 MB resource size limit.** Kubernetes enforces a 1 MB limit per
+  resource. The singleton design must account for namespaces with many
+  component relationships.
+
+- **Stale data.** When components are deleted, the nudging configuration
+  must not retain references to non-existent components, otherwise
+  downstream nudge attempts will fail or produce confusing errors.
+
+## Decision
+
+Introduce a new **`NudgeConfig`** Custom Resource Definition, implemented
+as an **optional singleton per namespace**, to store all component nudging
+relationships for that namespace. Namespaces that do not use nudging
+do not need to create a `NudgeConfig`. Ownership of nudging relationship
+storage moves from the build-service (`Component` CR) to the
+integration-service.
+
+### CRD Design
+
+```yaml
+apiVersion: konflux-ci.dev/v1alpha1
+kind: NudgeConfig
+metadata:
+  name: nudge-config          # singleton — one per namespace
+  namespace: my-tenant
+spec:
+  nudges:
+    - from: component-a       # source component (the one being built)
+      to: component-b         # target component (the one to be nudged)
+      mode: validated          # "immediate" | "validated"
+      gatingGroup: frontend-group  # required when mode=validated
+    - from: component-a
+      to: component-c          # mode defaults to "immediate"
+    - from: component-d
+      to: component-e
+      mode: validated
+      gatingGroup: backend-group
+status:
+  conditions:
+    - type: Valid
+      status: "True"
+      reason: AllComponentsExist
+      message: "All referenced components exist in namespace"
+    - type: Valid
+      status: "False"
+      reason: StaleReferences
+      message: "Components [component-x] referenced in nudges no longer exist"
+  lastValidationTime: "2026-05-11T10:00:00Z"
+```
+
+### Fields
+
+| Field | Type | Description |
+|---|---|---|
+| `spec.nudges` | `[]NudgeRelationship` | List of directed component-to-component nudge edges. |
+| `spec.nudges[].from` | `string` | Name of the source component whose build triggers the nudge. |
+| `spec.nudges[].to` | `string` | Name of the target component to be nudged. |
+| `spec.nudges[].mode` | `enum` | Optional. `immediate`: nudge fires after the source build PLR succeeds. `validated`: nudge fires after the Snapshot for the gating ComponentGroup passes integration tests. Defaults to `immediate` if omitted. |
+| `spec.nudges[].gatingGroup` | `string` | Name of the ComponentGroup whose Snapshot must pass integration tests before the nudge fires. Required when `mode: validated`; ignored for `immediate`. Phase 1 reserves this field; implementation deferred to Phase 2. |
+| `status.conditions` | `[]metav1.Condition` | Reports validation state — specifically whether all referenced components still exist. |
+| `status.lastValidationTime` | `metav1.Time` | Timestamp of the last stale-reference validation run. |
+
+### Singleton Enforcement
+
+The singleton constraint is enforced via a CEL validation rule in the
+CRD schema (`x-kubernetes-validations`) that restricts
+`metadata.name` to `nudge-config`. This runs in-process in the API
+server with no webhook overhead. Attempts to create an instance with
+any other name are rejected.
+
+### Validation Rules
+
+The following constraints are enforced on `NudgeConfig`, inheriting
+the DAG enforcement responsibility from Application-service (which is
+being deprecated per ADR 56/60). Stateless rules use CEL validation
+in the CRD schema; stateful rules that require cluster queries use a
+[ValidatingAdmissionPolicy](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/)
+or a lightweight webhook.
+
+**Stateless (CRD-level CEL):**
+
+1. **Self-nudge rejection.** An entry where `from` equals `to` is
+   rejected. A component cannot nudge itself.
+
+2. **Duplicate pair rejection.** Only one entry per `(from, to)` pair is
+   allowed. If a user needs to change the mode for an existing pair,
+   they must update the existing entry rather than add a second one.
+
+3. **Cycle detection.** The set of nudge entries forms a directed graph.
+   Any update that would introduce a cycle is rejected, including
+   transitive cycles (e.g., A→B→C→A). This prevents infinite nudge
+   loops. Validation is performed on the full graph within the
+   `NudgeConfig` on every create/update.
+
+**Stateful (ValidatingAdmissionPolicy or webhook):**
+
+4. **Component existence validation.** Both `from` and `to` must
+   reference components that exist in the namespace at the time of
+   creation. Post-creation staleness (component deleted after the entry
+   was added) is handled by the stale reference controller described
+   below, not by the admission check.
+
+### Nudging Flow
+
+In Phase 1, nudging is triggered by push (post-merge) build
+PipelineRuns only, consistent with the current ADR 29 behavior. The
+design does not preclude extending nudging to pull request builds in
+the future.
+
+#### Immediate Mode (`mode: immediate`)
+
+```
+Build PLR succeeds for component-a
+  → integration-service reads NudgeConfig
+  → finds edge: component-a → component-b (immediate)
+  → triggers Renovate nudge PipelineRun for component-b
+    in the tenant namespace
+  → component-b build starts immediately
+```
+
+#### Validated Mode (`mode: validated`)
+
+```
+Build PLR succeeds for component-a
+  → integration-service updates GCL in the gating ComponentGroup
+  → Snapshot is created with latest GCL images
+  → Integration tests run against the Snapshot
+  → All tests pass (Snapshot is marked as successful)
+  → integration-service reads NudgeConfig
+  → finds edge: component-a → component-b (validated, gatingGroup: frontend-group)
+  → triggers Renovate nudge PipelineRun for component-b
+    in the tenant namespace, using the validated image digest
+    from the passing Snapshot
+```
+
+The validated mode leverages the Global Candidate List (GCL) and
+Snapshot infrastructure to eliminate redundant downstream builds. Instead
+of nudging N times for N upstream builds, the integration-service waits
+for the Snapshot to pass and fires a single nudge with the validated
+image.
+
+Since a component can belong to multiple ComponentGroups
+([ADR 60](0060-component-groups.html)), the `gatingGroup` field
+explicitly resolves which group's Snapshot must pass before the nudge
+fires. This avoids ambiguity when the source component participates in
+multiple groups with different test configurations.
+
+### Migration from `build-nudges-ref`
+
+The integration-service will perform a one-time migration:
+
+1. Read existing `build-nudges-ref` values from all `Component` CRs in
+   each namespace.
+2. Generate the corresponding `NudgeConfig` singleton with `mode:
+   immediate` (preserving current behavior).
+3. After migration is confirmed, the `build-nudges-ref` field is
+   deprecated and eventually removed from the Component spec.
+
+Users who want validated nudging can then update their `NudgeConfig` to
+change specific edges to `mode: validated`.
+
+For users managing resources via GitOps, the automated migration
+creates the `NudgeConfig` on the cluster, but git-stored Component
+manifests must also be updated manually. Users should remove the
+`build-nudges-ref` field from their Component definitions and add the
+generated `NudgeConfig` manifest to their repository to prevent
+GitOps sync conflicts during the deprecation period. For managed
+production clusters, a migration script must be provided to
+automate these GitOps repository changes at scale.
+
+Existing nudge PR customization mechanisms are handled as follows:
+
+- **`build-nudge-files`** (annotation on the push build PipelineRun,
+  not the Component CR) specifies file-pattern regexes for Renovate to
+  match when updating image references. As part of this migration,
+  integration-service deprecates this PipelineRun annotation in favor
+  of ConfigMap-based file-pattern configuration, consistent with
+  rcerven's recommendation and the overall move away from
+  annotation-driven config. Users who rely solely on this annotation
+  (without any ConfigMap-based configuration) must be explicitly
+  notified of the deprecation via release notes and documentation
+  updates, since the annotation has the lowest priority when ConfigMaps
+  are present and these users may not be aware of the ConfigMap
+  alternative.
+
+- **`build-nudge-simple-branch`** (annotation on the Component CR)
+  controls Renovate branch naming. Since integration-service takes
+  over nudge orchestration, it reads this annotation from the
+  Component and propagates it to the Renovate PipelineRun it creates.
+
+Renovate behavior customization is configured via a two-tier
+ConfigMap mechanism, which integration-service inherits ownership of:
+
+- **Namespace-wide default:** A ConfigMap named
+  `namespace-wide-nudging-renovate-config` (mandatory name) applies to
+  all nudged components in the namespace. Supported keys include
+  `automerge`, `commitMessagePrefix`, `commitMessageSuffix`,
+  `fileMatch`, `automergeType`, `platformAutomerge`, `ignoreTests`,
+  `gitLabIgnoreApprovals`, `automergeSchedule`, and `labels`.
+
+- **Per-component override:** If the nudged (target) component has
+  annotation `build.appstudio.openshift.io/nudge_renovate_config_map`,
+  the named ConfigMap is used instead of the namespace-wide default.
+
+Note: user-created ConfigMaps are not mounted directly into the
+Renovate PipelineRun. The orchestrating controller reads the
+appropriate ConfigMap, merges the options into the generated Renovate
+configuration, and saves the final config as a temporary ConfigMap
+that is mounted into the PipelineRun and garbage-collected with it.
+Integration-service preserves this behavior.
+
+#### Behavioral Changes from ADR 29
+
+Under ADR 29, integration-service skips all testing for nudging
+components — builds are promoted to the GCL but never tested,
+deployed, or released directly. This ADR preserves that behavior for
+`immediate` mode: the nudging component's build is not independently
+tested before firing the nudge.
+
+In `validated` mode, this behavior changes: the nudging component's
+build is included in a Snapshot that must pass integration tests before
+the nudge fires. Users switching an entry from `immediate` to
+`validated` should be aware that this enables testing for the nudging
+component, which may require configuring appropriate
+IntegrationTestScenarios.
+
+### Stale Reference Handling
+
+A controller in the integration-service reconciles the `NudgeConfig`
+whenever:
+
+- A `Component` CR is deleted from the namespace.
+- The `NudgeConfig` itself is modified.
+
+On reconciliation:
+
+1. List all `Component` CRs in the namespace.
+2. Validate that every `from` and `to` reference in `spec.nudges` maps
+   to an existing component.
+3. Update `status.conditions` to report any stale references.
+4. Stale entries are **not** automatically deleted — the controller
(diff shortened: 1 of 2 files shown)


ADR/0067-nudging-relationship-singleton-crd.md
```diff
@@ -0,0 +1,282 @@
...
+  resource. The singleton design must account for namespaces with many
+  component relationships.
+
+- **Stale data.** When components are deleted, the nudging configuration
+  must not retain references to non-existent components, otherwise
+  downstream nudge attempts will fail or produce confusing errors.
+
+## Decision
+
+Introduce a new **`NudgeConfig`** Custom Resource Definition, implemented
```
mmorhun, line 112: I'd explicitly say that it's optional.

ADR/0067-nudging-relationship-singleton-crd.md
```diff
@@ -0,0 +1,282 @@
...
+```
+
+### Fields
+
+| Field | Type | Description |
+|---|---|---|
+| `spec.nudges` | `[]NudgeRelationship` | List of directed component-to-component nudge edges. |
+| `spec.nudges[].from` | `string` | Name of the source component whose build triggers the nudge. |
+| `spec.nudges[].to` | `string` | Name of the target component to be nudged. |
+| `spec.nudges[].mode` | `enum` | `immediate`: nudge fires after the source build PLR succeeds. `validated`: nudge fires after the snapshot containing the source component passes integration tests. |
```
mmorhun: Is `mode` field optional? What is the default value?

ADR/0067-nudging-relationship-singleton-crd.md
```diff
@@ -0,0 +1,282 @@
...
+| `status.conditions` | `[]metav1.Condition` | Reports validation state — specifically whether all referenced components still exist. |
+| `status.lastValidationTime` | `metav1.Time` | Timestamp of the last stale-reference validation run. |
+
+### Singleton Enforcement
+
+The CRD uses a validating webhook (or CEL validation rule) to enforce
```
mmorhun: A dedicated webhook can be a potential bottleneck, especially at scale.
If we need to enforce it, should we consider Kyverno rules?
Does it make sense to check resource named `nudge-config`  and just ignore all others? (However, such behavior will allow cluster pollution with invalid objects).

ADR/0067-nudging-relationship-singleton-crd.md
```diff
@@ -0,0 +1,282 @@
...
+that only one `NudgeConfig` can exist per namespace with the well-known
+name `nudge-config`. Attempts to create a second instance are rejected.
+
+### Nudging Flow
+
+#### Immediate Mode (`mode: immediate`)
+
+```
+Build PLR succeeds for component-a
```
mmorhun, line 214: Is it only for push pipelines? If so, it needs to be stated explicitly.

ADR/0067-nudging-relationship-singleton-crd.md
```diff
@@ -0,0 +1,282 @@
...
+  → Snapshot is created with latest GCL images
+  → Integration tests run against the Snapshot
+  → All tests pass (Snapshot is marked as successful)
+  → integration-service reads NudgeConfig
+  → finds edge: component-a → component-b (validated)
+  → triggers Renovate nudge PipelineRun for component-b
+    using the validated image digest from the passing Snapshot
+```
+
+The validated mode leverages the Global Candidate List (GCL) and
```
mmorhun, line 237: Will it help in the following situation: there is an operator bundle with a few references. We need to build bundle only after all references updated.
(This might worth a dedicated feature, though).

ADR/0067-nudging-relationship-singleton-crd.md
```diff
@@ -0,0 +1,282 @@
...
+   immediate` (preserving current behavior).
+3. After migration is confirmed, the `build-nudges-ref` field is
+   deprecated and eventually removed from the Component spec.
+
+Users who want validated nudging can then update their `NudgeConfig` to
+change specific edges to `mode: validated`.
+
+### Stale Reference Handling
+
+A controller reconciles the `NudgeConfig` whenever:
```
mmorhun: In which operator the controller should be? Integration Service?

ADR/0067-nudging-relationship-singleton-crd.md
```diff
@@ -0,0 +1,282 @@
...
+
+The validated mode leverages the Global Candidate List (GCL) and
+Snapshot infrastructure to eliminate redundant downstream builds. Instead
+of nudging N times for N upstream builds, the integration-service waits
+for the Snapshot to pass and fires a single nudge with the validated
+image.
+
+### Migration from `build-nudges-ref`
+
+The integration-service will perform a one-time migration:
```
mmorhun, line 251: What about migration for gitops users?

ADR/0067-nudging-relationship-singleton-crd.md
```diff
@@ -0,0 +1,282 @@
...
+    using the validated image digest from the passing Snapshot
+```
+
+The validated mode leverages the Global Candidate List (GCL) and
+Snapshot infrastructure to eliminate redundant downstream builds. Instead
+of nudging N times for N upstream builds, the integration-service waits
+for the Snapshot to pass and fires a single nudge with the validated
+image.
+
+### Migration from `build-nudges-ref`
```
mmorhun, line 249: It says nothing about the nudge PR config map ([docs](https://konflux-ci.dev/docs/building/component-nudges/#customizing-nudging-prs)). Is it remain unchanged?

ADR/0067-nudging-relationship-singleton-crd.md
dirgim: I think we want to keep this open for a potential future feature where we would support in-PR nudging.

ADR/0067-nudging-relationship-singleton-crd.md
dirgim: My vote is for the `immediate` mode to be the default value.

ADR/0067-nudging-relationship-singleton-crd.md
gbenhaim: 1. we shouldn't add dependencies on Kyverno, since upstream it's not a dependency of Konflux.
2. A singleton per namespace can be achieved using a constrains on the name of the resource (defined in the CRD).
3. If any policy is needed, we can use the built in policy engine the k8s api provides - https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/

ADR/0067-nudging-relationship-singleton-crd.md
```diff
@@ -0,0 +1,282 @@
...
+
+### Nudging Flow
+
+#### Immediate Mode (`mode: immediate`)
+
+```
+Build PLR succeeds for component-a
+  → integration-service reads NudgeConfig
+  → finds edge: component-a → component-b (immediate)
+  → triggers Renovate nudge PipelineRun for component-b
```
gbenhaim, line 217: In which namespace this pipeline will run ?

ADR/0067-nudging-relationship-singleton-crd.md
gbenhaim: This is a great comment. Today this is a pain point for operator teams, that want to update the bundle once all the operands has new builds, and not multiple times, one for each component.

Is this ADR going to help us solve this issue?

ADR/0067-nudging-relationship-singleton-crd.md
```diff
@@ -0,0 +1,282 @@
...
+4. Stale entries are **not** automatically deleted — the controller
+   reports them via status conditions so the user or UI can remediate.
+   This avoids silent data loss if a component is temporarily removed
+   and re-created.
+
+### 1 MB Size Limit Considerations
+
+Each nudge entry is approximately 100–150 bytes of JSON. A 1 MB resource
+can hold roughly 7,000–10,000 nudge relationships. Production data shows
+the largest namespaces have fewer than 100 components, making the
```
gbenhaim: I don't think we can rely on those numbers. We've got requests to support thousands of components per namespace.

ADR/0067-nudging-relationship-singleton-crd.md
```diff
@@ -0,0 +1,282 @@
...
+Date: 2026-05-11
+
+## Status
+
+Proposed
+
+Partially supersedes [ADR 29. Component Dependencies](0029-component-dependencies.html)
+with respect to where nudging relationships are stored. The nudging
+*behavior* described in ADR 29 remains valid; this ADR changes only the
+storage mechanism and ownership.
```
gbenhaim: **Concern — scope ambiguity: "storage" vs "execution" ownership**

The Status section says this ADR changes "only the storage mechanism and ownership," but the Immediate Mode flow (lines 154–159) shows the **integration-service** triggering Renovate — which is currently build-service's responsibility per ADR 29.

This is a bigger change than moving storage — it moves the nudging *execution* too. The ADR should either:
- Clarify that execution ownership also moves to integration-service and update this section accordingly, or
- Keep immediate mode execution with build-service (which would just read from NudgeConfig instead of Component CR)

ADR/0067-nudging-relationship-singleton-crd.md
```diff
@@ -0,0 +1,282 @@
...
+Build PLR succeeds for component-a
+  → integration-service updates GCL in ComponentGroup
+  → Snapshot is created with latest GCL images
+  → Integration tests run against the Snapshot
+  → All tests pass (Snapshot is marked as successful)
+  → integration-service reads NudgeConfig
+  → finds edge: component-a → component-b (validated)
+  → triggers Renovate nudge PipelineRun for component-b
+    using the validated image digest from the passing Snapshot
+```
```
gbenhaim, line 235: **Concern — validated mode: which ComponentGroup's Snapshot?**

This flow says "Snapshot is created → tests run → all tests pass → nudge fires." But a component can belong to **multiple ComponentGroups** (that's the core premise of ADR 60).

The ADR doesn't specify:
- Which ComponentGroup's Snapshot must pass before the nudge fires?
- Do *all* ComponentGroups containing the component need passing Snapshots?
- Is there a way to configure which ComponentGroup gates the nudge?

This ambiguity should be resolved since it's central to the validated mode's value proposition.

ADR/0067-nudging-relationship-singleton-crd.md
```diff
@@ -0,0 +1,282 @@
...
+| `spec.nudges[].to` | `string` | Name of the target component to be nudged. |
+| `spec.nudges[].mode` | `enum` | `immediate`: nudge fires after the source build PLR succeeds. `validated`: nudge fires after the snapshot containing the source component passes integration tests. |
+| `status.conditions` | `[]metav1.Condition` | Reports validation state — specifically whether all referenced components still exist. |
+| `status.lastValidationTime` | `metav1.Time` | Timestamp of the last stale-reference validation run. |
+
+### Singleton Enforcement
+
+The CRD uses a validating webhook (or CEL validation rule) to enforce
+that only one `NudgeConfig` can exist per namespace with the well-known
+name `nudge-config`. Attempts to create a second instance are rejected.
```
gbenhaim: **Concern — no cycle detection**

ADR 29 explicitly requires DAG enforcement: "Application-service is responsible for making sure that cycles are not introduced." This ADR doesn't mention cycle detection at all.

A cycle (A→B→A) would cause infinite nudge loops. The validating webhook should reject cycles, and this should be stated here.

Also worth addressing in this section:
- **Duplicate entry validation**: can a user create two entries for the same `(from, to)` pair with different modes? The webhook should enforce uniqueness of `(from, to)` pairs.
- **Self-nudge guard**: `from == to` is nonsensical and could cause loops. The webhook should reject it.

ADR/0067-nudging-relationship-singleton-crd.md
```diff
@@ -0,0 +1,282 @@
...
+Introduce a new **`NudgeConfig`** Custom Resource Definition, implemented
+as a **singleton per namespace**, to store all component nudging
+relationships for that namespace. Ownership of nudging relationship
+storage moves from the build-service (`Component` CR) to the
+integration-service.
+
+### CRD Design
+
+```yaml
+apiVersion: appstudio.redhat.com/v1alpha1
```
gbenhaim: **Suggestion — API group**: ADR 60 (ComponentGroups) uses `konflux-ci.dev/v1alpha1`. As a brand-new CRD, `NudgeConfig` should probably use the newer API group unless there's a specific reason to stay on `appstudio.redhat.com`.

ADR/0067-nudging-relationship-singleton-crd.md
```diff
@@ -0,0 +1,282 @@
...
+several problems for the current nudging storage:
+
+1. **Ownership boundary violation.** Under the revised model, the
+   build-service owns the `Component` CR while the integration-service
+   owns orchestration and testing. Keeping nudging relationships on the
+   Component spec forces the build-service to understand dependency
+   semantics it should not own.
+
+2. **Redundant builds.** The current immediate-nudge-on-build model
+   triggers downstream builds before upstream tests have passed. When
```
gbenhaim, line 59: **Minor nit — Application-scoped → Namespace-scoped broadening**

ADR 29 explicitly states nudge relationships are **Application-scoped**: "Declared relationships between Components in different Applications are not to be considered valid."

This ADR moves to namespace-scoped, which is broader — components from different former-Applications could now nudge each other. This broadening is likely fine given the Application deprecation (ADR 60), but it's a semantic change worth acknowledging explicitly.

ADR/0067-nudging-relationship-singleton-crd.md
```diff
@@ -0,0 +1,282 @@
...
+
+1. Read existing `build-nudges-ref` values from all `Component` CRs in
+   each namespace.
+2. Generate the corresponding `NudgeConfig` singleton with `mode:
+   immediate` (preserving current behavior).
+3. After migration is confirmed, the `build-nudges-ref` field is
+   deprecated and eventually removed from the Component spec.
+
+Users who want validated nudging can then update their `NudgeConfig` to
+change specific edges to `mode: validated`.
```
gbenhaim, line 261: **Suggestion — migration doesn't address behavioral changes**

This section covers data migration but not the behavioral change. ADR 29 says integration-service **always skips testing** for nudging components. Under the new model with `validated` mode, that behavior changes — integration-service needs to run tests and *then* nudge.

Consider clarifying that:
1. All migrated entries default to `mode: immediate` (preserving ADR 29 behavior) — this is stated, which is good.
2. Switching to `validated` mode changes the testing semantics for the nudging component (tests are no longer skipped).
3. Whether the ADR 29 "skip testing for nudging components" behavior is preserved or retired under the new model.

gbenhaim (commented):
_This review was posted by **Gal's Cursor**._

ADR/0067-nudging-relationship-singleton-crd.md
kasemAlem: Great point — this is a real pain point for operator teams today. Phase 1 (immediate mode) doesn't solve it: each operand build fires a separate nudge, so the bundle rebuilds N times. Validated mode         
partially helps by gating on test results, but doesn't batch multiple upstream builds into a single nudge.
                                                                                                                                                                                                                 
Added "batched nudging for operator bundles" as an explicit Phase 2 item. The singleton NudgeConfig design doesn't prevent adding a batch/aggregate mode later — it would be an additional mode alongside      
 immediate and validated.

ADR/0067-nudging-relationship-singleton-crd.md
kasemAlem: The nudge PipelineRun runs in the tenant namespace, consistent with how build and integration PipelineRuns work today. Clarified in both the immediate and validated mode flow diagrams

ADR/0067-nudging-relationship-singleton-crd.md
kasemAlem: Agreed — removed all production-specific numbers. The etcd argument now rests on the design's inherent properties: one singleton per namespace, only in namespaces that use nudging, with each entry at        
 ~100-150 bytes. Also added a maxItems: 5000 constraint on the CRD schema to enforce the 1 MB limit structurally, so it holds regardless of how many components a namespace scales to.

ADR/0067-nudging-relationship-singleton-crd.md
kasemAlem: Good catch — since a component can belong to multiple ComponentGroups, the gating Snapshot is indeed ambiguous without an explicit specifier.                                                                  
                                                          
  We've added a gatingGroup field to NudgeRelationship that names the ComponentGroup whose Snapshot must pass before the nudge fires. It is required when mode: validated and ignored for immediate.             
                                                         
  For Phase 1 (KONFLUX-3524), we implement immediate mode only — component-to-component, no gating Snapshot involved. The gatingGroup field is reserved in the schema but not enforced until Phase 2. This gives 
 us a clear design path for validated mode without blocking the initial implementation.
                                                                                                                                                                                                                 
 Updated in the CRD example, fields table, and validated mode flow

ADR/0067-nudging-relationship-singleton-crd.md
kasemAlem: Agreed — this was a gap. Added a Validation Rules section to the ADR. The validating webhook now enforces:                                                                                                     
 1. Transitive cycle detection (A→B→C→A) — inherits DAG enforcement from Application-service, which is being deprecated
 2. Self-nudge rejection (from == to)                                                                                                                                                                           
 3. Duplicate (from, to) pair rejection — one entry per pair, update to change mode
 4. Component existence at creation time — stale references post-creation handled separately by the stale reference controller

ADR/0067-nudging-relationship-singleton-crd.md
kasemAlem: Agreed — NudgeConfig is a new CRD with no legacy baggage. Updated the API group to konflux-ci.dev/v1alpha1 to align with ComponentGroup (ADR-60)

ADR/0067-nudging-relationship-singleton-crd.md
kasemAlem: Good call — this is an intentional broadening, now explicitly acknowledged. Added problem #5 in the Context section explaining that Application-scoping is no longer meaningful given the Application CR       
deprecation (ADR-56/60), and that namespace-level scoping aligns with Konflux's existing tenant isolation boundary.

ADR/0067-nudging-relationship-singleton-crd.md
kasemAlem: Good point — added a "Behavioral Changes from ADR 29" section under Migration. immediate mode preserves the ADR-29 "skip testing for nudging components" behavior. validated mode explicitly changes it: the   
  nudging component's build is included in a Snapshot that must pass tests before the nudge fires. Users switching modes need to be aware this enables testing and may require configuring 
  IntegrationTestScenarios.

ADR/0067-nudging-relationship-singleton-crd.md
kasemAlem: Correct — this is a bigger change than just storage. Updated the Status section to explicitly state that both storage and execution ownership move to integration-service.                                     
                     
  The split is:                                                                                                                                                                                                  
  - integration-service owns the NudgeConfig CRD and orchestrates when/whether a nudge fires (both immediate and validated modes)
  - build-service retains ownership of the build pipeline and Renovate tooling — integration-service tells it "nudge component-b with this digest," build-service executes the Renovate PipelineRun              
                                                                                                                                                                                                                  
  The previous wording ("changes only the storage mechanism") was misleading — fixed.

ADR/0067-nudging-relationship-singleton-crd.md
kasemAlem: The nudge PR customization ConfigMaps remain unchanged from a user perspective — same annotations, same mounting to Renovate PipelineRuns. Ownership moves to integration-service consistent with the overall  
 nudging ownership transfer, but build-service continues to mount them as it does today. No user-facing changes. Added this clarification in the Migration section.

ADR/0067-nudging-relationship-singleton-crd.md
kasemAlem: Good point, 
added a GitOps migration note. The automated migration handles the cluster side, but GitOps users also need to update their git-stored manifests: remove build-nudges-ref from Component          
definitions and add the NudgeConfig to their repo to prevent sync conflicts.
wdyt?

ADR/0067-nudging-relationship-singleton-crd.md
kasemAlem: Yes, integration-service — it owns the NudgeConfig CRD, so the stale reference controller lives there too. Clarified in the text.

ADR/0067-nudging-relationship-singleton-crd.md
kasemAlem: Good question. Phase 1 targets push (post-merge) build PipelineRuns only, preserving current ADR-29 behavior. Added this explicitly in the Nudging Flow section. 
As @dirgim suggests, the design doesn't preclude extending to pull request builds in the future — the `NudgeConfig` schema is agnostic to the trigger type.

ADR/0067-nudging-relationship-singleton-crd.md
kasemAlem: good point @mmorhun a dedicated webhook adds unnecessary overhead for stateless checks. Updated the ADR: 
- Singleton enforcement now uses a CEL validation rule in the CRD schema (x-kubernetes-validations on metadata.name) — runs in-process in the API server, no webhook hop.                                      
- Stateless validations (self-nudge, duplicate pairs, cycle detection) also use CRD-level CEL.
- Stateful validations (component existence) use a ValidatingAdmissionPolicy or lightweight webhook, since they need to query the cluster. 

thanks @gbenhaim for the ideas

ADR/0067-nudging-relationship-singleton-crd.md
kasemAlem: `mode` is optional, defaults to `immediate`  . Updated the Fields table to state this explicitly, and updated the YAML example to show an entry without mode to illustrate the    
  default behavior. This also means migrated entries from build-nudges-ref can omit mode entirely and preserve current ADR-29 behavior.

ADR/0067-nudging-relationship-singleton-crd.md
kasemAlem: good point , 
 updated the Decision section to explicitly state that `NudgeConfig` is an optional singleton. Namespaces that don't use nudging don't need to create one.

ADR/0067-nudging-relationship-singleton-crd.md
mmorhun: Note, that operator will edit Components that are managed via gitops too.

mmorhun (approved):

ADR/0067-nudging-relationship-singleton-crd.md
```diff
@@ -0,0 +1,388 @@
...
+creates the `NudgeConfig` on the cluster, but git-stored Component
+manifests must also be updated manually. Users should remove the
+`build-nudges-ref` field from their Component definitions and add the
+generated `NudgeConfig` manifest to their repository to prevent
+GitOps sync conflicts during the deprecation period. For managed
+production clusters, a migration script must be provided to
+automate these GitOps repository changes at scale.
+
+Existing nudge PR customization mechanisms — including the
+`build-nudge-files` and `build-nudge-simple-branch` annotations on
```
rcerven: build-nudge-files is annotation on nudging push pipeline run
that should be probably deprecated in favor of configMaps

build-nudge-simple-branch is annotation on nudging component
and is used https://github.com/konflux-ci/build-service/blob/main/internal/controller/renovate_util.go#L546

ADR/0067-nudging-relationship-singleton-crd.md
```diff
@@ -0,0 +1,388 @@
...
+### CRD Design
+
+```yaml
+apiVersion: konflux-ci.dev/v1alpha1
+kind: NudgeConfig
+metadata:
+  name: nudge-config          # singleton — one per namespace
+  namespace: my-tenant
+spec:
+  nudges:
```
rcerven, line 128: configMap with users configurations are missing here

by default is used : 'namespace-wide-nudging-renovate-config' (the name is mandatory)

and if nudged component has annotation 'build.appstudio.openshift.io/nudge_renovate_config_map'
that component specific configmap will be used

ADR/0067-nudging-relationship-singleton-crd.md
```diff
@@ -0,0 +1,388 @@
...
+GitOps sync conflicts during the deprecation period. For managed
+production clusters, a migration script must be provided to
+automate these GitOps repository changes at scale.
+
+Existing nudge PR customization mechanisms — including the
+`build-nudge-files` and `build-nudge-simple-branch` annotations on
+the Component CR and any associated ConfigMaps — remain unchanged
+from a user perspective. Ownership of these ConfigMaps moves to the
+integration-service, consistent with the overall nudging ownership
+transfer. The ConfigMaps continue to be mounted to Renovate
```
rcerven: those configMaps aren't mounted to renovate pipeline run
but operator reads that ConfigMap and modifies default renovate options

final renovate config is https://github.com/konflux-ci/build-service/blob/main/internal/controller/renovate_util.go#L568
that config is saved as temporary CM, and that is actually mounted to renovate pipeline run

available customization options in users CM are in : https://github.com/konflux-ci/build-service/blob/main/internal/controller/renovate_util.go#L66
https://konflux.pages.redhat.com/docs/users/building/component-nudges.html#customizing-nudging-prs

ADR/0067-nudging-relationship-singleton-crd.md
kasemAlem: you are right, 
I missed this important issue, if controller edits GitOps managed ones to remove `build-nudges-ref` then operator and GitOps will **fight** and we don't want this happen. 

  **Option A** — Operator never edits Components. Migration is read-only: create NudgeConfig from existing data, then integration-service treats NudgeConfig as the authoritative source and simply ignores            
  build-nudges-ref. The field stays on Components until it's removed from the CRD schema in a later release. GitOps users update their git repos at their own pace during the deprecation window. No drift, no
  conflict.                                                                                                                                                                                                        
                  
  **Option B** — Two-track migration. Operator strips build-nudges-ref from non-GitOps Components only (detected by absence of GitOps annotations). GitOps Components are left alone with documentation telling users  
  to update their repos. More complex, more edge cases.
                                                                                                                                                                                                                   
  Option A is simpler and avoids the problem entirely — the operator never writes to Components, so there's no GitOps conflict. The tradeoff is that stale build-nudges-ref fields linger on Components until users
   clean them up, but since integration-service ignores them, it's cosmetic.

ADR/0067-nudging-relationship-singleton-crd.md
kasemAlem: Good catch — corrected. build-nudge-files is on the PipelineRun, not the Component. build-nudge-simple-branch is on the Component.                                                                                                                                                                                
 1. Fix the annotation locations
2. Specify that integration-service deprecates build-nudge-files in favor of ConfigMap-based file-pattern config as part of the migration                                                                      
 3. Explicitly state that integration-service inherits responsibility for reading build-nudge-simple-branch from the Component

ADR/0067-nudging-relationship-singleton-crd.md
kasemAlem: You're right that the ConfigMap customization mechanism should be documented. These ConfigMaps configure Renovate behavior (automerge, commit messages, file matching, etc.) rather than nudge relationships,  so they don't belong in the NudgeConfig CRD schema — but the ADR should be explicit about how integration-service inherits them.
                                                                                                                                                                                                                 
 Updated the Migration section to document the two-tier ConfigMap resolution:                                                                                                                                   
 - Namespace-wide default: namespace-wide-nudging-renovate-config
 - Per-component override via build.appstudio.openshift.io/nudge_renovate_config_map annotation on the nudged component                                                                                         
                                                                                                                                                                                                                 
  Integration-service takes ownership of reading these ConfigMaps when orchestrating nudge PipelineRuns.

ADR/0067-nudging-relationship-singleton-crd.md
kasemAlem: Good correction — updated the ADR. User-created ConfigMaps are not mounted directly. The orchestrating controller reads the appropriate ConfigMap (namespace-wide or per-component), merges the options into   
 the generated Renovate configuration, and saves the final config as a temporary ConfigMap that is mounted into the PipelineRun and garbage-collected with it. Integration-service preserves this behavior.

ADR/0067-nudging-relationship-singleton-crd.md
```diff
@@ -0,0 +1,416 @@
...
+GitOps sync conflicts during the deprecation period. For managed
+production clusters, a migration script must be provided to
+automate these GitOps repository changes at scale.
+
+Existing nudge PR customization mechanisms are handled as follows:
+
+- **`build-nudge-files`** (annotation on the push build PipelineRun,
+  not the Component CR) specifies file-pattern regexes for Renovate to
+  match when updating image references. As part of this migration,
+  integration-service deprecates this PipelineRun annotation in favor
```
rcerven, line 277: probably should be also mentioned that deprecation should be communicated with users (for those who are using just that annotation and not any configMaps) (as this annotation has lowest priority after configMaps)

kasemAlem: important point,  thanks 
I've added to the ADR

rcerven (approved):

arewm:
@kasemAlem , have you announced this ADR on the community call yet? The ADR process requests that you do that before reviews start.
https://github.com/konflux-ci/community/blob/main/ADRs.md#process

kasemAlem:
> @kasemAlem , have you announced this ADR on the community call yet? The ADR process requests that you do that before reviews start. https://github.com/konflux-ci/community/blob/main/ADRs.md#process

I've added topic to the next community meet agenda, 
thanks @arewm for the heads up

dirgim (approved):
LGTM

arewm (approved):

ralphbean (approved):

cit1zen (approved):

PR #357, opened 2026-05-14
ADR 0070: Support for zstd:chunked image compression

Amanyd:
Draft ADR 0070 proposing dual-compression for container builds.


ADR/0068-dual-compression-for-container-builds.md
```diff
@@ -0,0 +1,79 @@
...
+
+### Separate intermediary task
+
+A new task could accept the gzip manifest digest, pull the image, and re-push it with zstd compression. This works for provenance, but adds complexity, a new task definition, new pipeline wiring, and an extra cycle. Pushing twice from the existing buildah task is simpler and keeps everything in one place.
+
+## Consequences
+
+Pros:
+
+* Faster pulls for modern clients. Users with Podman, CRI-O, or Docker 25+ will automatically benefit from zstd:chunked without changing their pipeline configuration.
```
chmeliik: > automatically benefit from zstd:chunked without changing their pipeline configuration

I'm missing how. The doc mentions a `io.github.containers.compression.zstd` annotation but doesn't say where this annotation should be set and which Task should set it

ADR/0067-dual-compression-for-container-builds.md
```diff
@@ -0,0 +1,79 @@
...
+
+The problem is backward compatibility. Older Docker versions (pre-[23.0](https://docs.docker.com/engine/release-notes/23.0/)) try to read zstd layers as gzip and fail with `archive/tar: invalid tar header`. Other tools like [buildah](https://github.com/containers/buildah) (1.36+), [podman](https://github.com/containers/podman) (5.1+), [skopeo](https://github.com/containers/skopeo) (1.15+), and [CRI-O](https://github.com/cri-o/cri-o) already support zstd. Docker is the main blocker.
+
+As described in [build-definitions#1264](https://github.com/konflux-ci/build-definitions/issues/1264), pushing both gzip and zstd:chunked variants into the same OCI image index works as long as gzip entries are listed first. Old Docker clients pick the first platform match (gzip) and ignore the zstd entries. Podman reads the `io.github.containers.compression.zstd` annotation and prefers zstd automatically.
```
chmeliik: > Old Docker clients pick the first platform match (gzip) and ignore the zstd entries.

> Podman reads the `io.github.containers.compression.zstd` annotation and prefers zstd automatically

documentation/source code link needed

ADR/0067-dual-compression-for-container-builds.md
```diff
@@ -0,0 +1,79 @@
...
+
+Add a dual-compression setting to the build pipeline. Push both gzip and zstd:chunked formats from the `buildah` task and expose both digests as Tekton results. Dual-compression also risks breaking release pipeline tasks that assume one manifest per architecture, so those need deduplication fixes before the default can change.
+
+### Implementation details
+
+The `buildah` task will get 2 new params:
+* `COMPRESSION_FORMAT` — `gzip` (default), `zstd-chunked`, or `dual`
+* `FORCE_COMPRESSION` — when `true`, recompresses all layers, including the base image layers
+
+In `dual` mode, the push step pushes the image twice, once with gzip, once with `--compression-format zstd:chunked`. Each push produces a different manifest digest. Both are exposed as task results (`IMAGE_DIGEST` for gzip, `IMAGE_DIGEST_ZSTD` for zstd). Tekton Chains uses suffix matching on result names, so it will generate provenance for both variants without any changes in its code.
```
chmeliik: > (`IMAGE_DIGEST` for gzip, `IMAGE_DIGEST_ZSTD` for zstd)

This will break pipelines if users set `COMPRESSION_FORMAT=zstd-chunked`. Tasks that are downstream of the build task accept the `IMAGE_DIGEST` result. If there's no gzip image, this result will be empty/missing and the rest of the pipeline will fail.

chmeliik: > Tekton Chains uses suffix matching on result names, so it will generate provenance for both variants without any changes in its code.

documentation/source code link needed

ADR/0067-dual-compression-for-container-builds.md
```diff
@@ -0,0 +1,79 @@
...
+
+The `build-image-index` task will get a `ZSTD_IMAGES` array param. During index creation, gzip images are added first, then zstd images to enforce ordering. The `IMAGES` result continues to collect all digests via `buildah manifest inspect`. With dual-compression, it includes all 2N digests, which are already handled by Chains for provenance.
```
chmeliik: Is a separate parameter necessary? Would passing the images in the correct order not be enough?

```yaml
  - name: build-image-index
    params:
    - name: IMAGES
      value:
      - $(tasks.build-images.results.IMAGE_REF_ZSTD[*])
      - $(tasks.build-images.results.IMAGE_REF[*])
```

Side note: if the build task returns only an `IMAGE_DIGEST_ZSTD`, I don't think it will be possible to construct the pipeline correctly. I don't think Tekton is flexible enough to build an array like this
```yaml
      value:
      - $(tasks.build-images.results.IMAGE_URL)@$(tasks.build-images.results.IMAGE_DIGEST_ZSTD[*])
```

And an experiment is needed here to find out what happens if the build task returns an empty/missing IMAGE_REF_ZSTD

chmeliik: > includes all 2N digests

Are we sure all the results will fit in Tekton's small result limit?

ADR/0067-dual-compression-for-container-builds.md
chmeliik: How would the individual images be tagged? I believe both the compressions types would need a tag to avoid garbage collection. And the tags would need to be guaranteed unique to avoid race conditions.

ADR/0067-dual-compression-for-container-builds.md
chmeliik: What about pipelines that don't build an image index? If the user sets `dual`, they'll push manifests with both compressions, but the overall `IMAGE_URL@IMAGE_DIGEST` result of the pipeline will point to the gzip manifest?

ADR/0067-dual-compression-for-container-builds.md
```diff
@@ -0,0 +1,79 @@
...
+
+Four release tasks in [release-service-catalog](https://github.com/konflux-ci/release-service-catalog) call `get-image-architectures` and loop over all the manifests. With dual-compression they will see duplicate entries per architecture. Each task needs a `jq` dedup filter that groups by `(architecture, os)` and keeps the non-zstd entry:
```
chmeliik: > dedup filter that groups by `(architecture, os)` and keeps the non-zstd entry

Does this mean we will be building zstd compressed images but never releasing them?

Or will we be half-releasing them, where the actual released OCI indexes will include the zstd manifest, but the metadata in Pyxis and other details will be missing?

ADR/0067-dual-compression-for-container-builds.md
```diff
@@ -0,0 +1,79 @@
...
+* `filter-already-released-advisory-images` — would treat zstd variants as "unreleased" since they have new digests
+* `extract-oot-kmods` — would miscount architectures and fail extracting kmods from zstd layers
+
+Signing tasks are not affected, they sign all manifests in the index, which is correct for both variants.
+
+### Rollout
+
+The default value of `COMPRESSION_FORMAT` stays as `gzip` until the release task fixes are merged and tested. Only then will the default change to `dual`. Once dual-compression adoption is stable and older Docker versions are no longer a concern, the default can move to `zstd-chunked` only, dropping the gzip variant entirely.
+
+## Alternatives considered
```
chmeliik: How about an alternative where the build task pushes an index with both the gzip manifest and the zstd manifest instead of pushing them individually? The build-image-index task, if enabled, would still flatten the indexes into a single index with 2N entries.

That might solve the problems caused by bolting on an additional output image. What problems would it cause in return?

chmeliik:
How will all of this be represented in the SBOM? Currently, each architecture gets an SBOM where the arch-specific manifest is the main component, and the index gets an SBOM pointing to each arch-specific manifest.

Example:
```bash
# index SBOM
$ cosign download sbom quay.io/konflux-ci/task-runner:1.6.0

# x86_64 manifest SBOM
$ cosign download sbom quay.io/konflux-ci/task-runner@sha256:b4b46f35d697dd7cbcd77b051ceac41d324164edcf85d39379b7b53f492d8394
```

If we push two different images for each arch, which one will be the main component in the SBOM? How will the other one be represented? Will we need to attach the SBOM to both of them?

If we push a single index with two manifests for each arch, can that be the main component? Is it OK if the SBOM doesn't reference the individual manifests? Or can we represent them as sub-components of the main component (in both CycloneDX and SPDX)? Will we need to attach the SBOM to the index and both the manifests, or just the index?

ADR/0067-dual-compression-for-container-builds.md
eskultety: Can you elaborate? To support this I think we'll always need to create an image index.

chmeliik: This proposes that the buildah task would first push the two individual manifests and then build-image-index would assemble them into an index later. The individual manifests would need unique tags, because if a manifest is orphaned (not referenced by any index or any tag), then the container registry may consider it deleted (Quay does, I believe). So the ADR should specify what kind of tagging scheme will be used to ensure the manifests don't become orphaned.

Alternatively, I think this would not be a problem if the buildah task pushed an index right away as suggested in https://github.com/konflux-ci/architecture/pull/357#discussion_r3240828799

ADR/0068-dual-compression-for-container-builds.md
Amanyd: The annotation is set at the `build-image-index` task, its a behavior of `conatiners/image` library, it injects the annotation in the index descriptor

ADR/0067-dual-compression-for-container-builds.md
Amanyd: Updated

ADR/0067-dual-compression-for-container-builds.md
Amanyd: `IMAGE_DIGEST` will always hold the primary digest. So in `zstd-chunked` mode, the zstd digest goes into `IMAGE_DIGEST` and gzip digest for gzip or dual mode, `IMAGE_DIGEST_ZSTD` is only for the dual mode, I,ve updated the ADR to clarify this

Amanyd: I've added the link

chmeliik: Ack, that should solve part of the problem. The problem that remains is that in gzip-only mode, the build-image-index task may receive invalid input due to the missing `ZSTD` result. Let's discuss that in https://github.com/konflux-ci/architecture/pull/357#discussion_r3240727464

chmeliik: According to the docs, IMAGE_DIGEST should be the suffix, not the prefix

> In this approach, one can write the url and digest of an output OCI artifact into 2 results that have same prefix, but the one for url has suffix IMAGE_URL and the one for digest has suffix IMAGE_DIGEST

ADR/0067-dual-compression-for-container-builds.md
Amanyd: A full image reference is at most 150bytes, so a 4 arch build will be ~1200, well within the limit(4096), I've added a note in cons

eskultety: > A full image reference is at most 150bytes, so a 4 arch build will be ~1200, 

@Amanyd ^this will be best backed with a PoC, tekton results is something that we can estimate, but the limit is tight and only real world testing can tell to be sure I think :) .

chmeliik: +1, 4096 bytes is a theoretical maximum but Tasks can never actually use all of it in practice. And it's not a static limit, it depends on the number of steps in the task, the number of different results etc. We've seen the buildah task error out on exceeding the result size with less than 1K worth of data

> A full image reference is at most 150bytes

Where does the number come from?

chmeliik: It would also be worth investigating _why_ build-image-index even returns the `IMAGES` result. I think its purpose is to signal to Chains that it should sign the images. But then, the buildah task has already emitted all the results required for that, so why does build-image-index duplicate it?

ADR/0067-dual-compression-for-container-builds.md
eskultety: > Does this mean we will be building zstd compressed images but never releasing them?

ultimately no, I would consider that only interim until we figure everything out gradually. We may want to emphasize that.

ADR/0067-dual-compression-for-container-builds.md
eskultety: >  an index with both the gzip manifest and the zstd manifest instead of pushing them individually?

Can you do this at the same time though?

ADR/0067-dual-compression-for-container-builds.md
Amanyd: Yes, half releasing is the idea here, OCI indexes will include zstd manifests for users to pull

ADR/0067-dual-compression-for-container-builds.md
chmeliik: I believe so - if I'm not mistaken, `buildah manifest add` can add locally stored images to the index, and `buildah manifest push` pushes the index along with the referenced manifests

ADR/0067-dual-compression-for-container-builds.md
chmeliik: Ack, let's clarify that handling releases properly is out of scope here. If I understand correctly, the target state is that release pipelines don't fail and the resulting image indexes include zstd, but any other handling for zstd may not work properly

eskultety: >  If I understand correctly, the target state is that release pipelines don't fail and the resulting image indexes include zstd, but any other handling for zstd may not work properly

Sounds good! Releasing will likely need an updated ADR when it comes to it, fine with me.

ADR/0067-dual-compression-for-container-builds.md
eskultety: > if I'm not mistaken, buildah manifest add can add locally stored images to the index

I think this will need to be expanded onto multi-arch (which is a good chunk of Konflux workloads these days?), don't you have isolated buildahs for each arch, IOW can you still access all archful images within one task to be able to do this?

ADR/0067-dual-compression-for-container-builds.md
Amanyd: I've corrected it

ADR/0067-dual-compression-for-container-builds.md
chmeliik: The idea is that each archful task would, in the `dual` case, produce an image index with two entries. The `build-image-index` task would still be responsible for collecting the outputs of the archful tasks and merging them into a single, flat index.

eskultety: Makes sense.

chmeliik:
@Amanyd when making changes that address some open threads, would you be able to respond in those threads?

ADR/0067-dual-compression-for-container-builds.md
Amanyd: Dropping the separate zstd results, buildah now pushes the "per-arch index" combining both variants, and returns digest for the index

Amanyd: I tested on local cluster, buildah task's total result was 653 bytes, and since we are not adding any new result now, the size should stay same as of now, 

Emitting all the image manifest digests from build-image-index was added as a workaround to make tekton chains work properly when users are using the matrix feature, it does duplicate the results the buildah task already emitted but it's necessary for chains to sign everything correctly for builds across a matrix.

ADR/0067-dual-compression-for-container-builds.md
Amanyd: `buildah` will now push both zstd and gzip vaiants to temporary tags, and then combine in the per arch index, and since the final index references both manifests, they shouldn't get orphaned or garbage collected

ADR/0067-dual-compression-for-container-builds.md
Amanyd: With the per-arch index, single arch pipelines will also output an index instead of a manifest, pipeline result will just point to that index. so it should work even without build-image-index.

ADR/0067-dual-compression-for-container-builds.md
Amanyd: As per the suggestion, i've updated the ADR to use this "per-arch index" pattern. i've already prototyped the buildah task pushing the per-arch index locally, and it works, i'm currently testing the build-image-index task to make sure it flattens those indices correctly.

Amanyd:
> How will all of this be represented in the SBOM? Currently, each architecture gets an SBOM where the arch-specific manifest is the main component, and the index gets an SBOM pointing to each arch-specific manifest.
> 
> Example:
> 
> ```shell
> # index SBOM
> $ cosign download sbom quay.io/konflux-ci/task-runner:1.6.0
> 
> # x86_64 manifest SBOM
> $ cosign download sbom quay.io/konflux-ci/task-runner@sha256:b4b46f35d697dd7cbcd77b051ceac41d324164edcf85d39379b7b53f492d8394
> ```
> 
> If we push two different images for each arch, which one will be the main component in the SBOM? How will the other one be represented? Will we need to attach the SBOM to both of them?
> 
> If we push a single index with two manifests for each arch, can that be the main component? Is it OK if the SBOM doesn't reference the individual manifests? Or can we represent them as sub-components of the main component (in both CycloneDX and SPDX)? Will we need to attach the SBOM to the index and both the manifests, or just the index?

As for the per-arch index now, the SBOM will just attach directly to the index digest, because both variants only differ in compression, one SBOM on the index is all we need, we don't have to generate separate SBOMs for each variant.

ADR/0067-dual-compression-for-container-builds.md
```diff
@@ -0,0 +1,96 @@
...
+* `FORCE_COMPRESSION` — when `true`, recompresses all layers, including the base image layers
+
+In `dual` mode, the push step creates a **per-arch index** containing both compression variants:
+
+1. Push the gzip variant to a temporary tag (`${TASKRUN_NAME}-gzip`)
+2. Push the zstd:chunked variant to a temporary tag (`${TASKRUN_NAME}-zstd`)
+3. Create a per-arch index with `buildah manifest create` + `buildah manifest add`, adding the gzip manifest first
+4. Push the per-arch index to the unique tag and the real image tag
+
+Both variant pushes must use `--format=oci`. Mixing OCI and Docker manifest formats within a single index causes "manifest invalid" error on Quay.
```
ralphbean: What's the behavior if the user tries to mix non-oci format with a non-gzip compression format? Should the task detect that this combination of options is incompatible and error out early, telling the user that this is an invalid combination?

ADR/0067-dual-compression-for-container-builds.md
```diff
@@ -0,0 +1,96 @@
...
+* `create-pyxis-image` — would create duplicate Pyxis entries and doesn't have zstd decompression logic
+* `populate-release-notes` — would list the same architecture twice in advisory content
+* `filter-already-released-advisory-images` — would treat zstd variants as "unreleased" since they have new digests
+* `extract-oot-kmods` — would miscount architectures and fail extracting kmods from zstd layers
+
+Signing tasks are not affected; they sign all manifests in the index, which is correct for both variants.
+
+### Rollout
+
+The default value of `COMPRESSION_FORMAT` stays as `gzip` until the release task fixes are merged and tested. Only then will the default change to `dual`. Once dual-compression adoption is stable and older Docker versions are no longer a concern, the default can move to `zstd-chunked` only, dropping the gzip variant entirely.
```
ralphbean:
```suggestion
The default value of `COMPRESSION_FORMAT` stays as `gzip` until the release task fixes are merged and tested. Only then will the default change to `dual`. Once dual-compression adoption is stable and older Docker versions are no longer a concern, the default can move to `zstd-chunked` only, dropping the gzip variant entirely.

There may be additional considerations to take into account when deciding to switch the default to dual or zst-chunked. For example, if major installations are still requiring the docker format instead of oci for images, then changing this default might need to wait until those installations are ready to move.
```

ralphbean (commented):
This is really good @Amanyd - thank you for working on it, and welcome to the Konflux community!

The @konflux-ci/kgc group talked about this today and one thought that came up was that this is a good idea that would be even better if it were framed generally. I think it is implied that the change here is meant to happen in the buildah task, but what about other tasks in the konflux-ci ecosystem that also produce images? What about tasks that produce non-image OCI artifacts (like the oci-copy task or the modelcar task)? The ADR would be stronger if the decision read like "Add a dual-compression setting to _all_ build pipelines."

Now, we understand that this is your GSoC project and we don't want to ask you to be responsible for adding that setting to _all_ pipelines. Keep the scope of the project as you have it, but if the ADR can set an expectation for the authors of others tasks and pipelines to conform with this behavior, then it acts better as an "architectural decision" that we align multiple implementations to.

For an example of this kind of thinking, check out https://github.com/konflux-ci/architecture/pull/300 where @adambkaplan proposed an ADR about a specific kind of build pipeline (for maven artifacts), but then revised it into a general ADR about build pipelines (in general) in https://github.com/konflux-ci/architecture/pull/303

I don't think you need to massively revise your ADR here. Just note that other build tasks and pipelines should follow the convention you establish in your decision here and leave the implementation details for those other tasks and pipelines as an exercise for the reader.

ADR/0067-dual-compression-for-container-builds.md
chmeliik: > I tested on local cluster, buildah task's total result was 653 bytes, and since we are not adding any new result now, the size should stay same as of now,

I was more concerned about the build-image-index task, which would return twice as many IMAGES with zstd enabled. Or not?

ADR/0067-dual-compression-for-container-builds.md
```diff
@@ -0,0 +1,96 @@
...
+
+`IMAGE_DIGEST` points to the per-arch index, which contains both variants. Downstream tasks, Tekton Chains, SBOM generation, and signing all consume `IMAGE_DIGEST` without changes. The `io.github.containers.compression.zstd` annotation is set on the zstd manifest descriptor within the index, which is how Podman and CRI-O [identify and prefer](https://github.com/containers/image/blob/main/internal/manifest/oci_index.go) the zstd variant.
+
+In `gzip` or `zstd-chunked` mode, the buildah task pushes a single manifest as it does today. `IMAGE_DIGEST` holds the primary digest (gzip in `gzip` mode, zstd in `zstd-chunked` mode).
+
+The `build-image-index` task requires no new parameters. Each entry in the existing `IMAGES` array is now a per-arch index (in `dual` mode) instead of a single manifest. The task's existing `--all` flag on `buildah manifest add` extracts child manifests from per-arch indexes and flattens them into the final multi-arch index. The ordering within each per-arch index (gzip first) is preserved in the flat index. The `IMAGES` result continues to collect all manifest digests via `buildah manifest inspect`, and Tekton Chains generates provenance for each one.
```
chmeliik: I have a concern here related to one of the rejected alternatives:

> Buildah's `--add-compression` flag automatically creates zstd variants at index push time. This is simpler, but the zstd manifests are created during `buildah manifest push` in the `build-image-index` task. They are not results of any TaskRun, so Chains cannot generate provenance for them. At release time, Conforma would find manifests without provenance and block the release.

Will this not have the same problem? The only occurrence of the zstd/gzip manifests will be in the IMAGES result of the build-image-index task. Will Conforma be okay with that?

And perhaps more importantly, *should* Conforma be okay with that? This means you could pass arbitrary images to the input array of build-image-index, causing Chains to generate provenance for them and making it look as if they were legitimate outputs of the build

ADR/0070-dual-compression-for-container-builds.md
```diff
@@ -0,0 +1,96 @@
...
+
+Add a dual-compression setting to the build pipeline. In `dual` mode, the `buildah` task pushes both gzip and zstd:chunked variants, bundles them into a per-arch OCI image index, and exposes the index digest as `IMAGE_DIGEST`. No new Tekton results are needed. It also risks breaking release pipeline tasks that assume one manifest per architecture, so those need deduplication fixes before the default can change.
+
+### Implementation details
+
+The `buildah` task will get 2 new params:
+* `COMPRESSION_FORMAT` — `gzip` (default), `zstd-chunked`, or `dual`
+* `FORCE_COMPRESSION` — when `true`, recompresses all layers, including the base image layers
+
+In `dual` mode, the push step creates a **per-arch index** containing both compression variants:
```
chmeliik, line 43: Were you able to test what effect this has on the post-build tasks that process the output image? Are they able to process a gzip+zstd index instead of a single manifest?

ADR/0067-dual-compression-for-container-builds.md
```diff
@@ -0,0 +1,96 @@
...
+### Implementation details
+
+The `buildah` task will get 2 new params:
+* `COMPRESSION_FORMAT` — `gzip` (default), `zstd-chunked`, or `dual`
+* `FORCE_COMPRESSION` — when `true`, recompresses all layers, including the base image layers
+
+In `dual` mode, the push step creates a **per-arch index** containing both compression variants:
+
+1. Push the gzip variant to a temporary tag (`${TASKRUN_NAME}-gzip`)
+2. Push the zstd:chunked variant to a temporary tag (`${TASKRUN_NAME}-zstd`)
```
chmeliik: Nitpick: I don't think the tags are necessary. I believe the index + the manifests can be pushed with one command, it's not necessary to push the manifests before pushing the index.

ADR/0067-dual-compression-for-container-builds.md
```diff
@@ -0,0 +1,96 @@
...
+
+In `gzip` or `zstd-chunked` mode, the buildah task pushes a single manifest as it does today. `IMAGE_DIGEST` holds the primary digest (gzip in `gzip` mode, zstd in `zstd-chunked` mode).
+
+The `build-image-index` task requires no new parameters. Each entry in the existing `IMAGES` array is now a per-arch index (in `dual` mode) instead of a single manifest. The task's existing `--all` flag on `buildah manifest add` extracts child manifests from per-arch indexes and flattens them into the final multi-arch index. The ordering within each per-arch index (gzip first) is preserved in the flat index. The `IMAGES` result continues to collect all manifest digests via `buildah manifest inspect`, and Tekton Chains generates provenance for each one.
+
+SBOM generation and upload work without changes. The SBOM attaches to the per-arch index digest (via `IMAGE_REF`), which is the primary artifact. Individual compression variants within the index do not receive separate SBOMs; they share identical content, only differing in layer compression.
```
chmeliik: I think that makes sense conceptually, but does it work for Conforma validation and for Release pipelines? I suspect that Conforma validates the individual manifests within the index, and tries to get the SBOM for each of them.

And I'm not sure what exactly Release pipelines do with SBOMs, but they do have handling for them

ADR/0067-dual-compression-for-container-builds.md
Amanyd: They do occur in IMAGES result only, but they already exist in the registry before build-image-index task runs, created and pushed by buildah, in the rejected alternative, manifests don't exist until the push in build-image-index. Since build-image-index will see both manifests, it includes them both in IMAGES, chains should sign them and conforma should be okay, i'll verify this in testing

About the arbitrary images, it is a valid concern, but it is how builds work today, build-image-index trusts whatever is passed in the array, and chains signs them

ADR/0070-dual-compression-for-container-builds.md
Amanyd: Not yet, this is on the testing plan, downstream tasks handle indexes in multi arch builds, but i'll check for single arch case and report ASAP

ADR/0067-dual-compression-for-container-builds.md
Amanyd: Agreed, i'll need to work out on how to keep keep both variants locally before push, but temp tags can be avoided, i'll update the ADR

ADR/0067-dual-compression-for-container-builds.md
chmeliik: After thinking about it some more, maybe it doesn't make sense conceptually. The thing is that, in a multi-arch pipeline, the "per-arch index" only really exists within the build pipeline. It's not an output that anyone sees. The 4 per-arch indexes get merged into a single index with 8 entries, which are the individual manifests, not the per-arch indexes.

It would probably make more sense to attach all the metadata to the manifests directly, not to the per-arch index.

ADR/0067-dual-compression-for-container-builds.md
chmeliik: Okay, yeah let's test it. I'm kind of hoping Conforma will reject the build, because if it doesn't, then our trust model is broken

ADR/0067-dual-compression-for-container-builds.md
Amanyd: Agreed, we'll need to update buildah to attach SBOMs directly to the gzip manifests instead, i looked into the Conforma side and I think we're fine, it validates the final index from the snapshot, and SBOM checks are actually skipped for image indexes, For release tasks, we'll need to add deduplication for the doubled manifest entries and fix the SBOM attachment point, both as prerequisite PRs.

ADR/0067-dual-compression-for-container-builds.md
Amanyd: Good idea, will add validation check in the task, If COMPRESSION_FORMAT=dual and IMAGE_FORMAT is not oci, the task will throw an early error

ADR/0067-dual-compression-for-container-builds.md
Amanyd: I have updated the ADR accordingly

ADR/0067-dual-compression-for-container-builds.md
chmeliik: For my knowledge, what makes it impossible to have dual compression with the docker format?

Amanyd:
> This is really good @Amanyd - thank you for working on it, and welcome to the Konflux community!
> 
> The @konflux-ci/kgc group talked about this today and one thought that came up was that this is a good idea that would be even better if it were framed generally. I think it is implied that the change here is meant to happen in the buildah task, but what about other tasks in the konflux-ci ecosystem that also produce images? What about tasks that produce non-image OCI artifacts (like the oci-copy task or the modelcar task)? The ADR would be stronger if the decision read like "Add a dual-compression setting to _all_ build pipelines."
> 
> Now, we understand that this is your GSoC project and we don't want to ask you to be responsible for adding that setting to _all_ pipelines. Keep the scope of the project as you have it, but if the ADR can set an expectation for the authors of others tasks and pipelines to conform with this behavior, then it acts better as an "architectural decision" that we align multiple implementations to.
> 
> For an example of this kind of thinking, check out #300 where @adambkaplan proposed an ADR about a specific kind of build pipeline (for maven artifacts), but then revised it into a general ADR about build pipelines (in general) in #303
> 
> I don't think you need to massively revise your ADR here. Just note that other build tasks and pipelines should follow the convention you establish in your decision here and leave the implementation details for those other tasks and pipelines as an exercise for the reader.

Thank you! I've updated ADR to frame this as a convention, will checkout #300 and #303 ASAP.

ADR/0067-dual-compression-for-container-builds.md
Amanyd: > For my knowledge, what makes it impossible to have dual compression with the docker format?

docker format spec defines media type only for gzip layer, so we need oci for zstd, and in testing, quay rejected mixing both formats in same index

ADR/0067-dual-compression-for-container-builds.md
arewm: We ran into issues where we overflowed the Tekton result by surfacing too many pullspecs in the task results. How likely are we to run into this issue with the current design? 

I assume that the build-image-index task will have the biggest potentially concerning result. It has fewer task steps so we should have more space available for the data at least.

ralphbean:
So — I want to walk through where I think we are on the Conforma question, because @chmeliik raised something important and I think it points us toward a slightly different shape for this.

The core tension: in `dual` mode, the buildah task pushes three things — a gzip manifest, a zstd manifest, and a per-arch index that bundles them. But only the per-arch index digest goes into `IMAGE_DIGEST` as a Tekton result. The individual gzip and zstd manifest digests surface later, when `build-image-index` enumerates them from the registry and puts them in its `IMAGES` result.

That means Chains generates provenance for those manifests based on `build-image-index` discovering them, not based on the buildah task declaring them as outputs. As @chmeliik pointed out — `build-image-index` could put any arbitrary digest in `IMAGES` and Chains would sign it. If Conforma accepts that, we have a trust model gap that exists independent of this ADR. If Conforma rejects it, then the zstd manifests won't pass validation and dual mode doesn't work as described.

Either way, the ADR shouldn't assert "No changes needed to Chains or Conforma" without resolving this.

We talked about this on the @konflux-ci/kgc call today and @arewm's suggestion is that we should unblock this by having the buildah task intentionally expose all three references — the per-arch index digest, the gzip manifest digest, and the zstd manifest digest — as Tekton results. That way Chains sees all three directly from the build task and generates provenance for each one. No ambiguity about where they came from.

The catch is the Tekton results size limit. A full image reference is ~150 bytes, and with three results per architecture instead of one, you'll hit the buffer ceiling faster — especially in matrix builds. We've seen the buildah task overflow with less than 1K of result data before. So this approach works for common cases but may not work for builds that are already tight on result space.

I think we should be explicit in the ADR that this is an experimental approach with a known constraint. https://github.com/tektoncd/community/pull/1248 is a TEP in progress that would relax the results size limit, which would resolve this pressure longer term. Until then, dual compression is opt-in and best-effort — users whose builds overflow can stay on `gzip`.

Does that framing work for everyone? @Amanyd would you be up for revising the implementation details and consequences sections along these lines?

arewm:
Thanks for that, Ralph. To be precise, the buildah tasks can just return `IMAGES` referencing the image manifests. Considering a pipeline run with only a buildah task (and not a build-image-index task), if Chains was configured to generate TaskRun provenance, it would be able to see the image index from the `IMAGE_*` results and then the image manifests from the `IMAGES` result. 

https://github.com/tektoncd/chains/issues/1070 is also a related issue.

arewm:
for transparency, I moved this ADR to the milestone next week: https://github.com/konflux-ci/architecture/milestone/28

I think that this ADR is largely architecturally sound. With the above comment, I think that this ADR can likely be approved after finalizing the remaining implementation detail and limitations. This finalization can likely happen in parallel with implementation of the work if desired.

chmeliik:
> I think we should be explicit in the ADR that this is an experimental approach with a known constraint. https://github.com/tektoncd/community/pull/1248 is a TEP in progress that would relax the results size limit, which would resolve this pressure longer term. Until then, dual compression is opt-in and best-effort — users whose builds overflow can stay on gzip.
> 
> Does that framing work for everyone?

@ralphbean +1, the framing and the suggested solution makes sense to me

> To be precise, the buildah tasks can just return IMAGES referencing the image manifests.

+1, that would be slightly better, since AFAIK even the results _names_, not just values, count towards the size limit (implementation detail though, not blocking)

@Amanyd if you're able to test what Conforma says about the proposed implementation (https://github.com/konflux-ci/architecture/pull/357#discussion_r3308965607), I would still be interested. It might mean we need to follow up outside of this ADR

Amanyd:
@ralphbean  @arewm  @chmeliik  Thank you for the discussion and the solution, I've updated the ADR along these lines, decision section now frames dual-compression as experimental, with the Tekton result size limit called out as a known constraint, I've also updated the Consequences to no longer claim Conforma accepts this, I ran a test build, conforma actually rejected it, The policies don't yet understand how to validate the child manifests alongside the index, downstream task `rpms-signature-scan` too crashed trying to extract zstd layer, we will need to follow up on these updates outside of this ADR
The buildah task's results in dual mode come out to ~700 bytes for a single-arch build, build-image-index only returns the index digest in its IMAGES (1 entry), so it doesn't grow further, Users hitting the ceiling in large matrix builds can stay on gzip until the TEP lands

ADR/0070-dual-compression-for-container-builds.md
arewm: If we cannot resolve this in the ADR then I think it would be fine to call it out as an unknown. Would it fit as a potential consequence or would it be better elsewhere?

ADR/0070-dual-compression-for-container-builds.md
```diff
@@ -0,0 +1,105 @@
...
+* Full backward compatibility. Old Docker clients continue to work because gzip is listed first in the image index.
+* Provenance and signing. Because the build task explicitly exports the child manifests in `IMAGES`, Tekton Chains can generate provenance for each manifest directly from the trusted build task, avoiding trust gaps where manifests are only discovered by downstream tasks.
+
+Cons:
+
+* The final image index contains 2N manifests instead of N (one per architecture per compression format). Exposing all of these in the `IMAGES` result increases the Tekton result size, pushing closer to the 4KB limit for large matrix builds (e.g., builds exceeding 4-5 architectures may hit this limit).
+* Registry storage per image roughly doubles. Each layer is stored in both gzip and zstd formats. Users who don't want this can keep `COMPRESSION_FORMAT=gzip`.
+* Builds take longer in dual mode. The zstd push recompresses all layers, adding time proportional to image size.
```
arewm, line 110: By default, we only add one more layer to the images. If we are recompressing all layers, will we lose this connection to the parent image?

I imagine so if the parent image doesn't have zstd:chunked, but if the parent image has that compression available, will this still be a consequence (longer builds due to recompressing) or will we be able to take advantage of the parent images' builds?

Amanyd: I think yes, only when the parent image doesn't already have zstd:chunked layers, in that case, the parent layers need to be recompressed, when zstd:chunked layers are already available, they can be reused directly, so we still get the layer sharing benefits and only new layers need compression.

arewm: Is there any additional logic needed to reuse the zstd:chunked compression from the parent image or that "just works?"

Amanyd: After looking into it more, the parent's zstd layers actually cannot be reused, Buildah stores all layers uncompressed locally and compresses them fresh at push time, when we specify --compression-format zstd:chunked, it forces recompression of all layers regardless, there's no way to check the registry for existing zstd blobs and skip the work, so the "longer builds" consequence holds either way

ADR/0070-dual-compression-for-container-builds.md
```diff
@@ -0,0 +1,105 @@
...
+
+In `gzip` or `zstd-chunked` mode, the build task pushes a single manifest as it does today. `IMAGE_DIGEST` holds the primary digest (gzip in `gzip` mode, zstd in `zstd-chunked` mode).
+
+The `build-image-index` task requires no new parameters. Each entry in the existing `IMAGES` array is now a per-arch index (in `dual` mode) instead of a single manifest. The task's existing `--all` flag on `buildah manifest add` is expected to extract child manifests from per-arch indexes and flatten them into the final multi-arch index, preserving the gzip-first invariant or explicit flattening logic will be added in the task if needed.
+
+SBOM generation requires an update; the SBOM must attach to both the gzip and zstd child manifest digests. The per-arch index is an intermediate build artifact that gets flattened by `build-image-index`, so it is not a valid long-term attachment point. Both compression variants receive the same SBOM since they share identical content, only differing in layer compression.
+
```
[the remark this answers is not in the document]
Amanyd: Ack, implementation will include post assembly validation, ADR intentionally leaves room for this

ADR/0070-dual-compression-for-container-builds.md
```diff
@@ -0,0 +1,105 @@
...
+`IMAGE_DIGEST` points to the per-arch index, which contains both variants. Because Tekton Chains does not currently support recursive signing of child manifests within an index ([tektoncd/chains#1070](https://github.com/tektoncd/chains/issues/1070)), the build task explicitly exports the digests for both the gzip and zstd child manifests in its `IMAGES` result. This allows Chains to generate provenance for each manifest directly from the trusted build task, closing potential trust gaps. Downstream tasks, SBOM generation, and signing consume `IMAGE_DIGEST` as the primary artifact. The `io.github.containers.compression.zstd` annotation is set on the zstd manifest descriptor within the index, which is how Podman and CRI-O [identify and prefer](https://github.com/containers/image/blob/main/internal/manifest/oci_index.go) the zstd variant.
+
+In `gzip` or `zstd-chunked` mode, the build task pushes a single manifest as it does today. `IMAGE_DIGEST` holds the primary digest (gzip in `gzip` mode, zstd in `zstd-chunked` mode).
```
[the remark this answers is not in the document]
Amanyd: It is existing tekton result convention used by both tasks today

ADR/0069-dual-compression-for-container-builds.md
```diff
@@ -0,0 +1,105 @@
...
+
+In `gzip` or `zstd-chunked` mode, the build task pushes a single manifest as it does today. `IMAGE_DIGEST` holds the primary digest (gzip in `gzip` mode, zstd in `zstd-chunked` mode).
+
+The `build-image-index` task requires no new parameters. Each entry in the existing `IMAGES` array is now a per-arch index (in `dual` mode) instead of a single manifest. The task's existing `--all` flag on `buildah manifest add` is expected to extract child manifests from per-arch indexes and flatten them into the final multi-arch index, preserving the gzip-first invariant or explicit flattening logic will be added in the task if needed.
+
+SBOM generation requires an update; the SBOM must attach to both the gzip and zstd child manifest digests. The per-arch index is an intermediate build artifact that gets flattened by `build-image-index`, so it is not a valid long-term attachment point. Both compression variants receive the same SBOM since they share identical content, only differing in layer compression.
```
chmeliik: > Both compression variants receive the same SBOM since they share identical content

There's one difference: the main component. The SBOMs reference the container image as the main component, and the component includes the image digest. What digest would we use for the main component in each SBOM?

I would say we have two options:
1. All SBOMs use the per-arch index as the main component
2. Each SBOM uses the manifest it is attached to as the main component

I don't know enough about how these SBOMs are consumed to say what is the correct approach. Option 2) feels more complicated but probably more correct.

chmeliik: > The per-arch index is an intermediate build artifact that gets flattened by `build-image-index`, so it is not a valid long-term attachment point.

This is not always true; for single-arch builds, the per-arch index is the main output. And even for multi-arch builds, it's not like the flattening operation will delete the per-arch index from the registry. The per-arch index will still be accessible by the unique tag. I think we should attach an SBOM to the per-arch index as well.

As for what the contents of that SBOM should be, I'm not sure. Probably similar as the SBOM for a single per-arch manifest? But we could also go the same way as the multi-arch index SBOMs, where the SBOM just lists the individual manifests as variants of the index.

For reference:
- multi-arch index SBOM: `cosign download sbom quay.io/konflux-ci/task-runner:1.8.1`
  - the per-arch index SBOM would look similar, but listing two entries of the same arch
- single manifest SBOM: `cosign download sbom --platform=linux/amd64 quay.io/konflux-ci/task-runner:1.8.1`

ADR/0069-dual-compression-for-container-builds.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Rollout
+
+The default value of `COMPRESSION_FORMAT` stays as `gzip` until the release task fixes are merged and tested. Only then will the default change to `dual`. Conforma also needs follow-up work before the default can flip: although Chains successfully signs all three images (index, gzip child, zstd child), it generates provenance with the gzip child as the attestation subject (from `IMAGE_REF`). Conforma validates each image separately and reports `slsa_build_scripted_build.subject_build_task_matches` violations on the index and zstd child because their digests do not match the attestation subject. Once dual-compression adoption is stable and older Docker versions are no longer a concern, the default can move to `zstd-chunked` only, dropping the gzip variant entirely. There may be additional considerations to take into account when deciding to switch the default to `dual` or `zstd-chunked`. For example, if major installations are still requiring the Docker format instead of OCI for images, then changing this default might need to wait until those installations are ready to move.
```
chmeliik: > it generates provenance with the gzip child as the attestation subject (from `IMAGE_REF`)

This is surprising to me, I thought each image referenced in the `IMAGES` result would gets its own attestation or that they would all become the subjects.

In any case, if Conforma and/or Chains don't support this out of the box, we will not be able to release `dual` builds. That is unfortunate, but if there's no better way to return task results to make things work, then we will probably need to deal with that outside the scope of this ADR

chmeliik (commented):
There's some unknowns remaining around SBOMs, I'd like if someone from @konflux-ci/the-collective could take a look as well.

It's unfortunate that we won't be able to get Conforma verification to work for this initially, but aside from that, the ADR looks reasonable to me.

ADR/0069-dual-compression-for-container-builds.md
BorekZnovustvoritel: TLDR; You are right. We want to use whatever digest the image is actually uniquely pullable by.

Option 2 also makes sense to me.

However we would also need to update the SBOM that is attached to the manifest (we call that the "index-image SBOM"). It would need to distinguish between the compressions so that the SBOM itself makes sense and doesn't just blindly list each architecture-specific child image twice with a different digest each time.

Then we may also want to attach the index-image SBOM to each single-arch image with multiple compressions to point to the correct digest? Or simply attach the SBOM to the per-arch index there? The flattening only for multiarch images makes this quite hard to do in a simple manner. 😅

All of the mapping data would have to be supplied by the build pipeline, Mobster does not have an easy way of gathering the data afterwards.

CC @Allda

chmeliik: Let's use an example to make sure we're using the same terminology:

```
 __________________________________________________
|                   amd64 index                    |
|  _____________________    _____________________  |
| | amd64-gzip manifest |  | amd64-zstd manifest | |
|  ---------------------    ---------------------  |
 --------------------------------------------------

 __________________________________________________
|                   s390x index                    |
|  _____________________    _____________________  |
| | s390x-gzip manifest |  | s390x-zstd manifest | |
|  ---------------------    ---------------------  |
 --------------------------------------------------

 __________________________________________________
|                   final index                    |
|  _____________________    _____________________  |
| | amd64-gzip manifest |  | amd64-zstd manifest | |
|  ---------------------    ---------------------  |
|  _____________________    _____________________  |
| | s390x-gzip manifest |  | s390x-zstd manifest | |
|  ---------------------    ---------------------  |
 --------------------------------------------------
```

I'll document the decisions as I understand them:
- The SBOMs we attach to the `{platform}-{compression} manifest` should be the same as the the SBOMs we use for single-arch images today. In each of these SBOMs, the main component should use the digest of the corresponding `{platform}-{compression} manifest` image.
  - this will still be done by the `buildah` task
- The SBOM we attach to the `final index` should also be the same as what we do today, except it will list 2N entries for N architectures. And as @BorekZnovustvoritel says, we should probably add something into this SBOM to distiniguish between gzip and zstd manifests.
  - this will still be done by the `build-image-index` task
- The SBOM for `{platform} index` is the biggest question mark for me (https://github.com/konflux-ci/architecture/pull/357#discussion_r3450920787). We could make it look like the `{platform}-{compression} manifest` SBOM, but use the `{platform} index` digest for the main component. Or we could make it look like the `final index` SBOM, listing the two manifests with different compressions in this SBOM.
  - either way, this SBOM can be created in the `buildah` task

BorekZnovustvoritel: Yes, you described precisely what I wanted to convey, thank you for clarification.

Amanyd: Since both variants share the same package list I thought one SBOM would work, but yes, the digest in the main component makes them different files. I checked the SBOM from a test build and it's actually worse, the SBOM currently attached to the gzip child has the per-arch index digest as its main component, not even the gzip child's own digest. So it doesn't match the image it's on today. Option 2 makes sense to me as well, each SBOM should use the digest of the manifest it's attached to. Mobster would need to run separately for each variant with the correct digest. And the point about distinguishing gzip from zstd in the index-image SBOM makes sense, I'll account for that in the update too.

Amanyd: Understood, For the SBOM content, making it look like the final multi-arch index SBOM but listing the two children of the same arch seems good, one package for the per-arch index itself (with the index digest as the main component), one package for the gzip child, and one package for the zstd child, each with their own digest. I checked task-runner:1.8.1 index SBOM for reference and it maps over directly. The per-arch index SBOM would be the same structure but with two children of the same arch instead of one child per arch.

The constraint is Mobster needs to know which child is gzip and which is zstd, and that mapping has to come from the build pipeline since Mobster can't gather it on its own. I'll update the ADR to attach SBOM to the per-arch index as well.

ADR/0069-dual-compression-for-container-builds.md
Amanyd: I need to correct what the ADR says here. The Conforma errors kept mentioning the gzip child digest so I assumed Chains was using only the gzip child as the attestation subject, and that's what I wrote in the ADR. But I checked the attestations from the test build again, and that's not what's happening. Chains takes all three images from the IMAGES result and makes them all subjects in one provenance attestation, the per-arch index, the gzip child, and the zstd child all show up as subjects. The same attestation is attached to all three images.
Issue is that Conforma's subject_build_task_matches rule checks each subject against IMAGE_URL@IMAGE_DIGEST from the build task, and since our IMAGE_DIGEST is the per-arch index, only the index subject matches. The gzip and zstd child subjects don't match IMAGE_DIGEST and that's what triggers the violations, so it's a Conforma policy limitation, the policy doesn't accept child manifest subjects when they don't match IMAGE_DIGEST. I'll correct the ADR.

ADR/0069-dual-compression-for-container-builds.md
BorekZnovustvoritel: If needed, Mobster can be reworked a little so that it builds one SBOM and makes it into multiple files with different digests (to save resources).

Amanyd: > If needed, Mobster can be reworked a little so that it builds one SBOM and makes it into multiple files with different digests (to save resources).

Yes, that helps for variant SBOMs but for per arch index SBOM would still need to be a seperate generation step though.

ADR/0069-dual-compression-for-container-builds.md
```diff
@@ -0,0 +1,111 @@
+---
+title: "69. Support for zstd:chunked image compression"
```
cit1zen: The ADR number will have to be changes as we merged ADR no. 69.

ADR/AGENTS.md
```diff
@@ -58,6 +58,7 @@ Status: Ac=Accepted Im=Implemented Ib=Implementable Pr=Proposed Ap=Approved Rp=R
 0031 Rp 43L Sprayproxy
 0068 Pr 145L Standardize on konflux-ci.dev as Official API Group
 0054 Ac 459L Start versioning Tekton Tasks responsibly
+0069 Pr 105L Support for zstd:chunked image compression
```
cit1zen: The ADR number has to be changed here also.

cit1zen (changes requested):

eskultety:
All technical comments have been responded to. I know we know we're facing technical hurdles here, but I'm wondering whether we can accept the ADR and tackle the hurdles one by one in PRs. We're a few weeks in GSoC already and mid-term evaluations are coming, so I think we would want to get hands dirty with some actual code changes.
cc @chmeliik @ralphbean @arewm

ADR/0069-dual-compression-for-container-builds.md
eskultety: Yep, easy fix :) .

Amanyd: Updated

ADR/AGENTS.md
Amanyd: Updated

chmeliik (approved):
LGTM from my side, good job!

eskultety (approved):

cit1zen (approved):

ADR/0070-dual-compression-for-container-builds.md
```diff
@@ -0,0 +1,111 @@
...
+
+The `build-image-index` task requires no new parameters. Each entry in the existing `IMAGES` array is now a per-arch index (in `dual` mode) instead of a single manifest. The task's existing `--all` flag on `buildah manifest add` extracts child manifests from per-arch indexes and flattens them into the final multi-arch index. The `--all` preserves both the gzip-first ordering and the `io.github.containers.compression.zstd` annotation on the zstd child manifest descriptors. If [ADR 0069](https://github.com/konflux-ci/architecture/pull/359) proposes a platform sort in `build-image-index` for reproducibility, it must either include compression type in the sort key or use a stable sort to preserve the gzip-first invariant.
```
ralphbean: **[nit]** ADR 0069 is merged now, so this could be declarative. The link could also point to the file instead of the closed PR.

```suggestion
The `build-image-index` task requires no new parameters. Each entry in the existing `IMAGES` array is now a per-arch index (in `dual` mode) instead of a single manifest. The task's existing `--all` flag on `buildah manifest add` extracts child manifests from per-arch indexes and flattens them into the final multi-arch index. The `--all` preserves both the gzip-first ordering and the `io.github.containers.compression.zstd` annotation on the zstd child manifest descriptors. [ADR 0069](0069-reproducible-container-builds.md) proposes a platform sort in `build-image-index` for reproducibility. That sort must either include compression type in the sort key or use a stable sort to preserve the gzip-first invariant.
```

ralphbean (approved):

ralphbean:
Oh, small thing. Needs to run `hack/util/generate-agents-md`.

ADR/AGENTS.md
```diff
@@ -59,6 +59,7 @@ Status: Ac=Accepted Im=Implemented Ib=Implementable Pr=Proposed Ap=Approved Rp=R
 0031 Rp 43L Sprayproxy
 0068 Pr 145L Standardize on konflux-ci.dev as Official API Group
 0054 Ac 459L Start versioning Tekton Tasks responsibly
+0070 Pr 112L Support for zstd:chunked image compression
```
ralphbean:
```suggestion
0070 Pr 111L Support for zstd:chunked image compression
```

arewm (approved):

Issue #299, opened 2025-12-08
Thoughts on Konflux and reproducible builds


jlebon:
In Hummingbird, we've recently completed some work to enable reproducible builds for all our images. You can see more information on this in the upstream README: https://gitlab.com/redhat/hummingbird/containers#reproducible-builds.

I know there have been discussions in the past around supporting reproducible builds in Konflux more globally. And the Hummingbird work has mostly been happening outside of this, with e.g. dedicated tooling that was developed (though some basic enablement work did land in the buildah task).

Obviously, a lot of that work is actually _not_ specific to Hummingbird, and Konflux is well-positioned to provide more generic tooling for this that would work for any Konflux-built image which claims to be reproducible.

I don't have a full story here to provide though in the process of working on this, I did gather various notes and ideas on what some elements of that story could/should be. And this is a semi-structured dump of those notes. Some of those points sound authoritative because it flows better, but obviously these are all suggestions and up for debate.

## Design Requirements

1. The Containerfile steps already need to be reproducible before it can take advantage of the Konflux reproducible builds story, e.g. it should have a `SOURCE_DATE_EPOCH` build arg and use buildah v1.41's reproducible builds enablement; Konflux _shouldn't_ start doing invasive things like injecting `ARG` directives or cleaning up non-reproducible caches "behind users' backs". This needs buy-in and ongoing maintenance from the team that owns the image.

2. Users need to be able to reproduce images built by Konflux without having to use Konflux. Additionally, rebuild steps should be as simple as possible and only require common tools like `cosign`, `podman`, and `jq`.

## Simplify the Build Task

The buildah task should as much as possible be about wrapping `buildah` and translating task parameters to buildah options. Mutations should be eliminated or otherwise moved to shared tooling. This allows developer/CI builds that may happen outside of Konflux to be closer to what is built by Konflux.

1. We need to drain the buildah task definition of a lot of the operations which mutate the result, such as adding labels and injecting `labels.json` and `content-sets.json` someplace more easily shared; perhaps the buildah-task container image. That makes it easier to be used outside of Konflux.

[...]

PR #360, opened 2026-05-21
Add ADR-0069: Reproducible Container Builds in Konflux

JamesOlaitan:
## Summary

Proposes decisions to move reproducible container builds in Konflux from an available primitive (the `buildah-oci-ta` task accepts `SOURCE_DATE_EPOCH`, `REWRITE_TIMESTAMP`, and `OMIT_HISTORY` as params from [build-definitions#2947](https://github.com/konflux-ci/build-definitions/pull/2947)) to a supported user feature.

## Decisions

- Expose three new opt-in pipeline parameters on every `docker-build*` variant: `source-date-epoch`, `rewrite timestamp`, `omit-history`. All default-off. `source-date-epoch` auto-sources from `commit-timestamp` only when `rewrite-timestamp` is also enabled (since that's the flag that actually needs a `SOURCE_DATE_EPOCH` value to clamp file mtimes against). Roadmap to default-on described in the ADR.
- Handle `io.buildah.version` at the task level. Buildah's `--identity-label` defaults to off when `--source-date epoch` is set, so users don't need `SKIP_INJECTIONS=true` for cross-task-version reproducibility. The current v0.9 task manually injects the label outside buildah's control. [An upcoming task version](https://github.com/konflux-ci/build-definitions/pull/3533) delegates the build step to a Go binary (`konflux-build-cli`) that lets buildah handle the label correctly. `SKIP_INJECTIONS` purpose is then limited to removing `content-sets.json`.
- Deliver `verify-reproducibility` as a separate, on-demand pipeline. The pipeline accepts an image reference and a source artifact, rebuilds once from the source, and compares the resulting digest against the supplied image reference. The asserted property is framed using SLSA's Verification Summary Attestation (`slsa.buildReproduced`) with a same-platform limitation acknowledged.
- Address the secondary-artifact tag-collision concern raised on the closed [#3263](https://github.com/konflux-ci/build-definitions/pull/3263) (the `/retest` scenario failure mode) by proposing two build-side options: tag secondary-artifact blobs with a per-pipelineRun identifier instead of the image digest, or move to the OCI Referrers API. The Conforma verification-side fix ([conforma/policy#1590](https://github.com/conforma/policy/pull/1590)) is described as necessary but not sufficient. The ADR commits to a follow-up that picks between the two options.
- Propose a sort step inside `build-image-index` as the Konflux-side fix for multi-arch manifest ordering ([buildah#6383](https://github.com/containers/buildah/issues/6383)) as the upstream issue has had no activity since September 2025.

## Empirical backing

A patched `docker-build-oci-ta` pipeline was run against a local `konflux-ci/konflux-ci` cluster. Findings (digests in the ADR):

- Alpine COPY-only: byte-wise reproducible.
- Go multi-stage with `-trimpath`: byte-wise reproducible.
- Fedora `dnf install curl`: not reproducible due to `/var/log/dnf5.log` content. Workaround is Containerfile-side (delete `/var/log/dnf*` in the same `RUN`).
- `SKIP_INJECTIONS=true` vs `false`: each is internally reproducible, and the two flags produce different digests as expected (injection adds a layer).
- Tag collision: empirically reproduced (`.sbom`, `.att`, `.sig`, `.dockerfile` all digest-derived and overwrite across identical-digest runs).

## Open question

Is the syft-generated SBOM byte-identical when run twice against a byte-identical input image? If not, the fix is either the off-digest tag derivation or the OCI Referrers API proposed in the secondary-artifact section.

## Refs

- konflux-ci/architecture#299 (long-term roadmap this aligns with)
- konflux-ci/build-definitions#1268 (tracking issue)
- konflux-ci/build-definitions#2947 (merged: task-level flag support)
- konflux-ci/build-definitions#3263 (closed: pipeline wiring)
- konflux-ci/build-definitions#3383, #3385, #3386 (closed: verify-reproducibility attempts)
- konflux-ci/build-definitions#3533 (open: upcoming task version using konflux-build-cli)
- conforma/policy#1590 (merged: latest-attestation fetch)


ADR/0067-reproducible-container-builds.md
tnevrlka, line 1: Please sign-off your commit and add a `Assisted-by: <AI model>` line **if** AI was used for the creation of this document.

ADR/0067-reproducible-container-builds.md
```diff
@@ -0,0 +1,137 @@
+---
+title: "0067. Reproducible Container Builds in Konflux"
```
tnevrlka: ADR 67 is already taken by https://github.com/konflux-ci/architecture/pull/354, it'll have to be a different number

ADR/0067-reproducible-container-builds.md
```diff
@@ -0,0 +1,137 @@
...
+
+### Sources of non-determinism
+
+We catalog the sources in four families. Each entry includes a severity ("blocks reproducibility" / "user-side") and the mitigation strategy this ADR commits to.
+
+**Timestamp-based.** The image config `created` field, file mtimes inside layer tarballs, build history timestamps, and gzip header timestamps all default to wall-clock time.
+
+*Severity:* blocks reproducibility.
+
+*Mitigation:* `--source-date-epoch` sets the config `created` field and clamps mtimes; `--rewrite-timestamp` clamps them further across the entire image; `--omit-history` removes build history timestamps. Buildah 1.43 makes `--timestamp` and `--source-date-epoch` mutually exclusive ("timestamp and source-date-epoch would be ambiguous if allowed together"), and the buildah-oci-ta task enforces this at the script level ([task/buildah-oci-ta/0.9/buildah-oci-ta.yaml line 700](https://github.com/konflux-ci/build-definitions/blob/main/task/buildah-oci-ta/0.9/buildah-oci-ta.yaml#L700)).
```
tnevrlka: > `--source-date-epoch` sets the config `created` field and clamps mtimes

I don't think it clamps mtimes by itself, isn't that what `--rewrite-timestamp` does? 
```golang
if options.SourceDateEpoch != nil {
	sourceDateEpochUTC := options.SourceDateEpoch.UTC()
	created = &sourceDateEpochUTC
	if options.RewriteTimestamp {
		layerLatestModTime = &sourceDateEpochUTC
	}
}
```
https://github.com/containers/buildah/blob/b9396adba9271dc375f6d234c85ba2e8d38de167/image.go#L1700-L1706

tnevrlka: >`--rewrite-timestamp` clamps them further across the entire image

"clamps them further" is imo not descriptive enough, I don't know what to imagine from that

From what I understand, it's what introduces the "clamping", see the commit message where it was introduced
https://github.com/containers/buildah/commit/6c82e7eac02930b33a43da1a29f92c3f49f458f7

ADR/0067-reproducible-container-builds.md
```diff
@@ -0,0 +1,137 @@
...
+
+* The verification scope is separated from production build. Pipelines that run today continue to run as a single build. Users who want verification opt into a separate pipeline whose only job is to re-run the build and diff.
+
+* The pipeline parameter grows by three fields (`SOURCE_DATE_EPOCH` auto-wired plus two opt-in flags). This sits within the precedent set by PR #2947 at the task level, which added the same three params there. Cleanup of the param surface, if any, will be in a future ADR.
+
+* `SKIP_INJECTIONS=true` for byte-wise reproducibility places a documentation burden on the user: their Containerfile must bake in any labels the pipeline would have injected. Third-party security scanners that depend on the current injection locations continue to work because the default does not change.
+
+* `dnf5.log` content non-determinism is a documented Containerfile authoring gotcha for the RPM scenario. The workaround is to delete `/var/log/dnf*` in the same `RUN` as the install; the user guide will call this out.
```
eskultety: What about the default `rpmdb` configuration that leads to creating sqlite write-ahead logs, i.e. `/var/lib/rpm/rpmdb.sqlite-wal` files, is `SOURCE_DATE_EPOCH` respected natively or do we need to start enforcing a particular configuration for rpmdb?

ADR/0067-reproducible-container-builds.md
```diff
@@ -0,0 +1,137 @@
...
+
+* Multi-architecture manifest ordering remains non-deterministic until the upstream buildah fix. The ADR does not commit to a task-level workaround because the right place to fix it is upstream.
+
+* Secondary-artifact tag collisions are partially addressed by the merged conforma/policy fix. The build-side inventory this ADR commits to may expose additional guardrails. If it does, this ADR will be amended; if it does not, the verification-side fix is sufficient.
+
+* The decisions in this ADR are forward-compatible with the two-stage prefetch-then-hermetic verification model described in the architecture roadmap in issue #299: our verification pipeline can later be extended to a fully hermetic rebuild without re-litigating any of the choices made here.
```
eskultety: The ADR is proposing making this opt-in (good). I believe the ultimate state is "enforced" though, am I correct? Would it be possible to outline steps towards that goal eventually as part of this ADR, i.e. what needs to happen to become the default?

ADR/0067-reproducible-container-builds.md
```diff
@@ -0,0 +1,137 @@
...
+
+We'll document this as a user-side requirement and update the buildah-oci-ta README accordingly. We will not remove the injection path or change its default. Third-party security scanners that read labels from the current injection locations ([raised by MartinBasti on PR #3263](https://github.com/konflux-ci/build-definitions/pull/3263)) must keep working; making `SKIP_INJECTIONS` opt-in keeps their
```
eskultety: general advice: This is a technical document, references is all we need, no need to mention any names directly, that is in fact an undesirable pattern to have in technical docs.

ADR/0067-reproducible-container-builds.md
```diff
@@ -0,0 +1,137 @@
...
+
+*Mitigation (`dnf5.log`):* a Containerfile-side fix. Deleting it in the same `RUN` as the install (`RUN dnf install ... && dnf clean all && rm -rf /var/log/dnf*`) removes the divergent bytes from the final layer. We are not proposing a task-level strip because making `buildah-oci-ta` silently delete files post-build would add exactly the kind of out-of-band Containerfile mutation that architecture issue #299 wants to eliminate. Logs are event records and are unlikely to ever be byte-deterministic upstream, hence the workaround.
+
+**External.** Unpinned package versions (`dnf install curl` resolves to different versions over time), floating base image tags (`FROM fedora:latest`), and network-fetched build inputs.
+
+*Severity:* user-side.
+
+*Mitigation:* documentation. The hermetic build mode (`HERMETIC=true` with `PREFETCH_INPUT`) addresses the network-fetching part; the rest are Containerfile-authoring practice.
+
+### Findings from experiments
```
tnevrlka: Kudos for experiments

ADR/0067-reproducible-container-builds.md
```diff
@@ -0,0 +1,137 @@
...
+
+An initial four-cell flag experiment was run across three Containerfile scenarios (Alpine COPY-only, Fedora `dnf install`, Go multi-stage), using plain `buildah build` from podman. Three findings carry into this ADR.
+
+First, `--timestamp` and `--source-date-epoch` are mutually exclusive in buildah 1.43+. The pipeline should wire only `SOURCE_DATE_EPOCH`, not both. The task already enforces this, so no pipeline-level guard is needed.
+
+Second, for the Alpine COPY-only case, `--source-date-epoch` plus `--rewrite-timestamp` produces byte-identical images at both the uncompressed (`diff_ids`) and compressed (blob digest) level. The
+trivial scenario is solvable today.
+
+Third, for the Fedora RPM case, the strongest flag combination still produces divergent images because `/var/log/dnf5.log` contains non-deterministic transaction text. Mtimes are clamped; the remaining
+source is content, not time. This is a known limitation we surface in Consequences.
```
tnevrlka: Perhaps a link to the Consequences section would make it a bit easier?

tnevrlka (commented):
I have more comments on my mind, but need more time to research and review. Submitting what I have for now

ADR/0067-reproducible-container-builds.md
```diff
@@ -0,0 +1,137 @@
...
+
+*Severity:* blocks reproducibility.
+
+*Mitigation:* `--source-date-epoch` sets the config `created` field and clamps mtimes; `--rewrite-timestamp` clamps them further across the entire image; `--omit-history` removes build history timestamps. Buildah 1.43 makes `--timestamp` and `--source-date-epoch` mutually exclusive ("timestamp and source-date-epoch would be ambiguous if allowed together"), and the buildah-oci-ta task enforces this at the script level ([task/buildah-oci-ta/0.9/buildah-oci-ta.yaml line 700](https://github.com/konflux-ci/build-definitions/blob/main/task/buildah-oci-ta/0.9/buildah-oci-ta.yaml#L700)).
+
+**Ordering-based.** Filesystem iteration order inside tar archives depends on the underlying filesystem and creation time. RPM installation order is not deterministic by default. Multi-architecture manifest ordering is subject to Go map iteration randomization (the underlying issue is [buildah#6383](https://github.com/containers/buildah/issues/6383), where `platformsForBaseImages` returns platforms in unpredictable order).
+
+*Severity:* blocks reproducibility for multi-arch.
+
+*Mitigation:* this ADR proposes an upstream-only fix path for the buildah issue.
```
tnevrlka: I'm not sure I get this, where is it proposed?

ADR/0067-reproducible-container-builds.md
```diff
@@ -0,0 +1,137 @@
...
+
+**Content-based.** The Konflux pipeline injects `labels.json` and `content-sets.json` into builds through appended COPY instructions in the buildah-oci-ta task. One of the labels written into `labels.json` is `io.buildah.version=$(buildah version)` ([line 844 in the task](https://github.com/konflux-ci/build-definitions/blob/main/task/buildah-oci-ta/0.9/buildah-oci-ta.yaml#L844)), which means identical source against a different task image version produces a different image. The Fedora/RHEL rpm database defaults to SQLite WAL mode, which is non-deterministic; Project Hummingbird (Red Hat, March 2026) demonstrated that switching to DELETE journal mode makes the rpmdb byte-stable. Preliminary experiment also found `/var/log/dnf5.log` contains non-deterministic transaction text (24 bytes of divergence between two otherwise identical builds, mtimes already clamped). `--rewrite-timestamp` cannot fix this because the divergence is content, not timestamp.
```
tnevrlka: > ([line 844 in the task](https://github.com/konflux-ci/build-definitions/blob/main/task/buildah-oci-ta/0.9/buildah-oci-ta.yaml#L844))

Note that this is not a permalink and the line number could therefore change if there was any line before it added or removed

ADR/0067-reproducible-container-builds.md
```diff
@@ -0,0 +1,137 @@
...
+per-run identifiers. The provenance-fetch path is covered by `conforma/policy#1590`. The remaining concern is the SBOM blob path: if syft is non-deterministic, attestations from the first run point at a SBOM digest that no longer matches the overwritten tag. We propose either making the SBOM blob tag derive from the TaskRun UID or ensuring syft determinism so the blob digest itself is stable across runs.
+
+### Upstream dependencies
+
+Multi-architecture manifest ordering is governed by Go map iteration randomization inside buildah's `platformsForBaseImages` ([buildah#6383](https://github.com/containers/buildah/issues/6383)). A fix there is upstream, not task-level. So, we'll be monitoring the issue and adding a sort step to the buildah-oci-ta task only if the upstream fix stalls.
+
+### Open Questions
+
+1. Is the Hermeto/cachi2 RUN-line Containerfile mutation byte-stable across runs given identical `PREFETCH_INPUT`? The first attempt failed because the test Containerfile lacked an `rpms-lock.yaml`, so cachi2 produced no prefetch artifact and the downstream `use-trusted-artifact` step had nothing to extract. To be re-run with a properly declared hermetic scenario.
```
tnevrlka: I would consider dropping the references to cachi2, it has been renamed to Hermeto for quite a while now

tnevrlka: And yes, the injection should be identical. cc @chmeliik

tnevrlka: See https://github.com/konflux-ci/konflux-build-cli/blob/f7770274910b5435deee5da684b2b202a6af2e40/pkg/commands/build.go#L1075 and the definitions of functions it uses

ADR/0067-reproducible-container-builds.md
```diff
@@ -0,0 +1,137 @@
...
+per-run identifiers. The provenance-fetch path is covered by `conforma/policy#1590`. The remaining concern is the SBOM blob path: if syft is non-deterministic, attestations from the first run point at a SBOM digest that no longer matches the overwritten tag. We propose either making the SBOM blob tag derive from the TaskRun UID or ensuring syft determinism so the blob digest itself is stable across runs.
+
+### Upstream dependencies
+
+Multi-architecture manifest ordering is governed by Go map iteration randomization inside buildah's `platformsForBaseImages` ([buildah#6383](https://github.com/containers/buildah/issues/6383)). A fix there is upstream, not task-level. So, we'll be monitoring the issue and adding a sort step to the buildah-oci-ta task only if the upstream fix stalls.
```
tnevrlka: > only if the upstream fix stalls

No update since September 2025 is in my opinion pretty stalled

ADR/0067-reproducible-container-builds.md
```diff
@@ -0,0 +1,137 @@
...
+
+We'll document this as a user-side requirement and update the buildah-oci-ta README accordingly. We will not remove the injection path or change its default. Third-party security scanners that read labels from the current injection locations ([raised by MartinBasti on PR #3263](https://github.com/konflux-ci/build-definitions/pull/3263)) must keep working; making `SKIP_INJECTIONS` opt-in keeps their
+integration stable.
+
+### Verification as a separate pipeline, not a build step
+
+We will not add a build-twice-and-compare step inside the default `docker-build*` pipeline (a verification step that builds twice does not belong there, as the review of PRs #3383, #3385, and #3386 established).
```
tnevrlka: I wouldn't consider reviews to be some truth set in stone that needs to be followed. 

I think it could be better to instead take the ideas from the reviews and try to reason and explain in the ADR why you think the build-twice-and-compare step should not be added

ADR/0067-reproducible-container-builds.md
```diff
@@ -0,0 +1,137 @@
...
+
+Instead, we'll deliver a standalone `verify-reproducibility` pipeline that accepts an image reference and a source artifact, runs the build twice from the same source with identical reproducibility parameters, compares the resulting digests, and emits `REPRODUCIBLE=true|false` plus a structured `DIFF_SUMMARY` result. On mismatch, it invokes [diffoci](https://github.com/reproducible-containers/diffoci) for a layer-by-layer diff. Users or CI trigger this pipeline on demand; it is independent of the production build flow. This also extends naturally toward the two-stage prefetch-then-hermetic model in architecture issue #299.
```
tnevrlka: > two-stage prefetch-then-hermetic model in architecture issue #299 

I'm not sure I see this model in the issue, would it be possible to link a more precise place where it's mentioned?

ADR/0067-reproducible-container-builds.md
```diff
@@ -0,0 +1,137 @@
...
+
+### Secondary artifacts and tag-collision awareness
+
+When primary image digests become reproducible, secondary artifacts (SBOM, SLSA provenance, SARIF results, source image) face a tag collision: the same digest causes secondary-artifact tags to overwrite each other in the registry. Ralph Bean's review on PR #3263 describes the failure mode in detail: secondary artifacts from the first run lose their registry tag to the second run, Quay's garbage collector then deletes them, and provenance attestations from the first run point at now-missing artifacts.
+
+[`conforma/policy` PR #1590](https://github.com/conforma/policy/pull/1590) (merged 2025-12-04) addresses the verification side by returning the latest pipelineRun attestation.
```
tnevrlka: I'm not sure whether the date when a PR was merged is needed

ADR/0067-reproducible-container-builds.md
tnevrlka, line 1: There are references to PRs in multiple places in the format of `#<number>`.
Since the referenced PRs are in other repositories than konflux-ci/architecture, I think it would make sense to link them every time anyway. 

I believe it could be done somewhat painlessly with substitution/find and replace...or LLMs... 😄

ADR/0067-reproducible-container-builds.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Decision
+
+We'll wire the existing buildah reproducibility primitives through Konflux build pipelines as opt-in parameters, expose `SKIP_INJECTIONS` as a prerequisite for byte-wise reproducibility, treat reproducibility verification as a separate pipeline, and defer the multi-arch and SBOM content questions to follow-up work described below.
+
+### Pipeline-level wiring of reproducibility parameters
+
+We'll wire `clone-repository.results.commit-timestamp` into the `buildah-oci-ta` task's `SOURCE_DATE_EPOCH` parameter in every `docker-build*` pipeline variant. We'll expose `REWRITE_TIMESTAMP` and `OMIT_HISTORY` as pipeline-level parameters whose default is `false`. We won't be introducing a new pipeline-level `SOURCE_DATE_EPOCH` parameter that competes with the auto-sourced one; reproducible builds should track the source commit, not an arbitrary epoch.
+
+Changes set in `konflux-ci/build-definitions`: `docker-build`, `docker-build-oci-ta`, `docker-build-multi-platform-oci-ta`, and `docker-build-oci-ta-min`. After modifying pipeline definitions, run `./hack/generate-everything.sh` so Trusted Artifact variants stay in sync.
```
tnevrlka: I'm not sure this line is needed, seems like implementation details to me

ADR/0067-reproducible-container-builds.md
```diff
@@ -0,0 +1,137 @@
...
+
+### Containerfile-side prerequisites for byte-wise reproducibility
+
+`SKIP_INJECTIONS=true` is a prerequisite for byte-wise reproducibility. With injection enabled, the task appends `COPY labels.json` and `COPY content-sets.json` instructions, and `labels.json` includes `io.buildah.version`. When the task image is upgraded, that label changes and identical source produces a different image. Users who want byte-wise reproducibility across task image rebuilds must bake their labels into the Containerfile and set `SKIP_INJECTIONS=true`. Within a single task image, both settings are equally reproducible: PipelineRuns of the patched pipeline against an Alpine COPY-only Containerfile produced identical digests at `SKIP_INJECTIONS=false` (`sha256:2296efeca6...`) and at `SKIP_INJECTIONS=true` (`sha256:fc04d038...`). The two settings differ because injection adds one layer (the appended COPY instructions). So `SKIP_INJECTIONS` is a content choice the user picks once and stays with.
```
tnevrlka: I'm asking because I actually don't know, so I'd love if somebody chimed in.

Is there a use case for having reproducibility across different versions of buildah? 

I guess having the option doesn't hurt, but I'm wondering if it's something that would be widely used.

chmeliik: It would be nice not to break the reproducibility when the user updates their buildah Task version, even if the new Task version comes with a new buildah version.

This won't always be possible, because different versions of buildah may inject different metadata or even construct the rootfs slightly differently. But eliminitating the `io.buildah.version` difference does make sense.

Note that `SKIP_INJECTIONS=true` isn't needed though: https://www.mankier.com/1/buildah-build#--identity-label
> Adds a label io.buildah.version with its value set to the version of buildah which built the image (default true unless [--timestamp](https://www.mankier.com/1/buildah-build#--timestamp) or [--source-date-epoch](https://www.mankier.com/1/buildah-build#--source-date-epoch) is used).

If you're using `--source-date-epoch`, which is a prerequisite for reproducibility, then buildah does not inject the label. Version 0.9 of the buildah Task doesn't handle this correctly and injects the version always. [Version 0.10](https://github.com/konflux-ci/build-definitions/pull/3533) will fix this (missed this change in the changelog, will update)

chmeliik: Also note that konflux-build-cli has a Reproducibility test with label injection enabled: https://github.com/konflux-ci/konflux-build-cli/blob/84da72285b74231d32d41f7c307a2f5325a4a54d/integration_tests/build_test.go#L1566
(which wouldn't catch cross-version problems, but does cover the same-version case)

ADR/0067-reproducible-container-builds.md
```diff
@@ -0,0 +1,137 @@
...
+
+Date: 2026-05-20
+
+## Status
+
+Proposed
+
+## Context
+
+Konflux container builds currently produce different image digests from identical source commits. That blocks third parties from independently verifying that a deployed artifact matches the signed-off source, the property reproducible builds are supposed to give us.
```
arewm: What do you mean the signed-off source here? Is this just the recorded source in the SLSA provenance?

ADR/0067-reproducible-container-builds.md
```diff
@@ -0,0 +1,137 @@
...
+
+*Severity:* blocks reproducibility.
+
+*Mitigation:* `--source-date-epoch` sets the config `created` field and clamps mtimes; `--rewrite-timestamp` clamps them further across the entire image; `--omit-history` removes build history timestamps. Buildah 1.43 makes `--timestamp` and `--source-date-epoch` mutually exclusive ("timestamp and source-date-epoch would be ambiguous if allowed together"), and the buildah-oci-ta task enforces this at the script level ([task/buildah-oci-ta/0.9/buildah-oci-ta.yaml line 700](https://github.com/konflux-ci/build-definitions/blob/main/task/buildah-oci-ta/0.9/buildah-oci-ta.yaml#L700)).
+
+**Ordering-based.** Filesystem iteration order inside tar archives depends on the underlying filesystem and creation time. RPM installation order is not deterministic by default. Multi-architecture manifest ordering is subject to Go map iteration randomization (the underlying issue is [buildah#6383](https://github.com/containers/buildah/issues/6383), where `platformsForBaseImages` returns platforms in unpredictable order).
```
arewm: You seem to touch on two different types of issues but we only discuss one mitigation further. My assumption is that the other issue (i.e. RPM installation order) is not solved. Should we state this more explicitly? Are there other occurrences of this type of ordering issue?

ADR/0067-reproducible-container-builds.md
arewm: You mention rpmdb being byte-stable. Does this relate to the ordering non-determinism at all?

ADR/0067-reproducible-container-builds.md
```diff
@@ -0,0 +1,137 @@
...
+([PRs #3383, #3385, #3386](https://github.com/konflux-ci/build-definitions/pulls?q=3383+3385+3386))
+also closed without merging, on the feedback that a "build twice and compare" step does not fit a pipeline that builds once.
+
+This ADR proposes decisions to move reproducible builds in Konflux from "available primitive" to "supported user feature." It aligns with the long-term roadmap in [architecture issue #299](https://github.com/konflux-ci/architecture/issues/299), which describes a future in which the buildah task is a thin wrapper around `buildah build` and rebuild verification is a separate prefetch-then-build-hermetically operation. The decisions here are compatible with that direction without depending on it.
+
+### Sources of non-determinism
```
arewm: I wonder if this information would be better if presented in a different way. You can talk generically about the sources of non-determinism. Then present two tables -- those sources of non-determinism within Konflux that we have a solution/mitigation for and those sources which we do NOT have a solution/mitigation for. We can clearly call out which ones can be solved by Konflux and which need to have user specific changes to resolve.

ADR/0067-reproducible-container-builds.md
```diff
@@ -0,0 +1,137 @@
...
+
+**External.** Unpinned package versions (`dnf install curl` resolves to different versions over time), floating base image tags (`FROM fedora:latest`), and network-fetched build inputs.
+
+*Severity:* user-side.
+
+*Mitigation:* documentation. The hermetic build mode (`HERMETIC=true` with `PREFETCH_INPUT`) addresses the network-fetching part; the rest are Containerfile-authoring practice.
+
+### Findings from experiments
+
+An initial four-cell flag experiment was run across three Containerfile scenarios (Alpine COPY-only, Fedora `dnf install`, Go multi-stage), using plain `buildah build` from podman. Three findings carry into this ADR.
```
arewm: Were you using `buildah build` or `podman build`? Buildah and podman are related but their own tools.

ADR/0067-reproducible-container-builds.md
arewm: Yes, thanks. Are these relevant for the ADR specifically? Is the ADR more about how to achieve reproducibility than how we tested to see how things are not reproducible.

This information helps to understand the context and it feeds the sources of non-determinism (should it be moved up if we include it?). But should they be kept separate _somewhere_ and just referenced?

If an AI agent is consuming this ADR to understand reproducibility, for example, this information would likely be largely irrelevant for them to consume.

ADR/0067-reproducible-container-builds.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Decision
+
+We'll wire the existing buildah reproducibility primitives through Konflux build pipelines as opt-in parameters, expose `SKIP_INJECTIONS` as a prerequisite for byte-wise reproducibility, treat reproducibility verification as a separate pipeline, and defer the multi-arch and SBOM content questions to follow-up work described below.
+
+### Pipeline-level wiring of reproducibility parameters
+
+We'll wire `clone-repository.results.commit-timestamp` into the `buildah-oci-ta` task's `SOURCE_DATE_EPOCH` parameter in every `docker-build*` pipeline variant. We'll expose `REWRITE_TIMESTAMP` and `OMIT_HISTORY` as pipeline-level parameters whose default is `false`. We won't be introducing a new pipeline-level `SOURCE_DATE_EPOCH` parameter that competes with the auto-sourced one; reproducible builds should track the source commit, not an arbitrary epoch.
```
arewm: Do we anticipate any adverse consequences of wiring the `SOURCE_DATE_EPOCH` to every build? How would users override this if they don't want it?

ADR/0067-reproducible-container-builds.md
```diff
@@ -0,0 +1,137 @@
...
+
+Changes set in `konflux-ci/build-definitions`: `docker-build`, `docker-build-oci-ta`, `docker-build-multi-platform-oci-ta`, and `docker-build-oci-ta-min`. After modifying pipeline definitions, run `./hack/generate-everything.sh` so Trusted Artifact variants stay in sync.
+
+### Default policy: opt-in, not default-on
+
+`REWRITE_TIMESTAMP` and `OMIT_HISTORY` ship default-false. The `SOURCE_DATE_EPOCH` auto-wiring ships enabled: matching the image's `created` field to the source commit is a strictly better default and does not change layer content. We propose revisiting the `REWRITE_TIMESTAMP` default-on question after a deprecation window once downstream consumers (registry tooling, cache invalidation logic that reads the `created` field) have had time to adapt.
```
arewm: This may violate assumptions. If artifacts are not reproducible (for example, they don't pin dependencies) then adjusting the created date to match the git commit seems like a bad practice. Then the users would not be able to use that as a valid signal for when dependencies are fetched if they want it.

ADR/0067-reproducible-container-builds.md
arewm: Is the issue just with `labels.json`? Are we just arbitrarily lumping `content-sets.json` into this same class?

ADR/0067-reproducible-container-builds.md
arewm: I generally agree with this statement. SLSA previously had reproducible as a property of a build, but I have come to think of it as a property of verification. When we introduced the SLSA source track, we also introduced the concept of verified properties within the Verification Summary Attestation (VSA). I managed to sneak in one related to reproducibility as well: https://slsa.dev/spec/v1.2/verified-properties#slsabuildreproduced (even though I haven't gotten around to formalizing reproducibility in SLSA yet).

That is to say, that reproducibility is something to be verified, so I think a separate build makes sense. What also makes sense in my mind is if you can try to reproduce an artifact by using the provenance as the build recipe. That being said, reproducing on the same build platform only gets you so far. It doesn't help to mitigate attacks against the build platform itself. Ideally, these would be done by some separate identity/entity and on a different build platform.

What aspects of a source are you trying to claim as reproducible though? Are you just saying that something is reproducible with the TA inputs to a buildah task? Or that artifacts are reproducible from the source commit itself?

ADR/0067-reproducible-container-builds.md
```diff
@@ -0,0 +1,137 @@
...
+
+Instead, we'll deliver a standalone `verify-reproducibility` pipeline that accepts an image reference and a source artifact, runs the build twice from the same source with identical reproducibility parameters, compares the resulting digests, and emits `REPRODUCIBLE=true|false` plus a structured `DIFF_SUMMARY` result. On mismatch, it invokes [diffoci](https://github.com/reproducible-containers/diffoci) for a layer-by-layer diff. Users or CI trigger this pipeline on demand; it is independent of the production build flow. This also extends naturally toward the two-stage prefetch-then-hermetic model in architecture issue #299.
+
+### Secondary artifacts and tag-collision awareness
+
+When primary image digests become reproducible, secondary artifacts (SBOM, SLSA provenance, SARIF results, source image) face a tag collision: the same digest causes secondary-artifact tags to overwrite each other in the registry. Ralph Bean's review on PR #3263 describes the failure mode in detail: secondary artifacts from the first run lose their registry tag to the second run, Quay's garbage collector then deletes them, and provenance attestations from the first run point at now-missing artifacts.
```
arewm: Can you be more specific about this issue? Is it just for artifacts that are using the tag-based backwards compatible referrer's API? Is it just for those using the actual referrer's API? Is it just for items like the trusted artifacts? 

There isn't enough context here for me to understand what we are trying to describe. You go into this a little more later, but can we instead start with the motivation for the concern instead of how this relates to reproducibility?

I cannot, for example, determine if this is an issue with the tag-based referrer's API data being deleted when a digest is re-pushed (the digest in the tag should be the same, so the reference should still be valid) or if there is just duplication. Also is there an issue with the other metadata which is pushed that might not be reproducible? Referenced digests (i.e. in the provenance) are not guaranteed to be permanently valid.

Is there an issue here that derives from how cosign and other tools differ in their functinality? When you push a new image, the attestations are preserved for older runs. WHen you push new SBOMs, etc, however, the tooling may not be as forgiving unless they also push unique tag identifiers to ensure that the digests remain valid. Would any of this be solved by moving to the actual referrer's API instead of the tag-based convention?

ADR/0067-reproducible-container-builds.md
```diff
@@ -0,0 +1,137 @@
...
+
+### Open Questions
+
+1. Is the Hermeto/cachi2 RUN-line Containerfile mutation byte-stable across runs given identical `PREFETCH_INPUT`? The first attempt failed because the test Containerfile lacked an `rpms-lock.yaml`, so cachi2 produced no prefetch artifact and the downstream `use-trusted-artifact` step had nothing to extract. To be re-run with a properly declared hermetic scenario.
+2. Is the syft-generated SBOM byte-identical when run twice against a byte-identical input image? Direct comparison was inconclusive in local experiments because the SBOM blob is stored under a tag derived from the primary image digest (`sha256-<digest>.sbom`), so identical primary digests caused the second run's SBOM tag to overwrite the first's before we could pull both. Resolving this needs either capturing the SBOM blob inside each TaskRun before push, or moving the tag derivation off the primary digest.
```
arewm: As I alluded to above, I suspect that this is a problem as well. We cannot just have metadata disappearing from our prior runs. This may be why Conforma needed to always get the latest provenance attestation ... because the digest-pinned data in any older ones is no longer relevant due to issues like this.

ADR/0067-reproducible-container-builds.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Consequences
+
+* Users gain an opt-in path to reproducible container image digests without changes to existing CI consumers. The default for new pipelines tracks the commit timestamp; layer content is unchanged unless the user opts in to `REWRITE_TIMESTAMP`.
```
arewm: As I mentioned before, all functionality isn't opt-in.

ADR/0067-reproducible-container-builds.md
```diff
@@ -0,0 +1,137 @@
...
+
+* The non-determinism catalog in Context becomes the reference for evaluating future changes to the buildah-oci-ta task. New code added to the task gets compared against the four families to confirm it does not introduce a new non-determinism source.
```
arewm: Are you proposing to create a tool that does this analysis or does is consequence logically derived from earlier content?

ADR/0067-reproducible-container-builds.md
```diff
@@ -0,0 +1,137 @@
...
+
+* The pipeline parameter grows by three fields (`SOURCE_DATE_EPOCH` auto-wired plus two opt-in flags). This sits within the precedent set by PR #2947 at the task level, which added the same three params there. Cleanup of the param surface, if any, will be in a future ADR.
+
+* `SKIP_INJECTIONS=true` for byte-wise reproducibility places a documentation burden on the user: their Containerfile must bake in any labels the pipeline would have injected. Third-party security scanners that depend on the current injection locations continue to work because the default does not change.
+
+* `dnf5.log` content non-determinism is a documented Containerfile authoring gotcha for the RPM scenario. The workaround is to delete `/var/log/dnf*` in the same `RUN` as the install; the user guide will call this out.
+
+* Multi-architecture manifest ordering remains non-deterministic until the upstream buildah fix. The ADR does not commit to a task-level workaround because the right place to fix it is upstream.
```
arewm: Is this something that can be resolved when we build the image index or is it completely out of our control?

arewm (commented):
@JamesOlaitan , thanks for this write up! I know that it is really hard to discuss something as nuanced as reproducible build. I left some comments in line.

@tnevrlka also commented a couple of times about links to other repositories. 

I think that one of the biggest thing that you should work on is to digest the various PRs and issues that you referenced and encapsulate that into your own representation of the devision. PRs referenced throughout this document may be a source for you to pull information from, but referring out to various PRs is hard to follow and sign off on as "authoritative." Take the various inputs and produce the single authoritative architectural decision record in this document that is fully encapsulated. It is fine if you agree or disagree with some of the other comments that have been made.

ADR/0067-reproducible-container-builds.md
JamesOlaitan: Yes. I've clarified it. --rewrite-timestamp uses the --source-date-epuch to clamp mtimes

ADR/0067-reproducible-container-builds.md
JamesOlaitan: I made it into a table so how to mitigate each reproducibility blocker is clearer rather than jumping across the ADR

ADR/0067-reproducible-container-builds.md
JamesOlaitan: Both issues are stated explicitly now in a table.

ADR/0067-reproducible-container-builds.md
JamesOlaitan: I've made the clarification. For the one that require an upstream fix but can also be patched at the task layer on Konflux, I also stated it and the patch we can apply

ADR/0067-reproducible-container-builds.md
JamesOlaitan: I removed the line number link and I pinned to the current head so following the link will lead to the same content.

ADR/0067-reproducible-container-builds.md
JamesOlaitan: I added the sort step as a proposal to be included in the task level. So, even if it's fixed upstream, it just becomes redundant and doesn't break functionality.

tnevrlka (changes requested):
Thank you for the changes! 

The Fullsend bot seems to have removed the `needs-manual-review` label and considers the PR `ready-to-merge`. 

I am not comfortable letting an ADR be merged without a human review and am therefore blocking this PR from merging for now. 

I do not have any review comments at this time.

ralphbean:
@tnevrlka note,

image

That review agent is not authorized to approve anything on this repository (it's a grey checkmark not a green checkmark), so it is equivalent to a random drive by contributor approving. It is informational only.

To my knowledge, nobody anywhere has granted that review bot the rights to _really_ approve something. Nobody has added it as a maintainer to their repo (and limited its power with CODEOWNERS) yet.

If you think it has done the wrong thing, use `/fs-retro` to indicate that the retrospective agent should investigate why it came to different conclusions than you did. We then use that signal to improve it.

arewm:
Thanks for your recent changes, @JamesOlaitan . There were many comments on this ADR and most of them are now outdated. Would you be able to respond to the open threads with how you addressed the comments and resolve the ones that are indeed addressed?

cit1zen:
@konflux-ci/kgc moved it to the next milestone (2026-06-11), due to open questions and discussions.

tnevrlka:
> That review agent is not authorized to approve anything on this repository (it's a grey checkmark not a green checkmark), so it is equivalent to a random drive by contributor approving. It is informational only.
> 
> To my knowledge, nobody anywhere has granted that review bot the rights to _really_ approve something. Nobody has added it as a maintainer to their repo (and limited its power with CODEOWNERS) yet.

@ralphbean Thanks for the reply. In that case, I don't really understand the point of `requires-manual-review` and `ready-for-merge`. Is it just there as a proof of concept for the future fully automated luxury ADLC?

ralphbean:
Yeah, they just represent the agent's opinion. If the agent isn't a trusted maintainer, then they represent the opinion of an untrusted contributor.

ADR/0067-reproducible-container-builds.md
JamesOlaitan: It's among the open questions. In my test (`dnf install -y curl`), /var/lib/rpm was identical across the two runs. Project Hummingbird's testing found ones where the rpmdb wasn't byte-stable, which they fixed with the DELETE mode after install (similar in a way to ` rm -rf /var/log/dnf*`). So, to prevent a case like that, it's either a task changes the user's Containerfile by running the delete mode after install (which we want to avoid), OR an upstream fix in Fedora's default RPM config OR we pull the images from, say, Hummingbird's catalog since they already run that step at build time.

ADR/0067-reproducible-container-builds.md
JamesOlaitan: I've added a "Roadmap to default on" section in the ADR.

ADR/0067-reproducible-container-builds.md
JamesOlaitan: I created a table for the sources of non-determinism and moved the findings to an evidence column, so each source is beside the empirical observation that backs it.

ADR/0067-reproducible-container-builds.md
JamesOlaitan: Yes. I've changed the wording to "the source recorded in its SLSA provenance"

ADR/0067-reproducible-container-builds.md
JamesOlaitan: both actually. I ran an initial experiment with podman build before I was able to set up Konflux locally, after which I used buildah-oci-ta (which calls buildah build) for the findings I used to back the ADR.

ADR/0067-reproducible-container-builds.md
JamesOlaitan: no. install order is about which package gets installed when, which affects file ordering in the layer. rpmdb  byte-stability is about the rpm database itself ending up byte-different even when the install order is identical (because of how SQLite writes things in WAL mode). I added them as separate rows in the table in the ADR.

ADR/0067-reproducible-container-builds.md
JamesOlaitan: Which API? Tag-based backwards-compat convention. Actual Referrers API doesn't hit it.

Tag-based data deleted when digest is re-pushed? Yes. The tag is overwritten, then the old blob becomes untagged and GC'd.

Other metadata that might not be reproducible? Yes. If syft (or any secondary-artifact producer) is non-deterministic, the two blob digests differ even when the primary digest is identical.

cosign tool difference? Yes. re-pushing the same image manifest is fine because OCI preserves it by digest. cosign's tag convention is where it goes wrong.

Would Referrers API solve it? Yes.

ADR/0067-reproducible-container-builds.md
JamesOlaitan: You are right. `--source-date-epoch` only sets the config `created` field, and `--rewrite-timestamp` is the one that clamps the mtime across the image using the `--source-date-epoch` value. I've corrected it in the updated ADR (the file mtimes row in the What We Can Mitigate Today table).

ADR/0067-reproducible-container-builds.md
JamesOlaitan: I fixed this by using tables, so each row has its own mitigation (how it's fixed) and an owner column (where the fix lives), so a reader won't have to jump around.

ADR/0067-reproducible-container-builds.md
JamesOlaitan: Thanks for the link. I see that the injection is deterministic, so identical `PREFETCH_INPUT` produces byte-stable output. I've removed it from the open questions.

ADR/0067-reproducible-container-builds.md
JamesOlaitan: Done. In the updated ADR, I explained why I think build-twice-and-compare shouldn't be added (not every user will want a verification at every moment, so we shouldn't force every user to pay that cost, and it allows us to cleanly separate "build failed" from "verification failed"). I also stated the drawback that comes with a separate verification pipeline: reproducing on the same platform doesn't defend against an attack on the platform itself.

ADR/0067-reproducible-container-builds.md
JamesOlaitan: https://github.com/konflux-ci/architecture/issues/299#issuecomment-3656614799. 

The issue was described as "the build would happen in two stages. First, the tool fetches all the inputs ... and then the build happens with --net=none. This needs to be separable as two invocations."

ADR/0067-reproducible-container-builds.md
JamesOlaitan: I removed them.

ADR/0067-reproducible-container-builds.md
JamesOlaitan: They are now linked

ADR/0067-reproducible-container-builds.md
JamesOlaitan: Removed.

ADR/0067-reproducible-container-builds.md
JamesOlaitan: This shaped my SKIP_INJECTIONS rewrite.  buildah's --identity-label defaults to off when  `--source-date-epoch` is set, so io.buildah.version is handled at the buildah level. The reason it breaks today is that v0.9 of the task injects the label manually outside of buildah's control. The upcoming task version (the one delegating to konflux-build-cli) removes that manual injection, so SKIP_INJECTIONS=true isn't required for the cross-version case. SKIP_INJECTIONS role is limited to the content-sets.json injection.

ADR/0067-reproducible-container-builds.md
JamesOlaitan: It's now opt-in in the updated ADR draft because of the reason you specified. The "Default policy" subsection keeps all three new params (source-date-epoch, rewrite-timestamp, omit-history) default-off.

ADR/0067-reproducible-container-builds.md
JamesOlaitan: `SOURCE_DATE_EPOCH` is opt-in in the new draft. I also included your reasoning to explain the "why".

ADR/0067-reproducible-container-builds.md
JamesOlaitan: I was lumping them. The updated ADR's "Containerfile-side prerequisites" section splits them into two separate problems. 

io.buildah.version is a buildah-level concern that the upcoming task version handles (buildah's --identity-label suppresses the label when `SOURCE_DATE_EPOCH` is set; v0.9 manually injects it anyway, and the new task fixes that). 

content-sets.json is a separate injection used by third-party security scanners, byte-deterministic within a single task version, but its content shifts across versions. SKIP_INJECTIONS handles the second one only.

ADR/0067-reproducible-container-builds.md
JamesOlaitan: same source + same task version + same prefetched inputs → same digest. That's the TA-inputs-to-buildah-task scope, and it's what the verify-reproducibility pipeline asserts. The new draft spells this out. 

You are right that reproducing on the same platform doesn't defend against attacks on the platform itself, and I've also included that limitation in the ADR. The verification pipeline's interface (take a recipe and inputs, produce a reproducibility result) stays the same whether it's run on Konflux or by an outside party on a different platform, so limiting to the same-platform case today doesn't lock us out of extending later.

ADR/0067-reproducible-container-builds.md
JamesOlaitan: The Secondary artifacts and tag-collision section now explicitly states that the Conforma fix is necessary but not sufficient, and proposes two build-side options (off-primary-digest tag derivation or the OCI Referrers API) to keep the referenced blobs alive.

ADR/0067-reproducible-container-builds.md
JamesOlaitan: The opt-in claim is now limited to the three new pipeline parameters. Existing pipelines are unaffected, and only users who explicitly set the parameters get the new behavior. The rest of the proposal (the upcoming task version handling io.buildah.version, the build-image-index sort for multi-arch, the secondary-artifact follow-up) is not opt-in

ADR/0067-reproducible-container-builds.md
JamesOlaitan: No tool. The catalog is a reference for reviews, and I've updated the ADR to state this. When future PRs add logic to the buildah-oci-ta task, reviewers can check whether the change introduces or moves an entry between the two tables.

ADR/0067-reproducible-container-builds.md
JamesOlaitan: I proposed a fix in the Multi-architecture manifest ordering section in the updated ADR: build-image-index is our task, and a sort step there enforces a deterministic platform order, regardless of what buildah does. An upstream buildah fix would let us drop it later.

JamesOlaitan:
### Summary of the updates in the latest ADR:

The Sources of non-determinism section is now two tables: one for what Konflux can mitigate today, and one for what's still open work. Each row has an Evidence column with the empirical finding that backs the claim, so findings sit next to their source instead of being in a separate Findings section.

On rpmdb SQLite WAL: it's in the "What is still open work" table. My scenario (dnf install -y curl) had /var/lib/rpm byte-identical across runs, but Project Hummingbird hit it under broader scenarios and fixes it with a "parking" step (switching SQLite to DELETE journal mode after install). The fix lives either upstream in Fedora's rpm config or in a curated base image that bakes the parking step in at build time, since a Konflux task can't change rpm config without mutating the user's Containerfile.

For the SKIP_INJECTIONS framing, I had two different injected things lumped together in the first draft: the `io.buildah.version` label and the `content-sets.json` file. They're split now. The `io.buildah.version` label is actually handled at the buildah level — buildah's `--identity-label` flag defaults to off when `--source-date-epoch` is set, so when the user opts into reproducibility, buildah itself stops emitting the label. Version 0.9 of the buildah-oci-ta task injects the label manually outside of buildah's control, defeating that default. An upcoming task version delegates the build step to konflux-build-cli (a separate Go binary that wraps buildah), which lets buildah handle the label correctly. So users don't need `SKIP_INJECTIONS=true` for cross-task-version reproducibility. `SKIP_INJECTIONS` keeps a limited role around `content-sets.json`.

The default policy is now opt-in for the three new pipeline parameters (`source-date-epoch`, `rewrite-timestamp`, `omit-history`), with a four-condition roadmap to default-on. The "why opt-in" reasoning: if the build isn't actually reproducible (e.g. unpinned dependencies), auto-setting the image config's `created` field to the source commit timestamp would lie about when the artifact was assembled.

I also explained why a build-once-then-compare step doesn't fit a pipeline whose contract is to build once (rather than just citing PRs as authority). I framed the asserted property using SLSA's Verification Summary Attestation (the `slsa.buildReproduced` verified property) and stated that reproducing on the same build platform doesn't defend against attacks on the platform itself.

The "Secondary artifacts and tag-collision awareness" section now begins with the actual problem instead of explaining how it relates to reproducibility: build a commit, the pipeline pushes the image plus secondary artifacts (SBOM, attestation, signature) under cosign-style tags derived from the image digest. Trigger `/retest`. The second run is reproducible and produces the same image digest, so its secondary artifact pushes the same tags. The first run's blobs become untagged, the registry garbage-collects them, and the first run's attestation now references a blob that doesn't exist. The Conforma change handles getting the latest attestation on the verification side, but doesn't keep the referenced blobs alive on the build side. The section proposes two options for the build side: tag secondary-artifact blobs with a per-pipelineRun identifier instead of the image digest, or move to the OCI Referrers API. The ADR doesn't pick between them yet and commits to a follow-up.

The Multi-architecture manifest ordering section proposes a sort step inside `build-image-index` (which is our task) as the Konflux-side fix. A fix to the upstream buildah issue would let us drop the sort later, but it doesn't block our implementation.

One open question remains: whether Syft produces a byte-identical SBOM when run twice against a byte-identical input image. (The Hermeto-mutated Containerfile byte-stability question is no longer open. Tomas's link to konflux build-cli showed the injection is deterministic, so identical PREFETCH_INPUT produces byte-stable output.)

A few smaller things: ADR renumbered to 0069 (PR #354 had 0067), title field switched to the no-leading-zeros, `applies_to` in the table header is now build-service and pipeline-service, cross-repo PRs linked with full URLs, and the references to cachi2 were renamed to Hermeto.

ralphbean (approved):

cit1zen (approved):
LGTM

tnevrlka (approved):

PR #362, opened 2026-05-28
Add ADR: Standardize on konflux-ci.dev as Official API Group

mmorhun:
Konflux currently uses different API Groups for its Custom Resource Definitions (CRDs).
Majority of the resources have legacy `appstudio` API Group.
This ADR ensures we use single consistent `konflux-ci.dev` API Group for all CRDs.
All existing resources should be migrated.

PR #367, opened 2026-06-11
Replace Appstudio naming in k8s labels, annotations, finalizers and other names.

mmorhun:
## What this ADR is about

The main goal of this ADR is to replace legacy `appstudio` naming in various places especially in k8s labels, annotations and finalizers with `konflux-ci.dev` to be consistent with the API Group domain ([ADR](https://github.com/konflux-ci/architecture/pull/362)).

### General idea

Replace `appstudio` in labels, annotations, finalizers and other places with `konflux-ci.dev`.
The ADR does not restrict prefixes / suffixes, so both naming conventions are possible:
- `sundomain.konflux-ci.dev/name`
- `konflux-ci.dev/service-itemname`

While keeping old names with just replaced `appstudio` part is recommended, it is still not mandatory and if a rename is beneficial (in term of clarity or new cahnges) it must be done.

### Migration

The change above requires migration.
Overall plan:
1. Controllers and operators are updated to recognize and write BOTH old (`appstudio`) and new (`konflux-ci.dev`) metadata.
  2. New resources created during the transition period receive BOTH old and new metadata to ensure backward compatibility.
  3. Controllers are updated to prefer reading from new metadata but fall back to old metadata if new is absent.
  4. After a deprecation period, support for old metadata is removed and only `konflux-ci.dev` metadata is written and read.

There are exceptions for the above plan, when the migration can be simplified:
1. If the items to migrate are scoped to internal resources of only one service.
2. The migration could be fully automated on cross services internal resources.

Also, resources in the new Component model can use only new names.
In such case, the migration should be done during migration to the new Component model.

### Documentation

Documentation must be updated according to the renaming above.

### DIFF
--- a/ADR/0071-konflux-labels-annotations-finalizers.md
+++ b/ADR/0071-konflux-labels-annotations-finalizers.md
@@ -1,0 +1,275 @@
+---
+title: "71. Replace Appstudio Labels, Annotations, and Finalizers with Konflux equivalents"
+status: Accepted
+applies_to:
+  - "*"
+topics:
+  - api
+  - kubernetes
+  - labels
+  - annotations
+  - finalizers
+  - branding
+  - migration
+---
+
+# 71. Replace Appstudio Labels, Annotations, and Finalizers with Konflux equivalents
+
+Date: 2026-06-11
+
+## Status
+
+Accepted
+
+Builds upon [ADR 68. Standardize on konflux-ci.dev as Official API Group](0068-konflux-ci-dev-api-group.md).
+
+## Context
+
+Following the decision in ADR 68 to standardize on `konflux-ci.dev` as the official API group for all Konflux CRDs, a similar branding misalignment exists in Kubernetes labels, annotations, and finalizers currently used throughout the platform.
+
+Konflux currently uses various prefixes for labels, annotations, and finalizers that reflect the project's original identity as "App Studio" and its Red Hat-branded origins:
+
+- `appstudio.openshift.io/*` — Most common prefix for labels and annotations
+- `appstudio.redhat.com/*` — Alternative prefix in some services
+- `build.appstudio.openshift.io/*` — Build-service specific labels/annotations
+- `appstudio.application/*`, `appstudio.component/*` — Legacy label patterns
+- `pac.test.appstudio.openshift.io/*` — Pipelines as Code integration labels
+- Various finalizers with `appstudio` in their names
+
+### Issues with Current Naming
+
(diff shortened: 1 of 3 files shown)


rcerven (approved):

johnbieren (approved):

mmorhun:
Rebased and updated the ADR number. Added link to k8s docs.

mmorhun:
Fixed misspelled words.

ADR/0071-konflux-labels-annotations-finalizers.md
```diff
@@ -0,0 +1,275 @@
...
+
+**Common Finalizer Patterns:**
+- `appstudio.redhat.com/*` — Various service finalizers
+- `appstudio.openshift.io/*` — Alternative finalizer prefix
+- Service-specific finalizers with `appstudio` embedded
+
+The exact inventory of labels, annotations, and finalizers will be maintained in a tracking document separate from this ADR, as the list is large and changes frequently during active development.
+
+## Decision
+
```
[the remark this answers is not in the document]
mmorhun: We don't use such resources

mmorhun:
/fs-retro review comments are added at wrong line numbers.

ADR/0070-konflux-labels-annotations-finalizers.md
```diff
@@ -0,0 +1,275 @@
+---
+title: "70. Replace Appstudio Labels, Annotations, and Finalizers with Konflux equivalents"
```
cit1zen: For posterity, change of the ADR number will be required.

ADR/AGENTS.md
```diff
@@ -48,6 +48,7 @@ Status: Ac=Accepted Im=Implemented Ib=Implementable Pr=Proposed Ap=Approved Rp=R
 0042 Ac 259L Provisioning Clusters for Integration Tests
 0000 Im 31L Record architecture decisions
 0024 Ac 97L Release Objects Attribution Tracking and Propagation
+0070 Ac 275L Replace Appstudio Labels, Annotations, and Finalizers with Konflux equivalents
```
cit1zen: Change of ADR number will be required also here.

cit1zen (approved):
Minor administrative changes are required. Otherwise LGTM and I am looking forward to the change.

mmorhun:
@cit1zen I've rebased the PR and updated the ADR number.

ralphbean (approved):

arewm (approved):

PR #372, opened 2026-07-19
ADR 0072: Batched Nudging for Many-to-One Component Dependencies

gbenhaim:
## Summary

Introduces ADR 0072 proposing a debounce-based batching mechanism for nudge targets with many-to-one component dependencies.

## Problem

When multiple nudging components all point to the same nudged component (e.g., 30 operands → 1 bundle), every upstream build fires an independent nudge, causing redundant downstream rebuilds, integration tests, and releases.

## Solution

A `targetConfig` section on NudgeConfig spec declares per-target batch policies. When integration-service is about to fire a nudge for a batched target, it records the build result in NudgeConfig status and resets a debounce timer. A single aggregated Renovate PipelineRun fires only when the timer expires or a hard deadline is reached.

## Key Properties

- **No new CRDs** -- all state in NudgeConfig singleton status
- **Trigger-agnostic** -- works for CVE rebuilds, user pushes, dependency bumps
- **Opt-in** -- targets without `batchPolicy` continue immediate/validated nudging

Builds upon [ADR 67](https://github.com/konflux-ci/architecture/blob/main/ADR/0067-nudging-relationship-singleton-crd.md).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/ADR/0072-batched-nudging-for-many-to-one-dependencies.md
+++ b/ADR/0072-batched-nudging-for-many-to-one-dependencies.md
@@ -481,8 +481,7 @@ fires with whatever has accumulated, excluding failed components. When
 with at least one build result in either `accumulated` or `failed`)
 have succeeded -- if any component is still in `failed`, the action is
 rejected and the batch remains in its current phase. On rejection the
-controller emits a Kubernetes Event (type `Warning`, reason
-`ForceFireRejected`) on the NudgeConfig and sets a transient Condition
+controller sets a transient Condition on the NudgeConfig status
 (`type: ForceFireRejected`, `status: "True"`, `message` listing the
 failed components). The Condition is cleared on the next successful
 reconciliation. This lets users distinguish "fire now with partial
@@ -496,11 +495,12 @@ failures present). This ensures the field never persists in a stale
 state across restarts. When the action successfully fires, the
 controller transitions the batch to `Firing` and creates the aggregated
 Renovate PipelineRun with whatever has accumulated.
-When the action is ignored or rejected, the controller emits a
-Kubernetes Event (type `Warning`, reason `ForceFireIgnored` for the
-ignored case, `ForceFireRejected` for the rejected case) explaining why
-the action had no effect. This provides consistent feedback regardless
-of outcome.
+When the action is ignored or rejected, the controller sets a transient
+Condition on the NudgeConfig status (`ForceFireIgnored` for the ignored
+case, `ForceFireRejected` for the rejected case) explaining why the
+action had no effect. These Conditions are cleared on the next
+successful reconciliation, providing consistent feedback regardless of
+outcome.
 
 This handles:
 - **Blocked batches:** one component's build failed and won't be fixed
@@ -574,15 +574,26 @@ fixed.
 
 #### Blocked batch observability
 
-When a batch transitions to `Blocked`, the controller emits a Kubernetes
-Event (type `Warning`, reason `BatchBlocked`) on the NudgeConfig listing
-the failed component(s). If the batch remains in `Blocked` phase past
-the configured `maxWaitTime` (which is not enforced as a firing trigger
-for blocked batches, but serves as an alerting threshold), the controller
-emits a second Event (type `Warning`, reason `BatchBlockedPastDeadline`)
-prompting the user to investigate or use `forceFire`. This makes the
-escape hatch discoverable without requiring users to actively poll
-NudgeConfig status.
+When a batch transitions to `Blocked`, the controller sets a Condition
+on the NudgeConfig status (`type: BatchBlocked`, `status: "True"`,
+`message` listing the failed component(s) and the target). The Condition
+is cleared when the batch leaves the `Blocked` phase (e.g., via retry
+success or `forceFire`). This is the primary observability mechanism —
+users (and the UI) inspect the NudgeConfig status directly, which is
+already the authoritative source for batch progress.
+
+#### Why not Kubernetes Events?
+
+We deliberately avoid Kubernetes Events for lifecycle signalling:
+
+- **Not surfaced in the UI.** The Konflux UI does not expose Kubernetes
+  Events, so users would never see them without direct cluster access.
+- **etcd overhead.** Each Event is a separate object written to etcd.
+  The platform is already working to reduce Event TTLs to relieve
+  storage pressure; adding new Event sources works against that goal.
+- **No discoverability advantage.** Users must already inspect the
+  NudgeConfig status for batch progress. A Condition on the same CR is
+  strictly easier to find than a separate Event resource.
 
 #### Controller restart mid-batch
 
@@ -734,10 +745,11 @@ from the nudge edges themselves -- no external group reference needed.
   entry continue to use immediate or validated nudging unchanged. Users
   adopt batching per-target at their own pace.
 
-- **Structured observability.** The `activeBatches` status provides a
-  structured API that the UI can consume to display progress bars,
-  failure details, countdowns, and a "Force Fire" button. The controller
-  emits Kubernetes Events for key lifecycle transitions.
+- **Structured observability.** The `activeBatches` status and
+  Conditions provide a structured API that the UI can consume to display
+  progress bars, failure details, countdowns, and a "Force Fire" button.
+  All observability is on the NudgeConfig CR itself — no separate
+  resources to discover.
 
 - **Formally verified.** The batch state machine was verified with TLA+
   model checking (656,827 states explored), confirming no deadlocks, no
--- a/ADR/0072-batched-nudging-for-many-to-one-dependencies.md
+++ b/ADR/0072-batched-nudging-for-many-to-one-dependencies.md
@@ -311,6 +311,7 @@ status:
   # New: active batch tracking
   activeBatches:
     - target: bundle
+      batchId: bundle-1721126400    # <target>-<createdAt-epoch>
       phase: Blocked               # Accumulating | Blocked | Firing | Failed | Completed
       createdAt: "2026-07-16T10:00:00Z"
       fireAt: ""                     # paused while blocked
--- a/ADR/0072-batched-nudging-for-many-to-one-dependencies.md
+++ b/ADR/0072-batched-nudging-for-many-to-one-dependencies.md
@@ -229,6 +229,9 @@ spec:
 | `spec.targetConfig[].batchPolicy.debounceTimeout` | `duration` | Overrides `batchDefaults.debounceTimeout` for this target. |
 | `spec.targetConfig[].batchPolicy.maxWaitTime` | `duration` | Overrides `batchDefaults.maxWaitTime` for this target. |
 | `spec.targetConfig[].batchPolicy.failurePolicy` | `enum` | Overrides `batchDefaults.failurePolicy` for this target. |
+| `spec.actions.forceFire` | `object` | One-shot action field. When set, the controller force-fires the batch for the specified target regardless of debounce state. Cleared after processing. |
+| `spec.actions.forceFire.target` | `string` | Required. Name of the target component whose batch should be force-fired. Must match a target in `activeBatches`. |
+| `spec.actions.forceFire.includePartial` | `bool` | Optional. Default: `true`. When `true`, fires with whatever has accumulated (excluding failed). When `false`, fires only if no component is in `failed`; otherwise the action is rejected (see Manual Override). |
 
 **Why batching is declared on the target, not per-edge:** Batching is
 inherently a property of the target component -- it controls how the
@@ -311,7 +314,7 @@ stateDiagram-v2
     direction LR
     [*] --> Accumulating : build completes
     Accumulating --> Accumulating : +build
-    Accumulating --> Blocked : failure
+    Accumulating --> Blocked : failure [Block]
     Accumulating --> Firing : timer / forceFire
     Blocked --> Accumulating : fix
     Blocked --> Firing : forceFire
@@ -414,9 +417,14 @@ fires with whatever has accumulated, excluding failed components. When
 `false`, the batch fires only if all known nudging components (those
 with at least one build result in either `accumulated` or `failed`)
 have succeeded -- if any component is still in `failed`, the action is
-rejected and the batch remains in its current phase. This lets users
-distinguish "fire now with partial results" from "fire now but only if
-everything succeeded."
+rejected and the batch remains in its current phase. On rejection the
+controller emits a Kubernetes Event (type `Warning`, reason
+`ForceFireRejected`) on the NudgeConfig and sets a transient Condition
+(`type: ForceFireRejected`, `status: "True"`, `message` listing the
+failed components). The Condition is cleared on the next successful
+reconciliation. This lets users distinguish "fire now with partial
+results" from "fire now but only if everything succeeded" and provides
+clear feedback when the action cannot proceed.
 
 The controller processes the action, transitions the batch to `Firing`,
 creates the aggregated Renovate PipelineRun with whatever has
@@ -447,10 +455,12 @@ component simply never contributes a result.
 
 #### Overlapping batches for the same target
 
-Only one active batch exists per target at any time. This is a hard
-invariant -- concurrent batches for the same target risk digest
-regression (a later-firing batch could overwrite a newer digest with an
-older one).
+At most one batch per target accepts new builds at any time (the
+two-slot concurrency model allows a second batch in
+`Firing`/`Failed`/`Completed`, but only the current accepting batch
+receives new results). This serialization invariant prevents digest
+regression -- a later-firing batch could overwrite a newer digest with
+an older one if concurrent accepting batches were allowed.
 
 - **`Accumulating` or `Blocked` phase:** New builds are added to the
   existing batch (debounce timer resets).
--- a/ADR/0072-batched-nudging-for-many-to-one-dependencies.md
+++ b/ADR/0072-batched-nudging-for-many-to-one-dependencies.md
@@ -394,14 +394,6 @@ when builds arrive during `Firing`/`Failed`/`Completed`.
 | **Failed** | Recorded in **next batch** (create if needed → next `Accumulating`). This batch stays `Failed`. | Same -- recorded in **next batch**. | n/a | → `Firing`. Create new PLR with same `accumulated`. | n/a | n/a |
 | **Completed** | **Next batch** created (→ next `Accumulating`). This batch cleaned up. | Same -- recorded in **next batch**. This batch cleaned up. | n/a | Ignored (nothing to fire). | n/a | n/a |
 
-The state machine was formally verified with TLA+ model checking. TLC
-explored 656,827 states with 3 components, 12 events, and both failure
-policies. All invariants hold: types are correct, `Firing` always has
-non-empty `accumulated`, and at most one batch is in `Firing` at any
-time.
-
-The TLA+ specification will be published in a companion repository once
-the ADR is accepted. It is available for review on request.
 
 ### Batch Grouping
 
@@ -754,9 +746,6 @@ from the nudge edges themselves -- no external group reference needed.
   All observability is on the NudgeConfig CR itself — no separate
   resources to discover.
 
-- **Formally verified.** The batch state machine was verified with TLA+
-  model checking (656,827 states explored), confirming no deadlocks, no
-  empty-firing, and serial firing invariants.
 
 ### Negative
 
--- a/ADR/0072-batched-nudging-for-many-to-one-dependencies.md
+++ b/ADR/0072-batched-nudging-for-many-to-one-dependencies.md
@@ -1,0 +1,625 @@
+---
+title: "72. Batched Nudging for Many-to-One Component Dependencies"
+status: Proposed
+applies_to:
+  - integration-service
+topics:
+  - nudging
+  - batching
+  - scalability
+---
+
+# 72. Batched Nudging for Many-to-One Component Dependencies
+
+Date: 2026-07-19
+
+## Status
+
+Proposed
+
+Builds upon [ADR 67. Nudging Relationship Storage via Singleton CRD](0067-nudging-relationship-singleton-crd.md).
+Implements the "batched nudging for operator bundles" phase deferred by
+ADR 67.
+
+## Context
+
+When multiple nudging components all point to the same nudged component
+(many-to-one), every upstream build fires an independent nudge. This
+creates a cascade of redundant downstream rebuilds, integration tests,
+and releases.
+
+**Concrete example:** An OLM operator with 30 operand images and one
+bundle. All operand images declare `build-nudges-ref` (or NudgeConfig
+edges) to the bundle. When a UBI CVE triggers rebuilds of all 30
+operands, the bundle is rebuilt 30 times. Only the final rebuild
+contains all 30 updated digests -- the first 29 are immediately
+superseded.
+
+**Impact at scale:**
+- **Resource waste:** 29 unnecessary Renovate PipelineRuns + 29 bundle
+  builds + 29 integration test suites + 29 potential releases. Each
+  PipelineRun consumes 10-50 KB of etcd storage and cluster compute.
+- **Inconsistent state:** If any operand build fails, the bundle ends up
+  in a partially updated state with no mechanism to detect or remediate
+  the partial update.
+- **Notification spam:** Users receive 30 sets of build/test/release
+  notifications for what is logically one event.
+
+### Current Architecture
+
+[ADR 67](0067-nudging-relationship-singleton-crd.md)
+introduced the `NudgeConfig` singleton CRD and migrated nudging
+orchestration from build-service to integration-service. It defines two
+modes:
+
+- **`immediate`** -- nudge fires when the source build PipelineRun
+  succeeds (current default, replicating ADR 29 behavior).
+- **`validated`** -- nudge fires after the gating ComponentGroup's
+  Snapshot passes integration tests (Phase 2 of ADR 67).
+
+### Constraints
+
+- **etcd pressure.** The Konflux infrastructure team has identified etcd
+  as the primary scaling bottleneck. This design must not introduce new
+  CRDs or per-batch objects. All state lives in the existing NudgeConfig
+  status.
+- **Pruner compatibility.** Tekton PipelineRuns are pruned after
+  completion. Batch state must be captured before pruning occurs.
+- **Trigger-agnostic.** The solution must work regardless of what
+  triggered the upstream builds (MintMaker CVE updates, user-initiated
+  changes, or any other source).
+
+### Alternatives Considered
+
+#### Orchestrator Pipeline (Matrix + fan-in)
+
+Changes the build model. Today, PaC triggers builds per git event. An
+orchestrator pipeline would need to own build triggering, conflicting
+with PaC. Also doesn't handle independently-arriving builds.
+
+#### Debounce at the Renovate/PR level
+
+Renovate already supports branch reuse (`build-nudge-simple-branch`), so
+multiple nudges update the same PR. But each PR update still triggers a
+rebuild of the nudged component. This reduces PR spam but not build spam.
+
+#### MintMaker-initiated explicit grouping
+
+Only works for MintMaker-initiated changes. The user requirement is
+trigger-agnostic. Also requires MintMaker to propagate group metadata
+through PaC to PipelineRuns, which PaC doesn't natively support.
+
+#### ChangeGroup CR ([ADR PR #308](https://github.com/konflux-ci/architecture/pull/308))
+
+Detailed proposal using a MutatingWebhook to intercept nudge
+PipelineRuns, a ChangeGroup CR to track membership, and `[skip ci]` on
+draft PRs. The debounce approach is preferred for several reasons:
+
+1. **CVE-centric design** -- the ChangeGroup proposal is framed around
+   coordinated rebuilds where a known event (e.g., a UBI CVE fix)
+   triggers many components at once. But batching is equally valuable
+   for organic, unrelated builds that happen to arrive close together
+   (routine dependency bumps, independent feature merges). The
+   ChangeGroup model requires someone to anticipate each batch and
+   create a CR for it, making it unsuitable for these everyday cases.
+   The debounce approach is trigger-agnostic -- it batches any builds
+   that arrive within the window regardless of why they were triggered.
+
+2. **Explicit group creation required** -- someone (a user, or
+   automation that doesn't exist yet) must create a ChangeGroup CR
+   before each batching event, specifying which components to wait for.
+   The debounce approach is fully automatic once `batchPolicy` is
+   configured -- it collects whatever builds arrive within the debounce
+   window with no per-event setup.
+
+3. **Membership must be known upfront** -- the ChangeGroup spec lists
+   `nudgingComponents` at creation time, which means the creator must
+   predict which components will rebuild. If an unexpected component
+   rebuilds (e.g., a transitive dependency), it falls outside the group.
+   The debounce approach doesn't require an enumeration -- any build for
+   a component with a nudge edge to the batched target is automatically
+   collected.
+
+4. **Misaligned with ADR 67** -- the ChangeGroup ADR was written before
+   ADR 67 moved nudging ownership to integration-service. The
+   MutatingWebhook intercepts nudge PipelineRuns created by
+   build-service, but [the ADR itself notes](https://github.com/konflux-ci/architecture/pull/308#:~:text=if%20integration-service%20embeds%20nudging)
+   that the approach breaks if integration-service embeds nudging as a
+   step rather than a separate PipelineRun. The debounce approach is
+   built into integration-service from the start, fully aligned with
+   ADR 67.
+
+5. **Risk of delivering stale updates** -- nothing prevents multiple
+   ChangeGroup CRs for the same `nudgedComponent`. If ChangeGroup A and
+   ChangeGroup B both target bundle-x, they collect digests
+   independently. If A fires after B but contains older digests for
+   overlapping components, merging A's PR regresses the bundle to
+   outdated images. The ChangeGroup design has no single-batch-per-target
+   invariant. The debounce approach enforces exactly one active batch per
+   target in NudgeConfig status, so a later batch always includes the
+   results of an earlier one and digest regression is impossible.
+
+6. **Operational risk** -- the MutatingWebhook sits in the admission
+   path for all nudge PipelineRun creation; any webhook latency or
+   downtime blocks builds across the cluster. The debounce approach runs
+   as part of integration-service's existing reconciliation loop with no
+   admission-path dependency.
+
+7. **Dual state machine** -- state lives in both the ChangeGroup CR and
+   the git PR (draft status, `[skip ci]`, commit history). If these
+   diverge (e.g., a user edits the PR description, or a force-push
+   race), reconciliation is complex. The debounce approach has a single
+   state machine in NudgeConfig status.
+
+8. **etcd pressure** -- each batching event creates a new ChangeGroup
+   CR, so a namespace with many targets produces many CRs. The debounce
+   approach stores all batch state in the existing NudgeConfig singleton
+   status (zero new objects).
+
+The ChangeGroup design doc is a valuable reference for edge cases
+(incremental updates, manual intervention, PR status tables).
+
+## Decision
+
+Introduce a **`targetConfig`** section on the NudgeConfig spec that
+declares per-target batch policies. A target is batched if and only if
+it has a `batchPolicy` in its `targetConfig` entry. When
+integration-service is about to fire a nudge for a target that has a
+`batchPolicy`, it records the build result in the NudgeConfig status and
+resets a debounce timer instead. A single aggregated Renovate
+PipelineRun fires only when the debounce timer expires (no new builds
+for the same target within the quiet period) or a hard deadline is
+reached.
+
+Batching is orthogonal to the nudge mode defined in ADR 67. In
+`immediate` mode the accumulation trigger is a successful build; in
+`validated` mode the trigger is a passing Snapshot. Either way, the
+batch collects the result and delays the Renovate PipelineRun.
+
+Targets without a `targetConfig` entry (or without `batchPolicy`)
+continue to use their configured mode (`immediate` or `validated`)
+without batching.
+
+### NudgeConfig Spec Changes
+
+```yaml
+apiVersion: konflux-ci.dev/v1alpha1
+kind: NudgeConfig
+metadata:
+  name: nudge-config
+  namespace: my-tenant
+spec:
+  batchDefaults:
+    debounceTimeout: 30m        # quiet period before firing
+    maxWaitTime: 4h             # hard cap regardless of debounce resets
+    failurePolicy: Block        # Block | ProceedWithPartial
+
+  targetConfig:
+    - target: bundle
+      batchPolicy:
+        debounceTimeout: 15m
+        maxWaitTime: 2h
+        failurePolicy: ProceedWithPartial
+
+    - target: fbc-catalog
+      batchPolicy: {}           # uses batchDefaults
+
+  nudges:
+    - from: operand-a
+      to: bundle              # batched (bundle has batchPolicy)
+    - from: operand-b
+      to: bundle              # batched (same reason)
+    # ... all operand → bundle edges are automatically batched
+
+    # Non-batched edges: target has no targetConfig entry
+    - from: standalone-lib
+      to: standalone-consumer # immediate (no targetConfig)
+```
+
+**Fields:**
+
+| Field | Type | Description |
+|---|---|---|
+| `spec.batchDefaults.debounceTimeout` | `duration` | Time to wait after the last build event before firing the batch. Default: `30m`. Range: 1m--24h. |
+| `spec.batchDefaults.maxWaitTime` | `duration` | Hard deadline from the first build event. Fires the batch when reached regardless of debounce resets. Only applies during `Accumulating` phase -- does not override `Block` (a blocked batch stays blocked until the failure is resolved or the user triggers `forceFire`). Default: `4h`. Must exceed `debounceTimeout`. |
+| `spec.batchDefaults.failurePolicy` | `enum` | `Block`: batch transitions to `Blocked` if any member build fails; it stays blocked indefinitely (ignoring `maxWaitTime`) until the user fixes the failure or uses `forceFire`. `ProceedWithPartial`: batch fires with whatever succeeded when the debounce or hard deadline expires; if all components failed (empty `accumulated`), the batch transitions to `Failed` instead of `Firing` -- there is nothing to nudge. Default: `Block`. |
+| `spec.targetConfig` | `[]TargetConfig` | Per-target configuration. A target with a `batchPolicy` receives batched nudges; targets not listed here (or listed without `batchPolicy`) receive immediate nudges. |
+| `spec.targetConfig[].target` | `string` | Name of the nudged (target) component. Must match the `to` value of at least one nudge edge. |
+| `spec.targetConfig[].batchPolicy` | `object` | Batch policy for this target. Presence of this field opts the target into batched mode. Fields omitted from `batchPolicy` fall back to `batchDefaults`. An empty object (`{}`) means "use all defaults." |
+| `spec.targetConfig[].batchPolicy.debounceTimeout` | `duration` | Overrides `batchDefaults.debounceTimeout` for this target. |
+| `spec.targetConfig[].batchPolicy.maxWaitTime` | `duration` | Overrides `batchDefaults.maxWaitTime` for this target. |
+| `spec.targetConfig[].batchPolicy.failurePolicy` | `enum` | Overrides `batchDefaults.failurePolicy` for this target. |
+
+**Why batching is declared on the target, not per-edge:** Batching is
+inherently a property of the target component -- it controls how the
+target receives updates. Declaring it per-edge would be redundant (all
+edges to the same target must behave the same way, otherwise one
+immediate edge would bypass the batch) and error-prone (adding a new
+edge requires remembering to set the mode). With `targetConfig`, adding
+a new nudge edge to a batched target automatically inherits the batch
+behavior.
+
+**Batch policy resolution:** For a batch targeting component X, the
+controller looks for an entry in `targetConfig` where `target` equals X
+and `batchPolicy` is present. Fields set in `batchPolicy` take
+precedence over `batchDefaults`; omitted fields fall back to defaults.
+
+**Validation rules:**
+
+1. No duplicate `target` values in `targetConfig`.
+2. Each `targetConfig[].target` must match the `to` value of at least
+   one nudge edge. Orphan entries are rejected.
+3. `maxWaitTime` must be greater than `debounceTimeout` (checked on both
+   `batchDefaults` and each resolved policy with fallback).
+4. `debounceTimeout` must be between 1m and 24h.
+
+### NudgeConfig Status Changes
+
+```yaml
+status:
+  # Existing conditions from ADR 67
+  conditions:
+    - type: Valid
+      status: "True"
+
+  # New: active batch tracking
+  activeBatches:
+    - target: bundle
+      phase: Blocked               # Accumulating | Blocked | Firing | Failed | Completed
+      createdAt: "2026-07-16T10:00:00Z"
+      fireAt: ""                     # paused while blocked
+      hardDeadline: "2026-07-16T14:00:00Z"
+
+      accumulated:
+        - from: operand-a
+          imageDigest: "sha256:abc123..."
+          buildPipelineRun: build-operand-a-xyz
+          capturedAt: "2026-07-16T10:00:00Z"
+        - from: operand-b
+          imageDigest: "sha256:def456..."
+          buildPipelineRun: build-operand-b-abc
+          capturedAt: "2026-07-16T10:02:00Z"
+
+      failed:
+        - from: operand-c
+          buildPipelineRun: build-operand-c-fail
+          reason: "TaskRun build-task failed"
+          capturedAt: "2026-07-16T10:10:00Z"
+
+      nudgePipelineRun: ""        # set when phase transitions to Firing
+      message: "2 accumulated, 1 failed (operand-c), batch blocked"
+```
+
+### Batch Phase State Machine
+
+At most two batches can coexist for the same target: the **current**
+batch (any phase) and a **next** batch (`Accumulating` or `Blocked`
+only). The next batch is created when builds arrive while the current
+batch is in `Firing`, `Failed`, or `Completed`. The next batch cannot
+transition to `Firing` while the current batch is in `Firing` phase
+(serialization). Once the current batch reaches `Completed` or `Failed`,
+the next batch is promoted to current and proceeds normally.
+
+**Batch creation trigger:** Any build completion (success or failure)
+for a nudging component whose target has a `batchPolicy` creates a new
+batch if none exists. A successful build adds to `accumulated`; a
+failed build adds to `failed` (and may immediately transition the
+batch to `Blocked` under `failurePolicy: Block`).
+
+```mermaid
+stateDiagram-v2
+    direction LR
+    [*] --> Accumulating : build completes
+    Accumulating --> Accumulating : +build
+    Accumulating --> Blocked : failure
+    Accumulating --> Firing : timer / forceFire
+    Blocked --> Accumulating : fix
+    Blocked --> Firing : forceFire
+    Firing --> Completed : PLR ok
+    Firing --> Failed : PLR fail
+    Failed --> Firing : forceFire
+    Failed --> [*] : superseded
+    Completed --> [*] : cleanup
+
+    note right of Firing
+        Builds start a next batch.
+        Next batch cannot fire
+        until this one completes.
+    end note
+
+    note right of Failed
+        Builds start a next batch.
+        Does not block next batch.
+        Superseded when next fires.
+    end note
+```
+
+| Phase | Meaning |
+|---|---|
+| `Accumulating` | Collecting build results. Debounce timer running. Builds are added to this batch. |
+| `Blocked` | A member build failed and `failurePolicy: Block`. Debounce paused. Builds still added to this batch. Unblocks if the failed component rebuilds successfully. |
+| `Firing` | Aggregated Renovate PLR created. Builds arriving now start a **next** batch that cannot fire until this batch completes. |
+| `Failed` | Renovate PLR failed (retries exhausted). Does **not** block the next batch from firing. Superseded when the next batch fires. |
+| `Completed` | Renovate PLR succeeded. Retained for observability. Cleaned up when next batch starts or after 24h. |
+
+**Transition table** -- every (state, event) pair. "This batch" refers
+to the current batch; "next batch" to the second batch slot created
+when builds arrive during `Firing`/`Failed`/`Completed`.
+
+| State | Build OK | Build fail | Timer expires | forceFire | PLR OK | PLR fail |
+|---|---|---|---|---|---|---|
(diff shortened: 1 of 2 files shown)


ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,607 @@
+---
+title: "72. Batched Nudging for Many-to-One Component Dependencies"
+status: Proposed
+applies_to:
+  - integration-service
+topics:
+  - nudging
+  - batching
+  - scalability
+---
```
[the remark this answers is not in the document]
gbenhaim: Not an issue — the old `generate-adr-index.yml` workflow (which used `head -1`) was deleted. The current index generator (`hack/util/generate-adr-table`) is frontmatter-aware and extracts titles correctly. All 69 existing ADRs in the repo use YAML frontmatter, and the lint CI validates them all successfully.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,607 @@
...
+
+Batches are grouped implicitly by target component: all nudge edges
+pointing to a target with a `batchPolicy` share one active batch. Only
+one active batch per target exists at any time. No explicit group ID
+is needed.
+
+When a build completes for a nudging component whose target has a
+`batchPolicy`, integration-service looks up the active batch for that
+target in `status.activeBatches`. If none exists, a new batch is
+created. If one exists, the build result is appended and the debounce
```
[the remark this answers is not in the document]
gbenhaim: Good catch — reworded in d90a181 to: "At most one batch per target is *accepting new builds* at any time" with explicit reference to the two-slot concurrency model.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,607 @@
...
+to fire early. Following the action pattern from the
+[revised component model](0056-revised-component-model.md):
+
+```yaml
+spec:
+  actions:
+    forceFire:
+      target: bundle
+      includePartial: true    # proceed even if some builds failed
+```
```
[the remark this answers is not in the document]
gbenhaim: Fixed in d90a181 — added a paragraph documenting that `includePartial` defaults to `true`, and that when `false` the action is rejected if any component is still in `failed` (letting users distinguish "fire with partial" from "fire only if all succeeded").

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,607 @@
...
+(serialization). Once the current batch reaches `Completed` or `Failed`,
+the next batch is promoted to current and proceeds normally.
+
+```mermaid
+stateDiagram-v2
+    direction LR
+    [*] --> Accumulating : build arrives
+    Accumulating --> Accumulating : +build
+    Accumulating --> Blocked : failure
+    Accumulating --> Firing : timer / forceFire
```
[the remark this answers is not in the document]
gbenhaim: Clarified in d90a181 — added an explicit paragraph before the mermaid diagram stating that any build completion (success or failure) creates a batch if none exists, and updated the diagram label from "build arrives" to "build completes".

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,625 @@
...
+nudge branches from non-batched edges.
+
+### Manual Override
+
+Users need an escape hatch for batches that are stuck or that they want
+to fire early. Following the action pattern from the
+[revised component model](0056-revised-component-model.md):
+
+```yaml
+spec:
```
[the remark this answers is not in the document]
gbenhaim: Fixed in 9ce48fe — added `spec.actions.forceFire`, `spec.actions.forceFire.target`, and `spec.actions.forceFire.includePartial` to the fields table with types, required/optional status, and descriptions.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,625 @@
...
+
+#### Overlapping batches for the same target
+
+Only one active batch exists per target at any time. This is a hard
+invariant -- concurrent batches for the same target risk digest
+regression (a later-firing batch could overwrite a newer digest with an
+older one).
+
+- **`Accumulating` or `Blocked` phase:** New builds are added to the
+  existing batch (debounce timer resets).
```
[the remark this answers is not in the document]
gbenhaim: Fixed in 9ce48fe — reworded to: "At most one batch per target accepts new builds at any time (the two-slot concurrency model allows a second batch in `Firing`/`Failed`/`Completed`, but only the current accepting batch receives new results)."

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,625 @@
...
+stateDiagram-v2
+    direction LR
+    [*] --> Accumulating : build completes
+    Accumulating --> Accumulating : +build
+    Accumulating --> Blocked : failure
+    Accumulating --> Firing : timer / forceFire
+    Blocked --> Accumulating : fix
+    Blocked --> Firing : forceFire
+    Firing --> Completed : PLR ok
+    Firing --> Failed : PLR fail
```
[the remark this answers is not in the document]
gbenhaim: Fixed in 9ce48fe — added guard condition to the mermaid transition: `Accumulating --> Blocked : failure [Block]`. The diagram is intentionally simplified (the transition table has full detail), but the guard makes it clear this only applies under `failurePolicy: Block`.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,625 @@
...
+```
+
+**`includePartial` field:** Default: `true`. When `true`, the batch
+fires with whatever has accumulated, excluding failed components. When
+`false`, the batch fires only if all known nudging components (those
+with at least one build result in either `accumulated` or `failed`)
+have succeeded -- if any component is still in `failed`, the action is
+rejected and the batch remains in its current phase. This lets users
+distinguish "fire now with partial results" from "fire now but only if
+everything succeeded."
```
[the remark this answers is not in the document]
gbenhaim: Fixed in 9ce48fe — specified that on rejection the controller emits a Kubernetes Event (type `Warning`, reason `ForceFireRejected`) and sets a transient `ForceFireRejected` Condition on NudgeConfig status with a message listing the failed components. The Condition is cleared on the next successful reconciliation.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,688 @@
...
+integration-service is about to fire a nudge for a target that has a
+`batchPolicy`, it records the build result in the NudgeConfig status and
+resets a debounce timer instead. A single aggregated Renovate
+PipelineRun fires only when the debounce timer expires (no new builds
+for the same target within the quiet period) or a hard deadline is
+reached.
+
+Batching is orthogonal to the nudge mode defined in ADR 67. In
+`immediate` mode the accumulation trigger is a successful build; in
+`validated` mode the trigger is a passing Snapshot. Either way, the
```
kasemAlem, line 184: validated mode interaction is underspecified.
  
  This sentence is the only specification for how batching behaves in validated mode. Three things are missing:
  
   1. Image digest source: In validated mode the relevant digest is the one from the passing Snapshot (the tested image), not the raw build output. The accumulated entry schema needs to clarify which one is 
   stored.
   2. Accumulation trigger when components share a gating group: If operand-a and operand-b share the same gatingGroup, a single Snapshot covers both. Does that one passing Snapshot add two entries to 
   accumulated, or one?
   3. What counts as a "failure" for failurePolicy: Is it a failed build PLR, a failed Snapshot (integration tests failed), or something else? The semantics differ meaningfully.
  
   Suggest adding a short sub-section under the Decision that answers these three questions for validated mode, similar to how the "Batch Phase State Machine" fully specifies the immediate mode flow

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,688 @@
...
+
+Introduce a **`targetConfig`** section on the NudgeConfig spec that
+declares per-target batch policies. A target is batched if and only if
+it has a `batchPolicy` in its `targetConfig` entry. When
+integration-service is about to fire a nudge for a target that has a
+`batchPolicy`, it records the build result in the NudgeConfig status and
+resets a debounce timer instead. A single aggregated Renovate
+PipelineRun fires only when the debounce timer expires (no new builds
+for the same target within the quiet period) or a hard deadline is
+reached.
```
kasemAlem, line 172: The debounce timer is implemented as a fireAt timestamp in status. On each build result capture, the controller updates fireAt = now + debounceTimeout and schedules RequeueAfter: debounceTimeout. 

On controller startup, the informer cache triggers a reconcile for every NudgeConfig, allowing expired fireAt values to be detected and fired immediately, and future fireAt values to be re-queued with the 
   appropriate remaining delay.

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,688 @@
...
+If the aggregated Renovate PipelineRun fails, the batch transitions to
+`Failed`. The controller retries by creating a new PipelineRun with an
+incremented attempt suffix (e.g., `nudge-batch-bundle-1721126400-retry1`)
+up to 3 times. After exhausting retries, the batch stays in `Failed`
+and the user can either investigate the failure, fix the underlying
+issue, and use `forceFire` to re-trigger, or manually clean up the
+batch.
+
+#### PipelineRun pruning during a batch
+
```
kasemAlem, line 644: Write ordering not specified — silent data loss window.
  
   This section describes the annotation mechanism but doesn't specify which write happens first. 
If the controller writes the component-nudge-processed annotation on the PLR before updating activeBatches in NudgeConfig status, a crash between the two leaves the build result permanently lost — the PLR is already marked processed and won't be re-evaluated on restart.
  
   The safe ordering is: write NudgeConfig status first, annotate the PLR second. A crash after the status write but before the annotation is safe — on restart the PLR has no annotation, the controller re-reads 
   it, and the dedup-by-from rule means writing the same result again is harmless.
  
   Please add one sentence stating the required write ordering, e.g.:
  
   ▎ "The NudgeConfig status update must be persisted before the component-nudge-processed annotation is written to the PLR. This ordering ensures that a controller crash between the two writes is always 
   recoverable — the PLR will be re-processed on restart and the dedup rule prevents duplicate entries."

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,688 @@
...
+created. If one exists, the build result is appended and the debounce
+timer is reset.
+
+**Deduplication:** If the same `from` component builds multiple times
+during a batch (e.g., a retry), the latest result replaces the previous
+one. Only the most recent digest per nudging component is included in the
+aggregated nudge.
+
+### Aggregated Renovate PipelineRun
+
```
kasemAlem, line 427: Batched PLR annotation schema is not defined.
  
   The existing single-component nudge PLR has four annotations (nudged-components, nudging-component, nudging-pipeline, nudging-image) that integration-service, the UI, and tests all depend on. For a batched PLR with N source components, none of these annotations generalize cleanly — particularly nudging-component, nudging-pipeline, and nudging-image which are currently single-valued.
  
   The ADR should define the multi-component annotation shape explicitly (e.g., comma-separated lists, JSON arrays, or new annotation names) so that integration-service and the UI can be implemented 
   consistently. Without this, the two will independently invent incompatible schemas.

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,688 @@
...
+a new nudge edge to a batched target automatically inherits the batch
+behavior.
+
+**Batch policy resolution:** For a batch targeting component X, the
+controller looks for an entry in `targetConfig` where `target` equals X
+and `batchPolicy` is present. Fields set in `batchPolicy` take
+precedence over `batchDefaults`; omitted fields fall back to defaults.
+
+**Validation rules:**
+
```
kasemAlem, line 281: Validation rules are one-directional — removing edges can orphan targetConfig entries.
  
   Rule 2 prevents adding a targetConfig entry with no matching nudge edge. But there is no symmetric rule preventing the removal of the last nudge edge pointing to a target that still has a targetConfig entry. 
   A user who removes all operand → bundle edges without also removing bundle from targetConfig ends up with an orphaned batch policy — silently invalid config that the webhook accepted.
  
   Suggest adding: "Removing a nudge edge is rejected if it would leave a targetConfig entry with no matching to edges. Both the edge removal and the targetConfig removal must be submitted in the same update."

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,688 @@
...
+      to: standalone-consumer # immediate (no targetConfig)
+```
+
+**Fields:**
+
+| Field | Type | Description |
+|---|---|---|
+| `spec.batchDefaults.debounceTimeout` | `duration` | Time to wait after the last build event before firing the batch. Default: `30m`. Range: 1m--24h. |
+| `spec.batchDefaults.maxWaitTime` | `duration` | Hard deadline from the first build event. Fires the batch when reached regardless of debounce resets. Only applies during `Accumulating` phase -- does not override `Block` (a blocked batch stays blocked until the failure is resolved or the user triggers `forceFire`). Default: `4h`. Must exceed `debounceTimeout`. |
+| `spec.batchDefaults.failurePolicy` | `enum` | `Block`: batch transitions to `Blocked` if any member build fails; it stays blocked indefinitely (ignoring `maxWaitTime`) until the user fixes the failure or uses `forceFire`. `ProceedWithPartial`: batch fires with whatever succeeded when the debounce or hard deadline expires; if all components failed (empty `accumulated`), the batch transitions to `Failed` instead of `Firing` -- there is nothing to nudge. Default: `Block`. |
```
kasemAlem, line 255: Block is indefinite with no observable signal.
  
   The design choice to ignore `maxWaitTime` in Blocked phase is reasonable, but the ADR specifies no mechanism to alert users that a batch is stuck. The only visibility into a blocked batch is actively querying  `status.activeBatches.message`. In practice this means a blocked batch can silently stall nudging for hours or days with no alert.
  
   Suggest specifying: (1) a Kubernetes Event (Warning/BatchBlocked) emitted when the batch first enters Blocked, and (2) a second Event emitted if the batch remains Blocked past `maxWaitTime`, prompting the user to consider forceFire. 
This costs nothing architecturally and makes the escape hatch discoverable without requiring users to actively poll `NudgeConfig` status.

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,688 @@
...
+If a component builds multiple times during a batch (retry, new push),
+the latest result replaces the previous one in `accumulated`. Only the
+most recent digest per nudging component is included.
+
+#### Component deleted during a batch
+
+If a nudging component is deleted while a batch is accumulating, the
+stale reference controller (from ADR 67) marks the NudgeConfig condition.
+The batch continues with whatever has accumulated -- the deleted
+component simply never contributes a result.
```
kasemAlem: This only covers deletion before the component contributes a result. The inverse case is unspecified.
  
   If a component is deleted after its build result is already in accumulated, the stale reference controller sets Valid=False on the NudgeConfig while the batch still holds that component's image digest. The 
   ADR doesn't say whether the batch should fire in this state, skip the deleted component's entry, or wait until the condition clears. 
An implementer choosing Option B (block while Invalid) would silently stall
   batches whenever any source component is deleted mid-flight — with no documentation to explain why. Please specify the intended behaviour explicitly.

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,688 @@
...
+target in `status.activeBatches`. If none exists, a new batch is
+created. If one exists, the build result is appended and the debounce
+timer is reset.
+
+**Deduplication:** If the same `from` component builds multiple times
+during a batch (e.g., a retry), the latest result replaces the previous
+one. Only the most recent digest per nudging component is included in the
+aggregated nudge.
+
+### Aggregated Renovate PipelineRun
```
kasemAlem, line 426: Renovate configuration inheritance is not specified for batched PLRs.
  
   ADR-0067 defines two mechanisms that IS applies when creating a Renovate PLR: (1) the two-tier ConfigMap lookup (`namespace-wide-nudging-renovate-config` + per-target annotation override), and (2) the 
   `build-nudge-simple-branch` annotation on source components. ADR-0072 mentions a batch-specific branch prefix but says nothing about the `ConfigMap` mechanism.
  
   An implementer reading this ADR without knowing ADR-0067 in detail could easily omit the `ConfigMap` lookup entirely, silently bypassing all user-configured Renovate behaviour for batched nudges. Please add an 
   explicit statement that the ADR-0067 ConfigMap mechanism applies unchanged, and note that build-`nudge-simple-branch` on source components is superseded by the batch branch prefix.

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,688 @@
...
+(`type: ForceFireRejected`, `status: "True"`, `message` listing the
+failed components). The Condition is cleared on the next successful
+reconciliation. This lets users distinguish "fire now with partial
+results" from "fire now but only if everything succeeded" and provides
+clear feedback when the action cannot proceed.
+
+The controller processes the action, transitions the batch to `Firing`,
+creates the aggregated Renovate PipelineRun with whatever has
+accumulated, and clears the `actions.forceFire` field.
+
```
kasemAlem, line 513: forceFire cleanup is only specified for the success path.
  
   The ADR says the controller "clears the actions.forceFire field" after firing the batch. But if the controller crashes after creating the PLR and before clearing the field, or if the batch is already in 
   Firing when forceFire is read (the "ignored" case in the transition table), the field is never cleared. The transition table entry for forceFire in Firing state says "Ignored" but doesn't say the field is 
   also cleared.
  
   Please specify that the controller clears spec.actions.forceFire on every reconcile where it is set — whether the action fires, is ignored, or is rejected — so the field never persists across restarts in a 
   stale state.

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,688 @@
...
+| **Firing** | Recorded in **next batch** (create if needed → next `Accumulating`). This batch stays `Firing`. | Same -- recorded in **next batch**. This batch stays `Firing`. | n/a | Ignored (already firing). | → `Completed`. Next batch (if any) is promoted to current. | Retry up to 3x. After retries exhausted: → `Failed`. Next batch (if any) can now fire. |
+| **Failed** | Recorded in **next batch** (create if needed → next `Accumulating`). This batch stays `Failed`. | Same -- recorded in **next batch**. | n/a | → `Firing`. Create new PLR with same `accumulated`. | n/a | n/a |
+| **Completed** | **Next batch** created (→ next `Accumulating`). This batch cleaned up. | Same -- recorded in **next batch**. This batch cleaned up. | n/a | Ignored (nothing to fire). | n/a | n/a |
+
+The state machine was formally verified with TLA+ model checking. TLC
+explored 656,827 states with 3 components, 12 events, and both failure
+policies. All invariants hold: types are correct, `Firing` always has
+non-empty `accumulated`, and at most one batch is in `Firing` at any
+time.
```
kasemAlem: TLA+ specification is not linked.

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,688 @@
...
+    debounceTimeout: 30m        # quiet period before firing
+    maxWaitTime: 4h             # hard cap regardless of debounce resets
+    failurePolicy: Block        # Block | ProceedWithPartial
+
+  targetConfig:
+    - target: bundle
+      batchPolicy:
+        debounceTimeout: 15m
+        maxWaitTime: 2h
+        failurePolicy: ProceedWithPartial
```
kasemAlem, line 232: Post-partial-fire state of the target is not documented.
  
   When a batch fires with partial results, the target component ends up with a mix of updated and stale image references — which is correct behaviour, but the ADR doesn't say what happens next. Users seeing a 
   partial fire have no documented answer to "what about the failed components?" Please add a note explaining that failed components' results are picked up in a subsequent batch cycle when those components 
   rebuild, so the target converges to a fully updated state without manual intervention.

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,688 @@
...
+| **Blocked** | Add to `accumulated`. If the succeeding component was in `failed`: remove it. If `failed` is now empty: → `Accumulating`, reset debounce. Otherwise stay `Blocked`. | Add to `failed`. Stay `Blocked`. | Ignored. `maxWaitTime` does not override `Block`. | → `Firing`. Create aggregated PLR. If `accumulated` is empty: → `Failed`. | n/a | n/a |
+| **Firing** | Recorded in **next batch** (create if needed → next `Accumulating`). This batch stays `Firing`. | Same -- recorded in **next batch**. This batch stays `Firing`. | n/a | Ignored (already firing). | → `Completed`. Next batch (if any) is promoted to current. | Retry up to 3x. After retries exhausted: → `Failed`. Next batch (if any) can now fire. |
+| **Failed** | Recorded in **next batch** (create if needed → next `Accumulating`). This batch stays `Failed`. | Same -- recorded in **next batch**. | n/a | → `Firing`. Create new PLR with same `accumulated`. | n/a | n/a |
+| **Completed** | **Next batch** created (→ next `Accumulating`). This batch cleaned up. | Same -- recorded in **next batch**. This batch cleaned up. | n/a | Ignored (nothing to fire). | n/a | n/a |
```
kasemAlem, line 402: Silent ignore is inconsistent with the ForceFireRejected case.
  
   When forceFire is rejected due to includePartial: false, the controller emits a Warning Event and sets a transient Condition. When forceFire is set on a Completed batch, it is silently cleared with no 
   feedback. From the user's perspective both cases look identical — the field disappears — but in one case they get an explanation and in the other they don't. Please add a ForceFireIgnored Event (or similar) 
   for the Completed case to make the behaviour consistent and discoverable.

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
gbenhaim: Is this a suggestion or a question?

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
gbenhaim: Fixed in d654a3d — added a "Validated mode specifics" sub-section that answers all three questions: (1) digest source is the passing Snapshot, (2) shared gating groups add one entry per component from a single Snapshot event, (3) failure = failing Snapshot, not a failed build PLR.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
gbenhaim: Fixed in d654a3d — added implementation detail about the `fireAt` timestamp mechanism, `RequeueAfter` scheduling, and controller startup recovery behavior.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
gbenhaim: Fixed in d654a3d — added explicit write ordering requirement: NudgeConfig status must be persisted before the `component-nudge-processed` annotation is written to the PLR. Documented why the reverse ordering loses data.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
gbenhaim: Fixed in d654a3d — defined the batched PLR annotation schema in a table: `nudging-components` (JSON array), `nudging-digests` (JSON object), and `batch-id`. The singular annotations are explicitly not set on batched PLRs.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
gbenhaim: Fixed in d654a3d — expanded the section to cover both cases. After contributing: the entry stays in `accumulated`, the batch fires normally, and `Valid=False` is informational only (the digest was valid at capture time).

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
gbenhaim: Fixed in d654a3d — added explicit statement that the ADR-0067 two-tier ConfigMap mechanism applies unchanged to batched PLRs. Also noted that `build-nudge-simple-branch` is superseded by the batch branch prefix for batched targets.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
gbenhaim: Fixed in d654a3d — specified that the controller clears `spec.actions.forceFire` on every reconcile where it is set, whether the action fires, is ignored (already firing/completed), or is rejected. It never persists in a stale state across restarts.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
gbenhaim: ~~Addressed in d654a3d — added a note that the TLA+ specification will be published in a companion repository once the ADR is accepted and is available for review on request.~~

**Update (3e885a5):** Removed all TLA+ references from the ADR. The formal verification work is not ready to be published alongside the ADR.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
gbenhaim: Fixed in d654a3d — added a "Post-partial-fire convergence" section explaining that failed components are picked up in a subsequent batch cycle when they rebuild, so the target converges to a fully updated state without manual intervention.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
gbenhaim: ~~Fixed in d654a3d — specified that when forceFire is ignored or rejected, the controller emits a `Warning` Event (`ForceFireIgnored` or `ForceFireRejected`) explaining why the action had no effect.~~

**Update (76eb737):** We moved away from Kubernetes Events entirely — they are not surfaced in the Konflux UI, add etcd pressure, and offer no discoverability advantage over the CR status. Instead, the controller sets a transient Condition on the NudgeConfig status (`ForceFireIgnored` or `ForceFireRejected`) explaining why the action had no effect. The Condition is cleared on the next successful reconciliation.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
gbenhaim: Agree that blocked batches shouldn't silently stall — addressed in 76eb737 with a `BatchBlocked` Condition on the NudgeConfig status.

We opted for a Condition rather than Events because:
1. Kubernetes Events are not surfaced in the Konflux UI
2. Events add etcd write pressure (we're already working to reduce Event TTLs)
3. Users already need to consult NudgeConfig status for batch progress — a Condition there is strictly easier to discover than a separate Event

Added a "Why not Kubernetes Events?" section to make this rationale explicit.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
gbenhaim: Good observation — this made us reconsider both rules. We initially added rule 5 (symmetric removal rejection) to address this, but on reflection neither rule 2 nor rule 5 is worth keeping.

An orphaned `targetConfig` entry is inert: the controller never matches it, so it can't cause incorrect behavior. Enforcing strict symmetry just forces users to keep `targetConfig` and `nudges` in lockstep (atomic removal), adding operational friction for no safety benefit. Users may also want to pre-configure batch policies before wiring up edges.

Removed both rules in 8b689aa and added an explicit paragraph explaining why orphans are allowed.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,804 @@
...
+PLRs the schema is extended:
+
+| Annotation | Batched value |
+|---|---|
+| `build.appstudio.redhat.com/nudged-components` | Unchanged (target component name). |
+| `build.appstudio.redhat.com/nudging-components` | JSON array of source component names: `["operand-a","operand-b",...]`. Replaces the singular `nudging-component`. |
+| `build.appstudio.redhat.com/nudging-digests` | JSON object mapping component name to digest: `{"operand-a":"sha256:abc...","operand-b":"sha256:def..."}`. Replaces the singular `nudging-image`. |
+| `build.appstudio.redhat.com/batch-id` | `<target>-<createdAt-epoch>` (e.g., `bundle-1721126400`). Links the PLR to the batch for observability. |
+
+The singular annotations (`nudging-component`, `nudging-pipeline`,
```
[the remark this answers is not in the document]
kasemAlem: @gbenhaim 

  `nudging-pipeline` has no plural equivalent defined.

  The table introduces nudging-components and nudging-digests as plural replacements for nudging-component and nudging-image, but `nudging-pipeline` is silently dropped with no explanation. For a batch with N
  source builds there are N build PLRs — it's not obvious whether the intent is:
  1. A new nudging-pipelines JSON array listing all N build PLR names, or
  2. The annotation simply doesn't apply to batched PLRs (and consumers should not expect it)

  Please add one line clarifying which it is.

  Also — "Consumers must check for the plural form first" identifies a migration requirement but not who those consumers are. Integration-service, the UI, and any tests that read nudging-component today will
  silently get nothing on batched PLRs. Please enumerate the known consumers and confirm they will be updated before batched nudging ships.

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,808 @@
...
+
+**Validation rules:**
+
+1. No duplicate `target` values in `targetConfig`.
+2. `maxWaitTime` must be greater than `debounceTimeout` (checked on both
+   `batchDefaults` and each resolved policy with fallback).
+3. `debounceTimeout` must be between 1m and 24h.
+
+Orphaned `targetConfig` entries (those with no matching nudge edges) are
+explicitly allowed. An orphaned entry is inert — the controller never
```
kasemAlem, line 288: Agreed that orphans are inert and the pre-configure use case is valid. One small gap: a user who accidentally leaves a stale `targetConfig` entry after removing all edges gets no feedback — the entry is silently
  ignored forever.

  Suggest adding a status condition (e.g. `OrphanedTargetConfig`) listing any targetConfig targets that have no matching nudge edges. This costs nothing at runtime (the controller already iterates targetConfig
  during reconciliation) and makes the pre-configure use case self-documenting — the condition clears automatically once edges are wired up.

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,808 @@
...
+`Accumulating` and the debounce timer is reset. This allows
+self-healing without manual intervention when the underlying issue is
+fixed.
+
+#### Blocked batch observability
+
+When a batch transitions to `Blocked`, the controller sets a Condition
+on the NudgeConfig status (`type: BatchBlocked`, `status: "True"`,
+`message` listing the failed component(s) and the target). The Condition
+is cleared when the batch leaves the `Blocked` phase (e.g., via retry
```
kasemAlem, line 589: The Condition covers entry to Blocked but not duration. A batch blocked for 3 days looks identical to one that just entered — the Condition stays True with the same `lastTransitionTime` and message.

  Suggest: when the batch remains Blocked past `maxWaitTime`, update the Condition message to note how long it has been blocked and prompt the user to consider forceFire. This needs no new mechanism — just a
  message update on each reconciliation when now - `lastTransitionTime` > `maxWaitTime`.

gbenhaim: I think it's pre mature to decide about this behavior. We need real world data and feedback from users imo for deciding if such features are needed.

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
gbenhaim: Good suggestion — added in 6a23294. The controller now sets an informational `OrphanedTargetConfig` Condition listing any targetConfig entries with no matching nudge edges. It clears automatically once edges are added or the stale entry is removed.

— Gal's Cursor

kasemAlem (approved):

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,814 @@
...
+nudge PLR annotations (`nudged-components`, `nudging-component`,
+`nudging-pipeline`, `nudging-image`) are single-valued. For batched
+PLRs the schema is extended:
+
+| Annotation | Batched value |
+|---|---|
+| `build.appstudio.redhat.com/nudged-components` | Unchanged (target component name). |
+| `build.appstudio.redhat.com/nudging-components` | JSON array of source component names: `["operand-a","operand-b",...]`. Replaces the singular `nudging-component`. |
+| `build.appstudio.redhat.com/nudging-digests` | JSON object mapping component name to digest: `{"operand-a":"sha256:abc...","operand-b":"sha256:def..."}`. Replaces the singular `nudging-image`. |
+| `build.appstudio.redhat.com/batch-id` | `<target>-<createdAt-epoch>` (e.g., `bundle-1721126400`). Links the PLR to the batch for observability. |
```
dirgim, line 445: Should we have the `batch-id` field in the NudgeConfig's status entry for that batch to make it easier to look them up?

gbenhaim: Good idea — added in f410dbe. The status entry now has an explicit `batchId` field (`<target>-<createdAt-epoch>`) so consumers can correlate directly with the PLR annotation without constructing the string manually.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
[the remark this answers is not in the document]
gbenhaim: Clarified in 50b046e.

**On `nudging-pipeline`:** Intentionally dropped with no plural equivalent. For a batch with N source builds, the build PLR names are already in `status.activeBatches[].accumulated[].buildPipelineRun` — reachable via the `batch-id` annotation. Duplicating N PLR names into an annotation risks hitting annotation size limits on large batches and adds no information not already in NudgeConfig status.

**On enumerating consumers:** I think this is an implementation/migration concern rather than something for the ADR itself. The ADR establishes the annotation contract; the implementation PRs will track which consumers (UI, integration-service internals, tests) need updating before batched nudging ships.

— Gal's Cursor

dirgim (approved):
LGTM

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,822 @@
...
+successfully rebuild. The next batch starts fresh, collects the new
+successful builds, and fires a nudge that updates the remaining stale
+references. The target converges to a fully updated state without manual
+intervention, just one batch cycle later.
+
+#### PipelineRun pruning during a batch
+
+For immediate-mode nudging, integration-service processes build
+PipelineRuns on completion via its reconciliation loop and marks them
+with a `component-nudge-processed` annotation to prevent duplicate
```
kasemAlem:
```suggestion
with a `test.appstudio.openshift.io/component-nudge-processed` annotation 
(IS-owned key, separate from build-service's `build.appstudio.openshift.io/component-nudge-processed`) to prevent duplicate

```

@rcerven , wdyt?

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,822 @@
...
+processing (PR #1604). This relies on the controller processing the PLR
+before the Tekton pruner deletes it -- the same race window that exists
+for immediate-mode nudging today.
+
+The batched path uses the same mechanism: when a build PLR completes,
+integration-service captures the result into `activeBatches` status and
+marks the PLR with the processed annotation.
```
kasemAlem:
```suggestion
marks the PLR with the Integration-Service processed annotation.
(`test.appstudio.openshift.io/component-nudge-processed`)

```
@rcerven , and this?

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,822 @@
...
+
+**Write ordering:** The NudgeConfig status update must be persisted
+before the `component-nudge-processed` annotation is written to the PLR.
+This ordering ensures that a controller crash between the two writes is
+always recoverable -- the PLR will be re-processed on restart and the
+dedup-by-`from` rule prevents duplicate entries. The reverse ordering
+(annotate first, then status) would lose the build result permanently if
+the controller crashes between the two writes.
```
kasemAlem:
```suggestion
**Write ordering and PLR lifecycle:** Integration-service adds a
`test.appstudio.openshift.io/nudge-pipelinerun` finalizer to the build
PLR before capturing its result, preventing the Tekton pruner from
deleting it before processing is complete. The NudgeConfig status update
must be persisted before the
`test.appstudio.openshift.io/component-nudge-processed` annotation is
written to the PLR, and the finalizer is removed only after both writes
succeed. This ordering ensures that a controller crash between any two
marks them with a `test.appstudio.openshift.io/component-nudge-processed` annotation
(IS-owned key, separate from build-service's `build.appstudio.openshift.io/component-nudge-processed`) to prevent duplicate processing (PR #1604).
```
@rcerven , wdyt?

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
rcerven: looks good

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
rcerven: looks good

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
rcerven: looks good

rcerven (approved):

Omeramsc (approved):
LGTM

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,814 @@
...
+nudging component's image reference to its new digest.
+
+**Annotation schema for batched PLRs:** The existing single-component
+nudge PLR annotations (`nudged-components`, `nudging-component`,
+`nudging-pipeline`, `nudging-image`) are single-valued. For batched
+PLRs the schema is extended:
+
+| Annotation | Batched value |
+|---|---|
+| `build.appstudio.redhat.com/nudged-components` | Unchanged (target component name). |
```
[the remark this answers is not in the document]
gbenhaim: Fixed in 93ca79a — updated the three new annotations to use `build.konflux-ci.dev/` prefix per ADR 71.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,814 @@
...
+The aggregated Renovate PipelineRun name is deterministic, derived from
+the target name and batch creation timestamp (e.g.,
+`nudge-batch-bundle-1721126400`). This makes the Firing transition
+idempotent: if the controller crashes after creating the PLR but before
+writing `nudgePipelineRun` to status, the restart attempts to create the
+same PLR name, receives an `AlreadyExists` response, and proceeds to
+update the status.
+
+#### Aggregated nudge PipelineRun failure
+
```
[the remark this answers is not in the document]
gbenhaim: Fixed in 93ca79a — clarified that the batch remains in `Firing` during retries and only transitions to `Failed` after all retries are exhausted. Also specified that the retry counter is stored in `status.activeBatches[].retryCount`.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,814 @@
...
+| `spec.batchDefaults.maxWaitTime` | `duration` | Hard deadline from the first build event. Fires the batch when reached regardless of debounce resets. Only applies during `Accumulating` phase -- does not override `Block` (a blocked batch stays blocked until the failure is resolved or the user triggers `forceFire`). Default: `4h`. Must exceed `debounceTimeout`. |
+| `spec.batchDefaults.failurePolicy` | `enum` | `Block`: batch transitions to `Blocked` if any member build fails; it stays blocked indefinitely (ignoring `maxWaitTime`) until the user fixes the failure or uses `forceFire`. `ProceedWithPartial`: batch fires with whatever succeeded when the debounce or hard deadline expires; if all components failed (empty `accumulated`), the batch transitions to `Failed` instead of `Firing` -- there is nothing to nudge. Default: `Block`. |
+| `spec.targetConfig` | `[]TargetConfig` | Per-target configuration. A target with a `batchPolicy` receives batched nudges; targets not listed here (or listed without `batchPolicy`) receive immediate nudges. |
+| `spec.targetConfig[].target` | `string` | Name of the nudged (target) component. Must match the `to` value of at least one nudge edge. |
```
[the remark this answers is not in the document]
gbenhaim: Fixed in 93ca79a — updated the field table description for `spec.targetConfig[].target` to note that orphaned entries are allowed but trigger an informational `OrphanedTargetConfig` Condition.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,822 @@
...
+| `spec.targetConfig[].batchPolicy` | `object` | Batch policy for this target. Presence of this field opts the target into batched mode. Fields omitted from `batchPolicy` fall back to `batchDefaults`. An empty object (`{}`) means "use all defaults." |
+| `spec.targetConfig[].batchPolicy.debounceTimeout` | `duration` | Overrides `batchDefaults.debounceTimeout` for this target. |
+| `spec.targetConfig[].batchPolicy.maxWaitTime` | `duration` | Overrides `batchDefaults.maxWaitTime` for this target. |
+| `spec.targetConfig[].batchPolicy.failurePolicy` | `enum` | Overrides `batchDefaults.failurePolicy` for this target. |
+| `spec.actions.forceFire` | `object` | One-shot action field. When set, the controller force-fires the batch for the specified target regardless of debounce state. Cleared after processing. |
+| `spec.actions.forceFire.target` | `string` | Required. Name of the target component whose batch should be force-fired. Must match a target in `activeBatches`. |
```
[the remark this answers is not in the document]
gbenhaim: Fixed in 93ca79a — added an explicit dispatch rule to the `forceFire.target` field description: when two batches coexist, forceFire applies to the batch that can act on it (the accepting batch in `Accumulating`/`Blocked`, or a `Failed` batch). Batches in `Firing` or `Completed` ignore the action.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,822 @@
...
+    - from: standalone-lib
+      to: standalone-consumer # immediate (no targetConfig)
+```
+
+**Fields:**
+
+| Field | Type | Description |
+|---|---|---|
+| `spec.batchDefaults.debounceTimeout` | `duration` | Time to wait after the last build event before firing the batch. Default: `30m`. Range: 1m--24h. |
+| `spec.batchDefaults.maxWaitTime` | `duration` | Hard deadline from the first build event. Fires the batch when reached regardless of debounce resets. Only applies during `Accumulating` phase -- does not override `Block` (a blocked batch stays blocked until the failure is resolved or the user triggers `forceFire`). Default: `4h`. Must exceed `debounceTimeout`. |
```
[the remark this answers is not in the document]
gbenhaim: Fixed in 93ca79a — added to the Blocked→Accumulating transition in the table: if `hardDeadline` has already passed when the batch unblocks, it fires immediately instead of resetting the debounce.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,829 @@
...
+before the Tekton pruner deletes it -- the same race window that exists
+for immediate-mode nudging today.
+
+The batched path uses the same mechanism: when a build PLR completes,
+integration-service captures the result into `activeBatches` status and
+marks the PLR with the Integration-Service processed annotation.
+(`test.appstudio.openshift.io/component-nudge-processed`)
+
+
+**Write ordering and PLR lifecycle:** Integration-service adds a
```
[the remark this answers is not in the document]
gbenhaim: Fixed in 93ca79a — removed the duplicated text and completed the truncated sentence about crash-safety: on restart the controller re-reads the NudgeConfig status and re-processes any PLR whose finalizer is still present but whose result is not yet in `activeBatches`.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,829 @@
...
+Users need an escape hatch for batches that are stuck or that they want
+to fire early. Following the action pattern from the
+[revised component model](0056-revised-component-model.md):
+
+```yaml
+spec:
+  actions:
+    forceFire:
+      target: bundle
+      includePartial: true    # proceed even if some builds failed
```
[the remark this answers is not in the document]
gbenhaim: Fixed in 93ca79a — added to the Blocked→Accumulating transition: if `hardDeadline` has already passed, the batch fires immediately instead of resetting the debounce timer.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,814 @@
...
+
+The singular annotations (`nudging-component`, `nudging-pipeline`,
+`nudging-image`) are not set on batched PLRs to avoid ambiguity.
+Consumers must check for the plural form first.
+
+**Renovate configuration inheritance:** The two-tier ConfigMap lookup
+from ADR 67 (`namespace-wide-nudging-renovate-config` + per-target
+annotation override) applies unchanged to batched PLRs. The controller
+resolves the Renovate configuration for the target component using the
+same mechanism as single-component nudges. The
```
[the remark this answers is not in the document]
gbenhaim: Acknowledged — this is already documented in the ADR (the Renovate configuration inheritance section notes that `build-nudge-simple-branch` is superseded by the batch branch prefix). No additional change needed.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,814 @@
...
+idempotent: if the controller crashes after creating the PLR but before
+writing `nudgePipelineRun` to status, the restart attempts to create the
+same PLR name, receives an `AlreadyExists` response, and proceeds to
+update the status.
+
+#### Aggregated nudge PipelineRun failure
+
+If the aggregated Renovate PipelineRun fails, the batch transitions to
+`Failed`. The controller retries by creating a new PipelineRun with an
+incremented attempt suffix (e.g., `nudge-batch-bundle-1721126400-retry1`)
```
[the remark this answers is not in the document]
gbenhaim: Addressed in 93ca79a — the retry counter is now specified as stored in `status.activeBatches[].retryCount`, which provides the deterministic naming derivation for crash recovery.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,822 @@
...
+| `Completed` | Renovate PLR succeeded. Retained for observability. Cleaned up when next batch starts or after 24h. |
+
+**Transition table** -- every (state, event) pair. "This batch" refers
+to the current batch; "next batch" to the second batch slot created
+when builds arrive during `Firing`/`Failed`/`Completed`.
+
+| State | Build OK | Build fail | Timer expires | forceFire | PLR OK | PLR fail |
+|---|---|---|---|---|---|---|
+| **Accumulating** | Append to `accumulated` (dedup by `from`), reset debounce. Stay `Accumulating`. | If `Block`: → `Blocked`. If `ProceedWithPartial`: record in `failed`, stay `Accumulating`. | → `Firing`. Create aggregated PLR. If `accumulated` is empty: → `Failed`. | → `Firing`. Create aggregated PLR. If `accumulated` is empty: → `Failed`. | n/a | n/a |
+| **Blocked** | Add to `accumulated`. If the succeeding component was in `failed`: remove it. If `failed` is now empty: → `Accumulating`, reset debounce. Otherwise stay `Blocked`. | Add to `failed`. Stay `Blocked`. | Ignored. `maxWaitTime` does not override `Block`. | → `Firing`. Create aggregated PLR. If `accumulated` is empty: → `Failed`. | n/a | n/a |
```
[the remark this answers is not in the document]
gbenhaim: The transition table is already complex; the Manual Override section documents the `includePartial: false` rejection path in detail. Adding a footnote to the table would hurt readability without adding information not already in the document.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,822 @@
...
+    Accumulating --> Accumulating : +build
+    Accumulating --> Blocked : failure [Block]
+    Accumulating --> Firing : timer / forceFire
+    Blocked --> Accumulating : fix
+    Blocked --> Firing : forceFire
+    Firing --> Completed : PLR ok
+    Firing --> Failed : PLR fail
+    Failed --> Firing : forceFire
+    Failed --> [*] : superseded
+    Completed --> [*] : cleanup
```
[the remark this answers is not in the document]
gbenhaim: The Mermaid diagram is intentionally simplified — the transition table directly below has the full compound conditions. Expanding diagram labels to include the full logic would make it unreadable.

— Gal's Cursor

ADR/AGENTS.md
```diff
@@ -5,6 +5,7 @@ Status: Ac=Accepted Im=Implemented Ib=Implementable Pr=Proposed Ap=Approved Rp=R
 0048 Ac 92L Attestable Build-Time Tests in Integration Service
 0040 Ac 142L Availability Probe Framework
 0059 Ib 66L Backend Usage Telemetry Collection from Konflux Clusters
+0072 Pr 804L Batched Nudging for Many-to-One Component Dependencies
```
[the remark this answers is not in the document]
gbenhaim: Fixed in 93ca79a — AGENTS.md line count updated.

— Gal's Cursor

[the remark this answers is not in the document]
gbenhaim: Fixed in 93ca79a — AGENTS.md regenerated with correct line count.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,829 @@
...
+on the NudgeConfig status (`type: BatchBlocked`, `status: "True"`,
+`message` listing the failed component(s) and the target). The Condition
+is cleared when the batch leaves the `Blocked` phase (e.g., via retry
+success or `forceFire`). This is the primary observability mechanism —
+users (and the UI) inspect the NudgeConfig status directly, which is
+already the authoritative source for batch progress.
+
+#### Why not Kubernetes Events?
+
+We deliberately avoid Kubernetes Events for lifecycle signalling:
```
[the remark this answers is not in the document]
gbenhaim: The transition table already covers this case: forceFire with empty `accumulated` transitions to `Failed`. The prose for the batchPolicy-removed edge case describes the general behavior; the transition table is authoritative for the exact state machine semantics.

— Gal's Cursor

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,804 @@
...
+| `spec.targetConfig[].target` | `string` | Name of the nudged (target) component. Must match the `to` value of at least one nudge edge. |
+| `spec.targetConfig[].batchPolicy` | `object` | Batch policy for this target. Presence of this field opts the target into batched mode. Fields omitted from `batchPolicy` fall back to `batchDefaults`. An empty object (`{}`) means "use all defaults." |
+| `spec.targetConfig[].batchPolicy.debounceTimeout` | `duration` | Overrides `batchDefaults.debounceTimeout` for this target. |
+| `spec.targetConfig[].batchPolicy.maxWaitTime` | `duration` | Overrides `batchDefaults.maxWaitTime` for this target. |
+| `spec.targetConfig[].batchPolicy.failurePolicy` | `enum` | Overrides `batchDefaults.failurePolicy` for this target. |
+| `spec.actions.forceFire` | `object` | One-shot action field. When set, the controller force-fires the batch for the specified target regardless of debounce state. Cleared after processing. |
+| `spec.actions.forceFire.target` | `string` | Required. Name of the target component whose batch should be force-fired. Must match a target in `activeBatches`. |
```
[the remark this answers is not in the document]
gbenhaim: The field description already states the target must match `activeBatches`. If no batch exists for the target, the field is simply ignored (same as `forceFire` on a `Completed` batch). The controller clears the field on every reconcile regardless.

— Gal's Cursor

gbenhaim:
@konflux-ci/kgc can you please review?

ralphbean:
@gbenhaim I think we were waiting for an announcement about this one, either on the community call or on the mailing list (per [process](https://github.com/konflux-ci/community/blob/main/ADRs.md#process)). I see there a lots of approvals on this already so clearly people are aware. Did @konflux-ci/kgc miss the request for comments announcement?

ralphbean (approved):
This LGTM. The outstanding findings from the review agent look right, but are nit-picky. They should be resolved before merge but are not essential.

From kgc meeting today, we just want to make sure that this got announced either on the list or on the community call to give the best chance for people to be aware of it before it gets merged.

Adding my approval now on good faith.

gbenhaim:
> @gbenhaim I think we were waiting for an announcement about this one, either on the community call or on the mailing list (per [process](https://github.com/konflux-ci/community/blob/main/ADRs.md#process)). I see there a lots of approvals on this already so clearly people are aware. Did @konflux-ci/kgc miss the request for comments announcement?

Thanks @ralphbean. I've sent now an email about the ADR to the mailing list.

ADR/0072-batched-nudging-for-many-to-one-dependencies.md
```diff
@@ -0,0 +1,832 @@
...
+- **Added complexity in integration-service.** The batch state machine
+  (5 phases, 2-batch concurrency model, debounce timers, retry logic)
+  adds significant controller complexity compared to the current
+  fire-and-forget nudge path.
+
+- **Delayed nudges by design.** Batching introduces intentional latency
+  (debounce timeout + maxWaitTime). For urgent single-component updates,
+  users must either not use batching for that target or use `forceFire`.
+
+- **Pruner race window.** Build results must be captured before the
```
mftb, line 785: Is there a way to eliminate this race condition? What are the hazards if the completed PipelineRuns are deleted before the build results are captured?

ralphbean (approved):

PR #376, opened 2026-08-13
Add ADR (0073): Konflux README status badges

Shiri-sh:
Problem: Konflux reports status via GitHub Checks, but there’s no badge URL teams can embed in READMEs.

Solution: A read-only Konflux Badge Service that queries Tekton results and returns SVG badges, starting with build and EC status.

### DIFF
--- a/ADR/0073-readme-status-badges.md
+++ b/ADR/0073-readme-status-badges.md
@@ -28,26 +28,15 @@ Today, users must open the Konflux UI (SSO) or the GitHub Checks popup to see st
 
 Beyond README badges, the same service can power **monitoring dashboards**, **portfolio status pages**, and **third-party integrations** (Slack bots, Jira, Confluence) via a JSON endpoint.
 
-### Spike findings
-
-We evaluated options and measured real repositories (`konflux-ci/konflux-ci`, `konflux-ci/segment-bridge`).
-
-| Approach | Tested | Verdict | Why |
-|---|---|---|---|
-| Shields.io `github/check-runs` | Live URLs + [source](https://github.com/badges/shields/blob/master/services/github/github-check-runs.service.js) | **Rejected** | Reads only first ~30 check-run rows. Busy repos have 50–70+; Konflux `*-on-push` often at row 31+ → `no check runs`. Third-party SaaS; weak for private repos. |
-| GitHub Action → commit `badge/*.svg` | Prototype workflow | **Rejected** | Mixes CI status into source; per-repo maintenance; org-wide pattern needs ADR anyway. |
-| Tekton task → git push SVG | Design review | **Rejected** | Same concerns as git-commit approach. |
-| Self-hosted Shields (`check-runs`) | Code review | **Rejected** | Same 30-row limit unless forked. Self-host helps RH control, not pagination. |
-| **Spike CLI: paginated Checks → SVG** | `go run` against segment-bridge | **Works** | Proves data + render logic; basis for service design. |
-
 ### Constraints
 
-- README is rendered by GitHub — badge URL must be **reachable without user SSO** (public HTTPS or designed private-repo auth).
+- README is rendered by GitHub — badge URL must be **reachable over public HTTPS**.
 - Konflux primary API is **Kubernetes CRDs** ([architecture overview](https://konflux-ci.dev/architecture/)); user-facing HTTP endpoints are exceptional but have established precedent (Tekton Results ingress per [Pipeline Service](../architecture/core/pipeline-service.md) and [ADR-0009](0009-pipeline-service-via-operator.md), PaC webhook receiver per [ADR-0039](0039-workspace-deprecation.md), registration-service). The [architecture overview](../architecture/index.md) permits these exceptions when justified.
 - Check runs are the status entries visible on each commit's **Checks tab** — the `-on-push` pipelines and `enterprise-contract` checks that PaC reports back to GitHub.
 - Check names are **per-component** (`*-on-push`, `*enterprise-contract*`), not one org-wide GitHub check name.
 - **Ring / release** status is not fully on GitHub Checks — requires cluster data (Tekton Results, Release CRs) later.
 - Konflux runs on **multiple production clusters**, and tenants from the same GitHub organization may be spread across different clusters.
+- Konflux onboarding data (Component CRs, tenant config) records **git URLs** but not GitHub repository visibility (public vs private). The proportion of private onboarded repos is unknown without querying the GitHub API per repository.
 
 ## Decision
 
@@ -58,6 +47,32 @@ Introduce a **Konflux Badge Service** — a new **read-only HTTP service** (host
 3. Uses the **paginated GitHub Checks API** initially for **build** and **EC** status on a branch HEAD.
 4. Extends later to **Tekton Results / Konflux CRs** for Ring and richer status, within the same codebase.
 5. Uses a **server-side GitHub credential** (GitHub App installation token stored in a Kubernetes Secret).
+
+### Data flow
+
+GitHub README images are fetched by GitHub's [camo proxy](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/about-anonymized-urls) — a plain HTTPS GET with **no viewer login** passed to the badge service. The badge service calls the GitHub API using the **PaC GitHub App** installation token (same App/credentials PaC uses to report check runs; stored in a cluster Secret).
+
+```mermaid
+sequenceDiagram
+    participant Viewer
+    participant GH as GitHub (README + camo)
+    participant Badge as Badge Service
+    participant GHAPI as GitHub API
+
+    Viewer->>GH: Open README
+    GH->>Badge: GET /badge/{org}/{repo}/{branch}/build
+    Badge->>GHAPI: GET /repos/{org}/{repo} (App token)
+    GHAPI-->>Badge: private: true/false
+    alt public repo
+        Badge->>GHAPI: GET check-runs
+        Badge-->>GH: SVG (pass/fail/running)
+    else private repo
+        Badge-->>GH: SVG (neutral — no status)
+    end
+    GH-->>Viewer: Render badge
+```
+
+**Public vs private (one URL, different response):** the service reads `repo.private` from the GitHub API. **Public** → fetch check runs, return real status SVG/JSON. **Private** → unauthenticated SVG returns a neutral badge only (no pass/fail); real status stays on GitHub Checks (for users with repo access). Returning real status for private repos on an open URL would let anyone who knows `org/repo/branch` learn build health without repo access.
 
 ### API shape
 
@@ -149,15 +164,14 @@ This decision is deferred because it depends on which cluster data sources are n
 
 ## Alternatives Considered
 
-| Alternative | Verdict | Rationale |
-|---|---|---|
-| Shields.io SaaS only | **Rejected** | Low effort but fails at scale (pagination, third party, private repos). |
-| Per-repo GitHub Action publisher | **Rejected** | Technically viable; CI status in git plus N workflows to maintain is not org-scalable. |
-| Konflux UI deep-link only | **Insufficient** | Meets "see status" but not embeddable badge — does not satisfy stated product goal. |
-| Self-hosted Shields (`check-runs`) | **Rejected** | Does not fix pagination without a fork. Acceptable only as renderer (`endpoint?url=<our JSON>`); direct SVG is simpler. |
-| Deploy per-cluster from the start | **Deferred** | Unnecessary for GitHub-only data path. Adds deployment complexity without benefit until the service reads cluster-internal data. |
-| Extend Tekton Results with a badge endpoint | **Rejected** | Tekton Results is an upstream project (`tektoncd/results`) — adding Konflux-specific badge logic there is out of scope. It also uses Bearer token auth, incompatible with GitHub's anonymous image fetching. |
-| JSON-only service + self-hosted Shields for SVG | **Rejected** | Requires deploying two services. Since `badge-maker` handles SVG rendering trivially, a single service serving both JSON and SVG is simpler to operate. |
+| Alternative | Evaluated | Verdict | Rationale |
+|---|---|---|---|
+| Shields.io SaaS `github/check-runs` | Live URLs on `konflux-ci/konflux-ci`, `konflux-ci/segment-bridge`; [source review](https://github.com/badges/shields/blob/master/services/github/github-check-runs.service.js) | **Rejected** | Reads only first ~30 check-run rows. Busy repos have 50–70+; Konflux `*-on-push` often at row 31+ → `no check runs`. Third-party SaaS; weak for private repos. |
+| GitHub Action → commit `badge/*.svg` to repo | Prototype workflow | **Rejected** | Mixes CI status into source; per-repo maintenance; |
+| Tekton task → git push SVG | Design review | **Rejected** | Same concerns as git-commit approach. |
+| Deploy per-cluster from the start | Design review | **Deferred** | Unnecessary for GitHub-only data path. Adds deployment complexity without benefit until the service reads cluster-internal data. |
+| Extend Tekton Results with a badge endpoint | Design review | **Rejected** | Tekton Results is an upstream project (`tektoncd/results`) — adding Konflux-specific badge logic there is out of scope. Bearer token auth incompatible with GitHub's anonymous image fetching. Badge service reads GitHub Checks (reported by PaC), not raw PipelineRun results. |
+| JSON-only service + self-hosted Shields for SVG | Design review | **Rejected** | Requires deploying two services. `badge-maker` renders SVG trivially; single service serving both JSON and SVG is simpler to operate. |
 
 ## Consequences
 
--- a/ADR/0073-readme-status-badges.md
+++ b/ADR/0073-readme-status-badges.md
@@ -1,0 +1,193 @@
+---
+title: "73. README Status Badges via Konflux Badge Service"
+status: Proposed
+applies_to:
+  - pipeline-service
+topics:
+  - github-integration
+  - badges
+  - external-api
+  - observability
+---
+
+# 73. README Status Badges via Konflux Badge Service
+
+Date: 2026-08-13
+
+## Status
+
+Proposed
+
+## Context
+
+### Problem
+
+Teams want **embeddable build status** on the default branch README (Jenkins-style pass / fail / running). Konflux reports pipeline status as **GitHub Checks** (Checks tab), not GitHub Actions. There is no official GitHub URL that renders a Checks badge image.
+
+Today, users must open the Konflux UI (SSO) or the GitHub Checks popup to see status. We want a **single dynamic URL** per repo that README can embed for **build**, **Enterprise Contract (EC)**, and later **Ring / release** data.
+
+Beyond README badges, the same service can power **monitoring dashboards**, **portfolio status pages**, and **third-party integrations** (Slack bots, Jira, Confluence) via a JSON endpoint.
+
+### Spike findings
+
+We evaluated options and measured real repositories (`konflux-ci/konflux-ci`, `konflux-ci/segment-bridge`).
+
+| Approach | Tested | Verdict | Why |
+|---|---|---|---|
+| Shields.io `github/check-runs` | Live URLs + [source](https://github.com/badges/shields/blob/master/services/github/github-check-runs.service.js) | **Rejected** | Reads only first ~30 check-run rows. Busy repos have 50–70+; Konflux `*-on-push` often at row 31+ → `no check runs`. Third-party SaaS; weak for private repos. |
+| GitHub Action → commit `badge/*.svg` | Prototype workflow | **Rejected** | Mixes CI status into source; per-repo maintenance; org-wide pattern needs ADR anyway. |
+| Tekton task → git push SVG | Design review | **Rejected** | Same concerns as git-commit approach. |
+| Self-hosted Shields (`check-runs`) | Code review | **Rejected** | Same 30-row limit unless forked. Self-host helps RH control, not pagination. |
+| **Spike CLI: paginated Checks → SVG** | `go run` against segment-bridge | **Works** | Proves data + render logic; basis for service design. |
+
+### Constraints
+
+- README is rendered by GitHub — badge URL must be **reachable without user SSO** (public HTTPS or designed private-repo auth).
+- Konflux primary API is **Kubernetes CRDs** ([architecture overview](https://konflux-ci.dev/architecture/)); user-facing HTTP endpoints are exceptional but have established precedent (Tekton Results ingress per [Pipeline Service](../architecture/core/pipeline-service.md) and [ADR-0009](0009-pipeline-service-via-operator.md), PaC webhook receiver per [ADR-0039](0039-workspace-deprecation.md), registration-service). The [architecture overview](../architecture/index.md) permits these exceptions when justified.
+- Check runs are the status entries visible on each commit's **Checks tab** — the `-on-push` pipelines and `enterprise-contract` checks that PaC reports back to GitHub.
+- Check names are **per-component** (`*-on-push`, `*enterprise-contract*`), not one org-wide GitHub check name.
+- **Ring / release** status is not fully on GitHub Checks — requires cluster data (Tekton Results, Release CRs) later.
+- Konflux runs on **multiple production clusters**, and tenants from the same GitHub organization may be spread across different clusters.
+
+## Decision
+
+Introduce a **Konflux Badge Service** — a new **read-only HTTP service** (hosted in its own repository, e.g. `konflux-ci/badge-service`) that:
+
+1. Exposes **dynamic badge URLs** for onboarded repositories.
+2. Returns **SVG badge images** for README embedding and **JSON** for programmatic consumers.
+3. Uses the **paginated GitHub Checks API** initially for **build** and **EC** status on a branch HEAD.
+4. Extends later to **Tekton Results / Konflux CRs** for Ring and richer status, within the same codebase.
+5. Uses a **server-side GitHub credential** (GitHub App installation token stored in a Kubernetes Secret).
+
+### API shape
+
+```
+GET /badge/{githubOrg}/{githubRepo}/{branch}/build      → image/svg+xml
+GET /badge/{githubOrg}/{githubRepo}/{branch}/ec         → image/svg+xml
+GET /badge/{githubOrg}/{githubRepo}/{branch}/build.json → application/json
+GET /badge/{githubOrg}/{githubRepo}/{branch}/ec.json    → application/json
+```
+
+The **SVG endpoints** return badge images for README embedding. The **JSON endpoints** return structured status data for programmatic consumers:
+
+```json
+{
+  "org": "konflux-ci",
+  "repo": "build-service",
+  "branch": "main",
+  "type": "build",
+  "status": "passing",
+  "sha": "abc123...",
+  "updated": "2026-08-12T10:30:00Z"
+}
+```
+
+- **Data path**: resolve branch HEAD SHA → paginate `GET /repos/{org}/{repo}/commits/{sha}/check-runs` → aggregate `*-on-push` / `enterprise-contract` → pass | fail | running.
+- **Rendering**: `badge-maker` library ([MIT OR Apache-2.0](https://github.com/badges/shields/tree/master/badge-maker)) or equivalent minimal SVG — **not** shields.io SaaS `check-runs` route.
+- **No tenant resolution needed initially**: The data source is the GitHub API, not cluster-internal data. The service only needs the GitHub org, repo, and branch to query check runs.
+
+### README usage
+
+```markdown
+![Konflux build](https://<badge-host>/badge/{org}/{repo}/main/build)
+```
+
+### Security and auth model
+
+**Data exposure**: The endpoints expose **only aggregated pass / fail / running status**. No build logs, error messages, pipeline details, or secrets are returned. For public repositories, this status is already visible on GitHub's Checks tab without authentication, so the badge service does not increase the exposure surface. The badge endpoint is intentionally **unauthenticated** because GitHub's markdown renderer fetches badge images as an anonymous HTTP client. The architecture guidance to use `SubjectAccessReview` for exceptional HTTP endpoints does not apply here; that mechanism is for authenticated cluster API access. Phase 3 cluster data queries will use SAR-guarded access when reading Tekton Results or Release CRs.
+
+**GitHub credential**: The service authenticates to the GitHub API using the **existing PaC GitHub App** on its hosting cluster ([Pipeline Service docs](../architecture/core/pipeline-service.md), [ADR-0039](0039-workspace-deprecation.md), [ADR-0009](0009-pipeline-service-via-operator.md)). Each Konflux cluster registers a separate GitHub App ([ADR-0039](0039-workspace-deprecation.md)); the badge service reads that cluster's App credentials (stored as a Kubernetes Secret). That App already has `checks:read` permission. No new GitHub App installation is required.
+
+**Public endpoint protection**:
+
+- **Rate limiting**: per-IP rate limiting (e.g. 60 requests/minute per IP) at the ingress level.
+- **No secrets in URLs**: badge URLs contain only the GitHub org, repo, and branch — information that is already public for public repositories.
+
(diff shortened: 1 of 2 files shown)


ADR/0073-readme-status-badges.md
```diff
@@ -0,0 +1,193 @@
...
+
+GitHub App installation tokens have a rate limit of [5,000 requests/hour](https://docs.github.com/en/apps/creating-github-apps/registering-a-github-app/rate-limits-for-github-apps) per installation. Without caching, a popular repo's badge could exhaust this budget quickly.
+
+**Caching strategy**:
+
+- The service caches the computed badge result per `{org}/{repo}/{branch}/{type}` with a **TTL of 30–60 seconds**. Badge status may lag real-time by up to one minute, which is acceptable for README display.
+- Cache invalidation is TTL-based initially. A future enhancement could use GitHub webhooks (`check_suite` events) for near-real-time invalidation.
+- SVG responses include `Cache-Control: no-cache, no-store` headers to prevent GitHub's [camo proxy](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/about-anonymized-urls) from serving stale images.
+
+**Rate budget**: With a 30-second cache TTL, each unique key generates at most 2 GitHub API calls per minute. An installation serving 100 repos × 2 badge types = 400 calls/minute, well within the 5,000/hour limit.
```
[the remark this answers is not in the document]
Shiri-sh: fixed

ADR/0073-readme-status-badges.md
```diff
@@ -0,0 +1,193 @@
...
+| GitHub Action → commit `badge/*.svg` | Prototype workflow | **Rejected** | Mixes CI status into source; per-repo maintenance; org-wide pattern needs ADR anyway. |
+| Tekton task → git push SVG | Design review | **Rejected** | Same concerns as git-commit approach. |
+| Self-hosted Shields (`check-runs`) | Code review | **Rejected** | Same 30-row limit unless forked. Self-host helps RH control, not pagination. |
+| **Spike CLI: paginated Checks → SVG** | `go run` against segment-bridge | **Works** | Proves data + render logic; basis for service design. |
+
+### Constraints
+
+- README is rendered by GitHub — badge URL must be **reachable without user SSO** (public HTTPS or designed private-repo auth).
+- Konflux primary API is **Kubernetes CRDs** ([architecture overview](https://konflux-ci.dev/architecture/)); user-facing HTTP endpoints are exceptional but have established precedent (Tekton Results ingress per [Pipeline Service](../architecture/core/pipeline-service.md), PaC webhook receiver, registration-service). The [architecture overview](../architecture/index.md) permits these exceptions when justified.
```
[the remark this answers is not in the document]
Shiri-sh: added

ADR/0073-readme-status-badges.md
```diff
@@ -0,0 +1,195 @@
...
+  - badges
+  - external-api
+  - observability
+---
+
+# 73. README Status Badges via Konflux Badge Service
+
+Date: 2026-08-12
+
+Related: [KFLUXSE-410](https://redhat.atlassian.net/browse/KFLUXSE-410)
```
cit1zen: Let's not mention RH JIRA in the upstream ADR.

Shiri-sh: removed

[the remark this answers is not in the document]
Shiri-sh: removed

ADR/0073-readme-status-badges.md
```diff
@@ -0,0 +1,193 @@
...
+
+**Public endpoint protection**:
+
+- **Rate limiting**: per-IP rate limiting (e.g. 60 requests/minute per IP) at the ingress level.
+- **No secrets in URLs**: badge URLs contain only the GitHub org, repo, and branch — information that is already public for public repositories.
+
+**Private repositories**: GitHub's image renderer cannot fetch badges for private repos (it makes an anonymous GET). Initially, private repo badges return a neutral "private" shield or a 404. Users of private repos can use the `.json` endpoint from authenticated contexts instead. A future signed-URL-with-TTL scheme could address this but requires further design.
```
Omeramsc: Can you elaborate? will the service won't be available to private repositories on phase 1? are we sure we can expand to them later?

ADR/0073-readme-status-badges.md
```diff
@@ -0,0 +1,193 @@
...
+1. Promote the JSON endpoint for dashboard and monitoring use cases.
+2. Consider webhook-based cache invalidation for near-real-time badge updates.
+
+**Phase 3 — Cluster data sources**:
+1. Extend to read Tekton Results for richer status (pipeline duration, last N runs).
+2. Extend to read Release CRs for ring/release badges.
+3. Evolve the deployment model to support cluster-internal data access (see Deployment model section).
+4. Introduce tenant resolution (org/repo → cluster/namespace mapping) for cluster queries.
+
+## Alternatives Considered
```
Omeramsc, line 167: Alternatives for what? as we already had an alternatives table above

ADR/0073-readme-status-badges.md
```diff
@@ -0,0 +1,193 @@
...
+- **Broader utility**: the JSON endpoint enables dashboards, Slack integrations, and portfolio views.
+- **Simple initial deployment**: single instance on one cluster, reusing PaC's existing GitHub App credentials.
+- **Follows established precedent**: public HTTP endpoints exist in Konflux (Tekton Results, PaC webhook receiver, registration-service).
+
+### Negative / compromises
+
+- **New operational component**: deploy, monitor, rate-limit, and secure a public HTTP endpoint.
+- **Private repo limitation**: badges unavailable for private repos until a signed-URL scheme is designed.
+- **Cache staleness**: badge status may lag real-time by up to 60 seconds.
+- **GitHub API rate limit**: each GitHub App installation supports a finite number of repos at a given cache TTL (see Caching section). Multi-cluster deployments need one badge service per cluster (or per GitHub App installation).
```
Omeramsc: What does it mean in our case? is it blocking repositories from using the feature, or just introduce a lag? if so, how long?

Omeramsc (changes requested):
Nice job, left some questions. Would love to see a diagram (activity?) of some sort here.

ADR/0073-readme-status-badges.md
Shiri-sh: Spike findings = research log (what we tried). Alternatives Considered = ADR decision section. I’ll combine them both so we don’t repeat the same rationale twice.

ADR/0073-readme-status-badges.md
Shiri-sh: Staleness is only lag (up to ~60s), not blocking. Rate limits matter only under heavy sustained load. Phase 1 should be fine, we’ll monitor shared quota with PaC.

ADR/0073-readme-status-badges.md
Shiri-sh: the service wont be available to private repos on phase 1, it will be on phase 2 - expended on that in the next commits

ADR/0073-readme-status-badges.md
```diff
@@ -0,0 +1,213 @@
...
+```
+GET /badge/{githubOrg}/{githubRepo}/{branch}/build      → image/svg+xml
+GET /badge/{githubOrg}/{githubRepo}/{branch}/ec         → image/svg+xml
+GET /badge/{githubOrg}/{githubRepo}/{branch}/build.json → application/json
+GET /badge/{githubOrg}/{githubRepo}/{branch}/ec.json    → application/json
+```
+
+JSON response example (public repo, or private repo with valid `sig` in Phase 2):
+
+```json
```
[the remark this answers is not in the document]
Shiri-sh: added

ADR/0073-readme-status-badges.md
```diff
@@ -0,0 +1,213 @@
...
+
+Beyond README badges, the same service can power **monitoring dashboards**, **portfolio status pages**, and **third-party integrations** (Slack bots, Jira, Confluence) via a JSON endpoint.
+
+### Constraints
+
+- README is rendered by GitHub — badge URL must be **reachable over public HTTPS**.
+- Konflux primary API is **Kubernetes CRDs** ([architecture overview](https://konflux-ci.dev/architecture/)); user-facing HTTP endpoints are exceptional but have established precedent (Tekton Results ingress per [Pipeline Service](../architecture/core/pipeline-service.md) and [ADR-0009](0009-pipeline-service-via-operator.md), PaC webhook receiver per [ADR-0039](0039-workspace-deprecation.md), registration-service). The [architecture overview](../architecture/index.md) permits these exceptions when justified.
+- Check runs are the status entries visible on each commit's **Checks tab** — the `-on-push` pipelines and `enterprise-contract` checks that PaC reports back to GitHub.
+- Check names are **per-component** (`*-on-push`, `*enterprise-contract*`), not one org-wide GitHub check name.
```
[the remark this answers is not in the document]
Shiri-sh: added

ADR/0073-readme-status-badges.md
```diff
@@ -0,0 +1,213 @@
...
+JSON response example (public repo, or private repo with valid `sig` in Phase 2):
+
+```json
+{
+  "org": "konflux-ci",
+  "repo": "build-service",
+  "branch": "main",
+  "type": "build",
+  "status": "passing",
+  "sha": "abc123...",
```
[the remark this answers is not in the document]
Shiri-sh: added

ADR/0073-readme-status-badges.md
```diff
@@ -0,0 +1,213 @@
...
+
+**Evolution for cluster data sources.** When the service extends to read Tekton Results or Release CRs, it will need access to cluster-internal APIs. The deployment model evolves within the same codebase — new data source implementations are added. Two routing options exist:
+
+1. **Per-cluster instances** behind a routing layer that directs requests to the correct cluster based on an org/repo → cluster mapping (derivable from the platform's tenant onboarding configuration).
+2. **Central service** that queries Tekton Results APIs on remote clusters via their existing HTTP ingresses.
+
+This decision is deferred because it depends on which cluster data sources are needed and how they expose their APIs.
```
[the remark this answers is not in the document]
Shiri-sh: added

ADR/0073-readme-status-badges.md
```diff
@@ -0,0 +1,213 @@
...
+3. Uses the **paginated GitHub Checks API** initially for **build** and **EC** status on a branch HEAD.
+4. Extends later to **Tekton Results / Konflux CRs** for Ring and richer status, within the same codebase.
+5. Uses a **server-side GitHub credential** (GitHub App installation token stored in a Kubernetes Secret).
+
+### Data flow
+
+GitHub README images are fetched by GitHub's [camo proxy](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/about-anonymized-urls) — a plain HTTPS GET with **no viewer login** passed to the badge service. The badge service calls the GitHub API using the **PaC GitHub App** installation token (same App/credentials PaC uses to report check runs; stored in a cluster Secret).
+
+```mermaid
+sequenceDiagram
```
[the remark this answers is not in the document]
Shiri-sh: Accepted risk for Phase 1. badge service will only call read endpoints. Separate read-only App might be a future option.

ADR/0073-readme-status-badges.md
```diff
@@ -0,0 +1,213 @@
...
+
+### Caching and GitHub API rate limits
+
+The [5,000 requests/hour](https://docs.github.com/en/apps/creating-github-apps/registering-a-github-app/rate-limits-for-github-apps) limit is **per GitHub App installation**, **shared with PaC** (check runs, webhooks, PR comments, etc.). Badge service must monitor `X-RateLimit-Remaining` and back off.
+
+- Cache badge results per `{org}/{repo}/{branch}/{type}` (TTL 30–60s).
+- Cache `repo.private` (~15 min).
+- At 60s TTL, ~120 GitHub calls/hour per badge key. If ~3,000/hour is used by PaC, ~16 badge keys (~8 repos × 2 types) fit in the remainder under sustained load. Size TTL and repo count to **remaining** quota.
+
+### Deployment model
```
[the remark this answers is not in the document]
Shiri-sh: will figure this out in implementation

ADR/0073-readme-status-badges.md
```diff
@@ -0,0 +1,214 @@
...
+
+## Context
+
+### Problem
+
+Teams want **embeddable build status** on the default branch README (Jenkins-style pass / fail / running). Konflux reports pipeline status as **GitHub Checks** (Checks tab), not GitHub Actions. There is no official GitHub URL that renders a Checks badge image.
+
+Today, users must open the Konflux UI (SSO) or the GitHub Checks popup to see status. We want a **single dynamic URL** per repo that README can embed for **build**, **Enterprise Contract (EC)**, and later **Ring / release** data.
+
+Beyond README badges, the same service can power **monitoring dashboards**, **portfolio status pages**, and **third-party integrations** (Slack bots, Jira, Confluence) via a JSON endpoint.
```
cit1zen, line 29: Is the JSON functionality something that users asked for right now? We may want to limit the initial scope of the service.

ADR/0073-readme-status-badges.md
```diff
@@ -0,0 +1,214 @@
...
+
+### Constraints
+
+- README is rendered by GitHub — badge URL must be **reachable over public HTTPS**.
+- Konflux primary API is **Kubernetes CRDs** ([architecture overview](https://konflux-ci.dev/architecture/)); user-facing HTTP endpoints are exceptional. The [architecture overview](../architecture/index.md) distinguishes two precedents: endpoints that expose **user/tenant cluster data** (e.g. Tekton Results ingress per [Pipeline Service](../architecture/core/pipeline-service.md) and [ADR-0009](0009-pipeline-service-via-operator.md)) require **SubjectAccessReviews** for consistent RBAC; **unauthenticated supporting endpoints** that do not expose tenant data (PaC webhook receiver per [ADR-0039](0039-workspace-deprecation.md), sprayproxy, registration-service) follow a separate precedent. **Phases 1–2** read public GitHub Checks data only — same category as the latter.
```
cit1zen, line 34: Minor optional ask, can you use unusual instead or rare of exceptional. Exceptional evokes in me that they are better than average.

ADR/0073-readme-status-badges.md
```diff
@@ -0,0 +1,214 @@
...
+
+**Phase 3 — Cluster data sources**:
+1. Extend to read Tekton Results for richer status (pipeline duration, last N runs).
+2. Extend to read Release CRs for ring/release badges.
+3. Evolve the deployment model to support cluster-internal data access (see Deployment model section).
+4. Introduce tenant resolution (org/repo → cluster/namespace mapping) for cluster queries.
+
+## Alternatives Considered
+
+| Alternative | Evaluated | Verdict | Rationale |
```
cit1zen, line 169: Have you considered using cluster/Konflux data directly from the start instead of relying on Github as a proxy source of truth?

ADR/0073-readme-status-badges.md
```diff
@@ -0,0 +1,214 @@
...
+
+### Security and auth model
+
+**GitHub credential**: reuses the **PaC GitHub App** on the hosting cluster ([Pipeline Service](../architecture/core/pipeline-service.md), [ADR-0039](0039-workspace-deprecation.md), [ADR-0009](0009-pipeline-service-via-operator.md)). No new App installation required.
+
+**Exposure**: SVG endpoint is unauthenticated (required for README/camo). Returns pass/fail/running for **public** repos only; **private** repos get a neutral SVG. No logs, secrets, or pipeline details. Phase 3 cluster reads will use SAR-guarded access.
+
+**Protection**: per-IP rate limiting at ingress; public badge URLs contain only org/repo/branch.
+
+### Phase 2 — Private repository access (planned)
```
cit1zen: I would simplify the name of the section.

```suggestion
### Private repository access
```

ADR/0073-readme-status-badges.md
cit1zen: You are already proposing using some cluster data and we would remove the dependency on GitHub API, plus we would make the feature implementable also for GitLab and other providers.

ADR/0073-readme-status-badges.md
```diff
@@ -0,0 +1,214 @@
...
+- **Simple initial deployment**: single instance on one cluster, reusing PaC's existing GitHub App credentials.
+- **Follows established precedent**: public HTTP endpoints exist in Konflux (Tekton Results, PaC webhook receiver, registration-service).
+- **Private-repo safety**: neutral badge avoids status leakage on unauthenticated SVG endpoint.
+
+### Negative / compromises
+
+- **New operational component**: deploy, monitor, rate-limit, and secure a public HTTP endpoint.
+- **Private repo README badges show no status** in Phase 1 (neutral badge only). Most onboarded repos may be private; primary value for them is JSON behind auth and Checks tab (unchanged).
+- **Cache staleness**: badge status may lag real-time by up to 60 seconds.
+- **Shared GitHub API rate limit**: badge service competes with PaC for the same 5,000/hour per installation; must monitor remaining quota and size cache TTL accordingly.
```
cit1zen: MintMaker is rate-limited fairly regularly. Would this worsen the existing problem?

ADR/0073-readme-status-badges.md
```diff
@@ -0,0 +1,214 @@
...
+- Cache `repo.private` (~15 min).
+- When GitHub rate limit is exhausted or the service backs off to protect PaC quota, return a **neutral badge** (no status) — do not serve stale or guessed status.
+- At 60s TTL, ~120 GitHub calls/hour per badge key. If ~3,000/hour is used by PaC, ~16 badge keys (~8 repos × 2 types) fit in the remainder under sustained load. Size TTL and repo count to **remaining** quota.
+
+### Deployment model
+
+**Initial deployment: single instance on one cluster.** Since the data source is the GitHub Checks API, the badge service does not need access to tenant namespaces for Phase 1. It deploys as a Deployment on a single Konflux cluster, reading that cluster's PaC GitHub App credentials. A Kubernetes Ingress or Route exposes it at a public URL. It serves badges only for repos where **its hosting cluster's** PaC GitHub App is installed and has `checks:read` access. Repos whose tenants are onboarded on other clusters use a different GitHub App per [ADR-0039](0039-workspace-deprecation.md); those repos require a badge service instance (or equivalent routing) on the cluster that owns that App installation.
```
cit1zen: I would maybe describe the deployment model a little bit more generically as RH is not the only one using Konflux.

If I understand it right, the service does not have to be deployed in Konflux cluster at all. It just needs a cluster as all the data used is fetched from Github.

cit1zen (changes requested):

ADR/0073-readme-status-badges.md
Shiri-sh: sure. I just picked that up from architecture/index.md.

ADR/0073-readme-status-badges.md
Shiri-sh: good catch- I missed MintMaker. Updated the ADR to explicitly list MintMaker as a quota consumer alongside PaC and to state the badge service must be the lowest-priority consumer (back off first). Also added the 5-min TTL option (~12 calls/hour per key vs ~120 at 60s) as a concrete lever when quota is tight.

ADR/0073-readme-status-badges.md
Shiri-sh: Good point. JSON was not explicitly requested by users - the ask is specifically for README badge images. I included it because the implementation cost is near-zero, the data-fetching and aggregation logic is identical for SVG and JSON. the only difference is the response serializer. 

By having those endpoints we can gain a big picture on all repos...I was thinking of sort of dashboard...Nice to have though. I can mark JSON as optional/deferred in the rollout if you feel it muddies the initial scope. It won't change the architecture or the API shape - just whether we wire up the .json routes in Phase 1.

ADR/0073-readme-status-badges.md
Shiri-sh: You're right, updated the deployment section to describe the requirements generically: a GitHub App token and a public HTTPS endpoint. The service can run on any Kubernetes cluster.

one question back: how would you recommend provisioning the GitHub App credentials?

ADR/0073-readme-status-badges.md
Shiri-sh: Thanks for raising this. For what I understand, starting with cluster data directly would involve a few challenges, but correct me if I'm wrong:

**Security precedent:** As far as I know, no Konflux service currently exposes cluster-internal pipeline data (PipelineRuns, Tekton Results) through a public unauthenticated HTTP endpoint. GitHub Checks data is already visible on public repos. I'd expect that going straight to cluster data would require SAR/RBAC design and broader review- a significantly larger scope for an MVP. But if there's an established pattern for this that I'm missing, I'd be glad to learn about it.

**Cluster isolation:** My understanding of the architecture constraints is that each cluster is independent and shouldn't coordinate across clusters. A cluster-data badge service would need access to tenant namespaces and would need to run per-cluster. That's the Phase 3 deployment model described in the ADR, but it felt like too much for a first iteration.

**Git-provider portability**: You're right that cluster data would make the service provider-agnostic (PipelineRun results are the same regardless of whether GitHub or GitLab triggered them). That's a real advantage of the cluster-data approach. The trade-off I saw was scope and approval complexity for the MVP vs. shipping something useful quickly with GitHub Checks.

The phased approach lets us get operational experience with a simpler service first and then evolve to cluster data sources. Would like to here your thoughts about it

ADR/0073-readme-status-badges.md
```diff
@@ -0,0 +1,214 @@
...
+```
+
+**Public vs private (one URL, different response):** the service reads `repo.private` from the GitHub API (cached ~15 min). **Public** → fetch check runs, return real status SVG/JSON. **Private** → requests without a valid signature get a neutral badge only (Phase 1). Phase 2 adds an HMAC query parameter for private README embeds (see below). **Fail-closed:** if repo visibility cannot be determined (API failure, rate limiting, cache miss with no response), return a **neutral badge** — never assume public.
+
+### API shape
+
+```
+GET /badge/{githubOrg}/{githubRepo}/{branch}/build      → image/svg+xml
+GET /badge/{githubOrg}/{githubRepo}/{branch}/ec         → image/svg+xml
+GET /badge/{githubOrg}/{githubRepo}/{branch}/build.json → application/json
```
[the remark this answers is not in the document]
Shiri-sh: the comment is fair but for the ADR its OK to leave as is.

ADR/0073-readme-status-badges.md
```diff
@@ -0,0 +1,214 @@
...
+- **Data path**: resolve branch HEAD SHA → paginate `GET /repos/{org}/{repo}/commits/{sha}/check-runs` → aggregate `*-on-push` / `enterprise-contract` → pass | fail | running | **unknown** (no matching check runs, e.g. newly onboarded or unconfigured branch).
+- **Rendering**: `badge-maker` library ([MIT OR Apache-2.0](https://github.com/badges/shields/tree/master/badge-maker)) or equivalent minimal SVG — **not** shields.io SaaS `check-runs` route.
+- **No tenant resolution needed initially**: The data source is the GitHub API, not cluster-internal data. The service only needs the GitHub org, repo, and branch to query check runs.
+
+### README usage
+
+```markdown
+![Konflux build](https://<badge-host>/badge/{org}/{repo}/main/build)
+```
+
```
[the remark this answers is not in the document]
Shiri-sh: @cit1zen Do you think it’s a hard blocker if the service doesn't have read-only permissions scoped down? Strictly speaking, the service will only execute read ops anyway, but is the exposure a high severity issue for us? If so, we’d need to spin up a dedicated GitHub App for this, which adds scope. Though on the plus side, a separate app gives us a fresh rate limit pool.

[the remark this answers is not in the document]
cit1zen: Yes. This could be quite problematic. I think that @konflux-ci/infrastructure folks could chime it.

ADR/0073-readme-status-badges.md
cit1zen: I do not have experience with provisioning of the GitHub app creds, so I cannot help you :(

ADR/0073-readme-status-badges.md
cit1zen: You make good point.

I think that going right away for the solution is better approach, especially the GitHub approach involves several hard problems like being rate-limited by the GitHub. Even if the initial investment into cluster-data centric solution is bigger, any other feature is easier to implement.

Shiri-sh: Make sense, and honestly the review process has been proving it- most of the issues raised (rate limits, MintMaker quota competition, over-privileged token, fail-closed on API failure) only exist because of the GitHub API data path.

I'll do a deeper dive into how we can implement this service. Since Jenkins already handles this, I'll see if we can mirror their approach here.

ADR/0073-readme-status-badges.md
```diff
@@ -0,0 +1,214 @@
...
+```
+GET /badge/{org}/{repo}/{branch}/build?sig=<hmac>
+```
+
+The service verifies `sig` against a cluster signing secret (`HMAC(secret, org/repo/branch/type)`). Valid signature on a private repo → return real status; missing or invalid → neutral badge (same as Phase 1). Signatures are **not short-lived** (README markdown is static). Konflux UI or CLI generates the signed URL when the user copies embed markdown. Rotating the signing secret invalidates all existing signed badge URLs; users must re-copy embed markdown.
+
+### Caching and GitHub API rate limits
+
+The [5,000 requests/hour](https://docs.github.com/en/apps/creating-github-apps/registering-a-github-app/rate-limits-for-github-apps) limit is **per GitHub App installation**, **shared with PaC** (check runs, webhooks, PR comments, etc.) **and MintMaker** (dependency update scans via the same App token — see [MintMaker](../architecture/add-ons/mintmaker.md)). MintMaker is already rate-limited on some installations. Badge service must be the **lowest-priority consumer**: monitor `X-RateLimit-Remaining` and back off before PaC or MintMaker are affected.
+
```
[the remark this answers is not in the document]
Shiri-sh: not relevant anymore. see [9ccf87e](https://github.com/konflux-ci/architecture/pull/376/commits/9ccf87e2917bfe71731679f7e6f91953b6260ead)

ADR/0073-readme-status-badges.md
Shiri-sh: not relevant anymore. see [9ccf87e](https://github.com/konflux-ci/architecture/pull/376/commits/9ccf87e2917bfe71731679f7e6f91953b6260ead)

ADR/0073-readme-status-badges.md
Shiri-sh: not relevant anymore. see [9ccf87e](https://github.com/konflux-ci/architecture/pull/376/commits/9ccf87e2917bfe71731679f7e6f91953b6260ead)

ADR/0073-readme-status-badges.md
Shiri-sh: @cit1zen please review the new changes, thanks!

ADR/0073-readme-status-badges.md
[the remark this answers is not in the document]
Shiri-sh: not relevant anymore. see [9ccf87e](https://github.com/konflux-ci/architecture/pull/376/commits/9ccf87e2917bfe71731679f7e6f91953b6260ead)

ralphbean:
Thank you for working on this, but I'm a strong veto on serving badges from the cluster with anonymous read. If the network is on a private network, this requires punching holes in firewalls. That's not going to happen - or, it can't be a pattern we recommend.

A few alternatives to consider:

S3 bucket with push controller — Single infra-provided public S3 bucket, controller on cluster pushes badge SVGs there. Problem: all badge info for all tenants would be visible in the same place. If we're worried about leaking that a team is even using Konflux, that's a blocker. We could require opt-in — disable by default, only enabled when a team sets an annotation on their Component. Then it's their choice.

No service, teams DIY — Have teams do this themselves with a `finally` task in their pipelines (build, test, release — or a finally in the release pipeline, not just a task). They bring their own S3 bucket, their own creds, wire it up, own the operation. Higher overhead for them, but no surprises.

Reconsider shields.io — It's built for this, right? I see the ADR mentions limitations, but if we make badges readable on the internet without authz (via S3 or a cluster endpoint), it's world-readable anyway. Third-party services accessing the data is what this whole thing is about. What's the concern with shields.io that makes it a non-starter? I don't yet understand the "many rows" thing. (:crossed_fingers:  I hope this can be made to work.)

Shiri-sh:
@ralphbean Thanks for the feedback. We will pivot away from exposing an unauthenticated cluster endpoint and evaluate the S3 push controller vs. the DIY pipeline task approaches you suggested.

To answer your question regarding Shields.io and the "many rows" issue: The public Shields.io SaaS is unfortunately a non-starter due to a hard limitation in how it interacts with GitHub's API pagination.

**The 30-Row Limit**: Shields.io queries the GitHub check-runs REST API. By default, GitHub paginates this endpoint, returning a maximum of 30 check runs per page.

**No Pagination in Shields**: The Shields.io backend does not implement pagination for the check-runs route. It strictly reads the first page of results (the first 30 checks) and stops.

**Repositories are Heavy**:  repositories using Konflux frequently generate 20 to 70+ check runs on a single commit. The specific checks we need to track (like *-on-push or enterprise-contract) often land on row 35 or 50, placing them on page 2 or 3 of the API response.

**The Result**: Shields.io looks at the first 30 rows, fails to find the Konflux check, and renders a broken "check not found" badge.

Unless we fork and maintain our own internal version of Shields.io specifically to add pagination support, we can't rely on it.

Omeramsc:
@Shiri-sh I think there are 2 possibilities worth researching before going through the overhead of an S3 bucket.
1. We can try to contribute to the Shield.io repository a fix that will allow filtering the result on our hand, before handing out a `check-name` parameter that will narrow down the results. 
2. using a Dynamic JSON to feed shield.io the results. however, here we will have massive tradeoffs (limitation of 60 calls per hour + no colored badges). 

Let's investigate if something here can provide value. if not, then we might go the S3 route, but I want to see it as a last resort. 

cc @ralphbean

Shiri-sh:
**Regarding the Dynamic JSON route:**
I think it's a dead end. Pointing Shields directly at GitHub hits the 60-call/hour anonymous limit immediately. If we end up building our own service to read the GitHub API (like our original plan f83585d2d473898097e4d82d3d332c8aef8d12d0), we wouldn't even use JSON or Shields - we'd just render the SVGs ourselves. But that puts us right back into building and maintaining a custom public service that has its own limits as mentioned.

**About the upstream fix to Shields.io - looks promising:**
Right now they fetch page 1 and filter client-side. If we just pass check_name to GitHub's API, it filters server-side and completely bypasses the pagination issue.

I opened a [PR](https://github.com/badges/shields/pull/12150) for this (it's literally a one-line fix). If merged, badges just work - no S3, no firewall holes, no DIY tasks.
_(based on recent history, small PRs merge in 2-5 days)_

It does have limitations (public repos only, GitHub only, Build/EC only). But for a start, it's pretty solid

ralphbean:
Driving changes upstream is the best! :crossed_fingers: on the shields patch getting accepted!

ralphbean:
Ok, thinking. The reason I’m leaning toward closing this without merging is that the current ADR records a Konflux-owned badge service with anonymous cluster access, and that architecture no longer looks viable given the network and firewall constraints.\n\nIf we use Shields.io, this probably doesn’t need an ADR: we’d be consuming an external service. If teams use their own pipeline task and S3 bucket, that also seems like implementation guidance rather than a Konflux architecture decision. An ADR would make sense if Konflux provides shared S3 infrastructure, a push controller, or a standard task with platform-level ownership, security, and tenancy implications.\n\nWith that in mind, I intend to close this PR without merging the current ADR once we’ve settled the implementation path. I’ll leave it open for now so we can capture that direction here; if we introduce Konflux-owned shared infrastructure, we can open a new ADR focused on that decision.

gbenhaim:
In github, the status of all the checks for the recent commit that was merged appear in the main page of the repo with a v/x signs, I think it serves the same purpose. for example:

image


image

ralphbean:
@gbenhaim yeah, the badge just serves as nice candy for the README.md. ;)

Omeramsc:
> In github, the status of all the checks for the recent commit that was merged appear in the main page of the repo with a v/x signs, I think it serves the same purpose. for example:
> 
> image image

Yup, the idea is increased visibility for the pipeline health (and possibly more things in the future). but I agree with Ralph that the scale and context of this probably does not justify an ADR (our original approach), especially considering the two solutions at hand (and everything else have too much of an overhead for the value users will get out of this IMO).

Shiri-sh:
Thanks @ralphbean @Omeramsc - this framing really helps.

I pulled the numbers on our onboarded repos to see if building shared S3 infrastructure is actually worth the overhead. It turns out **~57% of our total repos (including GitLab) are public GitHub repos**. These will be perfectly covered by Shields.io once our upstream check_name PR merges (which is moving forward either way).

I'll look more closely into the S3 implementation, and if it's feasible, we can open a new, focused ADR for it.

Given all this, I'm going to close this.

PR #380, opened 2026-09-10
Add ADR-0072: Component ownership and routing metadata

rhmdnd:
Propose carrying ownership and routing metadata on the artifact as standard OCI metadata: labels on non-OCI images and annotations on OCI manifests, per ADR-0061 and verified by Conforma. Reuse existing keys and policy rather than invent new ones -- org.opencontainers.image.authors for the ownership contact (deprecating maintainer in its favor), org.opencontainers.image.source for the source repository, and Conforma's required_labels/effective_on for phased enforcement.

Enforcement ratchets from a build/staging warning to a blocking managed release task that resolves each value against the deployment's backend. Issue-tracker routing stays deployment-specific and out of the upstream-mandated set, shown by example (com.redhat.tracker).

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/ADR/0072-component-ownership-metadata.md
+++ b/ADR/0072-component-ownership-metadata.md
@@ -1,0 +1,111 @@
+---
+title: "72. Component Ownership and Routing Metadata"
+status: Proposed
+applies_to:
+  - build-service
+  - release-service
+  - enterprise-contract
+topics:
+  - ownership
+  - routing
+  - container-images
+  - provenance
+  - enforcement
+---
+
+# 72. Component Ownership and Routing Metadata
+
+Date: 2026-09-10
+
+## Status
+
+Proposed
+
+[//]: # (One of: Proposed, Withdrawn, Deferred, Accepted, Implementable, Implemented, Replaced)
+
+## References
+
+* [ADR-0061](0061-vcs-info-specification.md) - Carries metadata as image labels and OCI manifest annotations, enforced by Conforma. This ADR reuses that pattern and its source-repository key.
+* [ADR-0056](0056-revised-component-model.md) - Revised Component Model; target for any resource-level extension.
+* [OCI Image Spec: Annotations](https://specs.opencontainers.org/image-spec/annotations/) - the `org.opencontainers.image.*` keys and the reverse-domain rule for custom keys.
+* [rhtap-ec-policy `rule_data.yml`](https://github.com/release-engineering/rhtap-ec-policy/blob/main/data/rule_data.yml) - an example downstream Conforma label policy.
+
+## Context
+
+Konflux cannot reliably say who is accountable for a shipped artifact or where its
+issues should be routed. Today the answer lives in people's heads, in
+spreadsheets, or in ad hoc tooling. Operators feel it most, since their work is
+tracked in issue-tracker projects.
+
+Even though components, operators, and namespaces all need ownership, we can
+solve all three by carrying OCI metadata on the artifact, making it available
+for validation in the release pipeline.
+
+## Decision
+
+Carry ownership and routing metadata as standard OCI metadata. Specifically,
+labels on non-OCI images and annotations on OCI manifests, per
+[ADR-0061](0061-vcs-info-specification.md) and verified by Conforma per
+`mediaType`. Reuse existing keys and policy rather than invent new ones:
+
+* Ownership: `org.opencontainers.image.authors`, the OCI freeform contact
+  string, instead of a new `owner` label. Deprecate the legacy `maintainer` in
+  its favor through Conforma's `deprecated_labels` and `effective_on`, with a
+  consistency check during the transition.
+* Issue Routing: deployment-specific, since the OCI spec has no issue tracker
+  key. This ADR mandates none; a deployment adds one with a namespaced label and
+  a release task.
+
+Tekton Chains already captures image labels into the signed build provenance, so
+the metadata is attestation-backed with no pipeline change. Enforcement reuses
+Conforma's `required_labels` and `effective_on`, and ratchets:
+
+1. Build and staging: a build missing the metadata warns; nothing blocks.
+2. Production release: a managed release task resolves each value against the
+   deployment's backend — `authors` could read as a comma-delimited list
+   against an identity provider (for example, each a group that exists and
+   meets a minimum size), a tracker reference against the issue tracker — and
+   blocks the release if any fails.
+
+The backends, and how a value resolves, are deployment-specific and out of scope.
+Upstream defines the gate; it does not name the backend. An RPM carries neither
+labels nor annotations, so covering it needs a resource-level carrier
+([ADR-0056](0056-revised-component-model.md)) and is out of scope.
+
+### Deployment example
+
+A Red Hat deployment relaxes its `com.redhat.component` (once a Bugzilla
+component) to accept a tracker component, and adds `com.redhat.tracker` for the
+project URL. Custom keys use reverse-domain notation, so they stay out of the
+upstream set. On a non-OCI image the same keys become annotations on a manifest:
+
+```dockerfile
+LABEL org.opencontainers.image.authors="team@example.com"
+LABEL org.opencontainers.image.source="https://github.com/example/component"
+LABEL com.redhat.component="Example Component"
+LABEL com.redhat.tracker="https://issues.example.com/projects/EXAMPLE"
+```
+
+## Consequences
+
+* The metadata is a queryable, signed property of the artifact. Anyone holding
+  it, like a security team with a CVE, reads the contact and routing from the
+  labels or the verified provenance, tamper-evident, with no access to the
+  producing cluster.
+* Reusing OCI keys, [ADR-0061](0061-vcs-info-specification.md)'s carriers, and
+  Conforma's policy machinery means no new upstream label and no build-pipeline
+  change; the only new part is adding a validation task to the release pipeline.
+* The owning team fixes a stale record directly. A correction applies going
+  forward, not to artifacts already shipped, so provenance stays immutable.
+* `effective_on` dates make enforcement a migration, not a cutover.
+* Artifacts that carry neither labels nor annotations (RPMs) fall outside this
+  mechanism and need a resource-level carrier in a follow-up.
+
+## Open Questions
+
+* `authors` is freeform, so the list format and normalization are the validating
+  task's to define, not the spec's.
+* An artifact may map to several tracker projects; single- versus multi-valued
+  routing, and how a set is validated, is open.
+* A private tracker will not resolve for outside consumers; whether external
+  artifacts need a different routing signal is open.
--- a/ADR/AGENTS.md
+++ b/ADR/AGENTS.md
@@ -9,6 +9,7 @@ Status: Ac=Accepted Im=Implemented Ib=Implementable Pr=Proposed Ap=Approved Rp=R
 0047 Ib 70L Caching for container base images used during builds
 0007 Ac 89L Change Management Process
 0029 Ac 192L Component Dependencies
+0072 Pr 111L Component Ownership and Routing Metadata
 0060 Pr 294L ComponentGroups
 0027 Pr 109L Container Image Management Practice
 0035 Ac 59L Continuous Chaos Testing of Apps in AppStudio


ADR/0072-component-ownership-metadata.md
```diff
@@ -0,0 +1,326 @@
...
+Ownership today lives in people's heads, in out-of-band spreadsheets, or is
+inferred from namespace naming conventions that were never designed to carry
+that meaning.
+
+Several efforts have converged on the same underlying need from different
+angles:
+
+* ownership metadata on Konflux resources generally,
+* ownership scoped specifically to operators, and
+* ownership at the namespace level.
```
rhmdnd, line 47: I kept these last two in here, but I don't know if we really need them.

Technically, if ownership is associated via a Containerfile label, then we don't really need to add it to the Konflux resources do we? If we do, we'll have to keep them in sync. What happens if Konflux resources have a different owner from what documented in the containerfile?

Additionally, could namespace ownership be inferred a container image produce by a component in that namespace?

I'm just wondering if we're making things harder by documenting the same information in multiple places, and having to keep it all in sync. Another benefit for using the Containerfile label, is that it's already slurped into the image attestation. If we want ownership of the Konflux resources to be cryptographically verified using cosign, we'll need to propagate ownership data from CRD metadata to image provenance at some point (which I don't think we currently have?).

ADR/0072-component-ownership-metadata.md
```diff
@@ -0,0 +1,326 @@
...
+provenance automatically, and enforced along a ratcheting path that ends by
+blocking a production release until the owner resolves to a valid group in an
+identity provider. The model is described first, then the three stages that
+describe the enforcement.
+
+### One ownership model across all consumers
+
+There is a single ownership model. The same label, the same semantics, and the
+same enforcement apply whether the consumer is a component, a component group,
+an operator, or a namespace. Consumers differ only in which resource carries
```
rhmdnd, line 112: This isn't really true. The ADR describes how we can enforce containerfile labels in the release pipeline, but it doesn't elude to enforcing the Konflux CRD or namespace owners.

ADR/0072-component-ownership-metadata.md
```diff
@@ -0,0 +1,326 @@
...
+  downstream consumers and release gating actually read.
+* On the Konflux resources — `Component`, `Component Group`, and the tenant
+  namespace — as `metadata.labels.owner`. This covers ownership of resources
+  that are not themselves images (a `Component Group`, a namespace) and of
+  components that have not produced an image yet.
+
+Example — in the `Containerfile`:
+
+```dockerfile
+LABEL owner="my-team"
```
rhmdnd, line 143: Another really nice thing to add here would be:

```dockerfile
LABEL jira-project="FOO"
```

But then we'd need to make an assumption about the instance to validate that against.

ADR/0072-component-ownership-metadata.md
```diff
@@ -0,0 +1,326 @@
...
+why ownership should be backed by group membership, and not individual owners.
+
+### Stage 3 — Validating the owner group at production release (blocking)
+
+The final stage runs at the production-release gate, in the managed release
+pipeline, and is the first stage that blocks. A release-pipeline task resolves
+the `owner` reference against the deployment's identity/group backend (e.g.
+LDAP or Google Groups) and fails the release unless the reference resolves to a
+valid group that satisfies policy. For example, a group that exists and has at
+least a minimum number of members.
```
rhmdnd, line 219: We should consider if we want to enforce Jira project validity, too. That is really useful information for consumers (e.g., security teams routing findings).

ADR/0072-component-ownership-metadata.md
```diff
@@ -0,0 +1,326 @@
...
+  artifacts and experiment, and ownership presence is surfaced (not blocked)
+  through Stages 1 and 2. The cost is that an owner value may be a placeholder
+  until Stage 3 validates it at the production-release gate.
+
+* Teams get a migration window instead of a cutover. The trade-off is that
+  production releases are eventually blocked on a valid owner group, and teams
+  that ignore the earlier stages hit that block.
+
+## Open Questions
+
```
rhmdnd, line 308: Should we also add repository as a required label?

ADR/0072-component-ownership-metadata.md
calfonso: I see it as more than nice to have, actually. When we discover work that is needed on an operator, our tracking system to do that is in a jira project. without this it's anything goes to go figure out the right jira project. that is known for some via cyborg, which isn't consistently present by anything outside of hybrid platforms. Not having this information causes lots of work and inaccuracy.

ADR/0072-component-ownership-metadata.md
calfonso: There is some level of confusion I have with the term ownership. I've seen org evolution in some places move to a fungible allocation of who works on what, and there is no clear ownership contact. The org I run is in exactly this scenario where we work in pods, those pods get assigned component areas for a 4 month dev cycle and then it changes the next dev cycle. I don't have a clear owner of components across our repos which span across 11 jira projects.

rhmdnd: When new work comes in through Jira, how do you dispatch it to pods (tmp teams)? Perhaps a better question is, how are bugs routed for features that no longer have a pod?

ADR/0072-component-ownership-metadata.md
ralphbean, line 143: From https://specs.opencontainers.org/image-spec/annotations/, should we use `org.opencontainers.image.authors` (freeform string) here instead of the flat `owner`?

ADR/0072-component-ownership-metadata.md
ralphbean: If Red Hat were to do this, they should namespace this to something like `com.redhat.jira/project` or something and leave this detail out of the upstream ADR. There's no certainty that anyone else deploying and using konflux would have JIRA as part of their infrastructure.

ADR/0072-component-ownership-metadata.md
```diff
@@ -0,0 +1,326 @@
...
+membership, does not enumerate members, and does not confirm the group exists
+until the final enforcement stage. Resolving the reference to people is the job
+of a deployment-specific backend and release pipeline.
+
+The reference is carried as an `owner` label in two complementary places:
+
+* On the container image, set via a `Containerfile` `LABEL` (the preferred
+  mechanism). This is the surface that travels with the artifact and that
+  downstream consumers and release gating actually read.
+* On the Konflux resources — `Component`, `Component Group`, and the tenant
```
ralphbean, line 135: When I first read this, I thought the purpose of the `metadata.labels.owner` field on the custom resources was to serve as a *default value* that should be _applied_ as the `owner` label on the image in a situation where the no explicit owner label was provided for the image. A convenience feature.

But, looking again, I think you mean these two things to be orthogonal. You just want an owner label on resources so that anyone looking can understand the ownership .. of that resource. The owner label on the Component might differ from the owner label on an image. Is that your intent (just trying to clarify)?

ralphbean: FWIW, operationally, I tend to interpret the tenant namespace as the "owner" of all the resources in it. Tenant namespaces are typically of the form `<team>-tenant` and RBAC rules apply broadly within the namespace. With that habit in mind, I see the value of declare an explicit owner on the labels of the image, but the owner label on the custom resources doesn't seem valuable to me at first glance, given the existing implied ownership information carried by the namespace name.

ralphbean:
(Thanks for submitting this!)

FYI, there is a [process](https://github.com/konflux-ci/community/blob/main/ADRs.md#process). A key step is that you're expected to either (a) announce this ADR on the [community call](https://github.com/konflux-ci/community/#get-involved) or (b) announce it on the [public mailing list](https://github.com/konflux-ci/community/#get-involved) - so that people know to look at it, think about it, and comment.

ADR/0072-component-ownership-metadata.md
calfonso: That would be very wise in that it provides a very useful mapping. There is all sorts of good in the repo, not just the source code. Having this info would let us build additional tooling ideas around the meta data that lives there.

ADR/0072-component-ownership-metadata.md
calfonso: Great question. In my org we have a bug dashboard with backing jira filter that aggregates jira projects that the org is responsible for. So if you file an issue in one of the 11 projects it shows up on our boards. And we have a rotating group of engineers that monitor, triage, fix. Agree with your comment above.

ADR/0072-component-ownership-metadata.md
calfonso: Thinking about what we would do with the owner, is it to communicate or assign work? depending on the org, the process of how issues are assigned is differs greatly. some don't assign things until they are being worked on, some people have a default assignment of an email list. some people have default associates.

rhmdnd: Actually - it looks like this is already accounted for in the EC policy:

https://github.com/release-engineering/rhtap-ec-policy/blob/main/data/rule_data.yml#L82-L83

```yaml
  required_labels:
  - name: com.redhat.component
    description: The Bugzilla component name where bugs against this container should be reported by users.
```

We could update that to accept Jira, and add validation there, too. Or we could supplement it with something like:

```yaml
  required_labels:
  - name: com.redhat.component
    description: The Bugzilla or Jira component name where bugs against this container should be reported by users.
  - name: com.redhat.tracker
    description: The Bugzilla or Jira project URL.
```

Then, consumers would have the component and the project routing information.

ADR/0072-component-ownership-metadata.md
p-rog: I agree with @ralphbean here. The tenant namespace as the "owner" of all the resources in it. Additionally, adding "Jira" project reference won't address the problem if that project really owns the particular container build. In Jira many times a few container builds are grouped into one Jira component that reflect the specific functionality in the project. Don't forget also about non-public Jira projects. Specifying such Jira project name in the image labels won't bring any value to the image consumer.

rhmdnd: Understood.

I also realized that documenting a `metadata.labels.owner` on resources with the intention that it matches the `Containerfile` `owner` label introduces a precedence and syncing problem.

If they're not using the same value, which one takes precedence?

I'll rip out the details regarding Konflux CR ownership and keep the ADR focused on ownership enforcement through `Containerfile` labels.

ADR/0072-component-ownership-metadata.md
rhmdnd: Updated the ADR to account for this. Let me know your thoughts.

ADR/0072-component-ownership-metadata.md
rhmdnd: Already reviewed!

https://github.com/konflux-ci/architecture/blob/main/ADR/0061-vcs-info-specification.md

Refactored this ADR to be specific to routing and ownership since 0061 takes care of the repos.

ADR/0072-component-ownership-metadata.md
rhmdnd: I was thinking this would be for communication. But, I'm thinking of this from a security perspective where I have what I think is a legitimate finding and I need a contact that I can disclose it to.

rhmdnd: Incorporated the OCI spec annotation, similar to 0061, instead of rolling our own annotation/label.

ADR/0072-component-ownership-metadata.md
rhmdnd: I left the tracker example in there, primarily because the OCI spec doesn't have an annotation or label for "issue trackers". I kept the example specific to a downstream implementation, but only to show how a deployment could achieve this (custom label with a verification task prior to release).

ADR/0072-component-ownership-metadata.md
```diff
@@ -0,0 +1,111 @@
...
+A Red Hat deployment relaxes its `com.redhat.component` (once a Bugzilla
+component) to accept a tracker component, and adds `com.redhat.tracker` for the
+project URL. Custom keys use reverse-domain notation, so they stay out of the
+upstream set. On a non-OCI image the same keys become annotations on a manifest:
+
+```dockerfile
+LABEL org.opencontainers.image.authors="team@example.com"
+LABEL org.opencontainers.image.source="https://github.com/example/component"
+LABEL com.redhat.component="Example Component"
+LABEL com.redhat.tracker="https://issues.example.com/projects/EXAMPLE"
```
rhmdnd, line 86: @calfonso - kept this here to show how we could continue to associate trackers with images or OCI artifacts (even though it's a downstream example).

We could eventually add it as a required label like we do with `com.redhat.component` - https://github.com/release-engineering/rhtap-ec-policy/blob/main/data/rule_data.yml#L82-L83

